PR #43, opened 2022-05-12
WIP: Add concurrent load test for feed generating query

Aivean:
Add simple test that simulates concurrent feed generation requests like:

```sql
SELECT site_id FROM user_sites WHERE user_id = ?;

SELECT * from posts where site_id in (?) order by commented_at desc limit ?, ?;
```
Usage example (100 concurrent users, offset 1-8 pages, run for 20 seconds):

    npm run load-test -- -u 100 -p 8 -t 20

Additional changes: moved all utility scripts into `backend/utils` directory.


mugabe (changes requested):

Aivean:
stale

Issue #58, opened 2022-05-16
Reload posts when monster icon is clicked


kotkotoff:

Aivean:
Fixed in  #334

Issue #21, opened 2022-05-05
Vote list tweaking and refactroing


mugabe:
- [x] ~votes should be sorted in reverse order~ (fixed in #52)
- [x] long usernames should not break layout
- [x] transparent scrollbar
- [ ] api should not be requested second time if the rating has not been changed

PR #59, opened 2022-05-16
Vote list improvements

kotkotoff:
Reload posts when monster icon is clicked #58
Vote list tweaking and refactroing #21
  long usernames
  transparent scrollbar


ArtCraft:
Problem was solved within another PR

Issue #51, opened 2022-05-13
Post/comment edit


mugabe:
Stage 1:
- Posts and comments edit feature itself
- Simple history of changes

Stage 2:
- Visual diff

PR #63, opened 2022-05-17
Post and comment editing

mugabe:
resolves: #51

Done:
- Edit buttons in own comments and posts (shown only at post page)
- History of changes

To lazy to do right now:
- Diff and percentage of changes


Aivean:
Note: since the repo is now public, we can use “draft” status instead of “WIP in the title:
WIP  Post and comment editing by mugabe · Pull R…

PR #76, opened 2022-05-25
No subdomains

mugabe:


frontend/src/AppState/AppState.tsx
```diff
@@ -1,78 +1,153 @@
...
+    }
+
+    @action
+    setUserInfo(value: UserInfo | undefined) {
+        this.userInfo = value;
+    }
+
+    @action
+    setUserStatus(value: UserStatus) {
+        if (value.notifications !== this.userStatus.notifications) {
```
Aivean, line 68: Why not use  `observable.struct` for `userStatus` instead of manual deep equality checks?

Aivean (commented):
~~Pagination is broken when viewing subsite (e.g. http://orbitar.local/s/subsite47): pages lead to the main site.~~

UPD: fixed in 591a253579f32522f366b4d72731f2455c3181de

frontend/src/Pages/FeedPage.tsx
```diff
@@ -5,65 +5,64 @@ import PostComponent from '../Components/PostComponent';
 import Paginator from '../Components/Paginator';
 import {Link, useMatch, useSearchParams} from 'react-router-dom';
 import {FeedType, useFeed} from '../API/use/useFeed';
+import {APIError} from '../API/APIBase';
 
 export default function FeedPage() {
```
Aivean, line 10: `FeedPage` must be made `observer`, as otherwise it doesn't react to the `site` change (noticeable, if you click on subsites in the left panel).

Aivean:
Left panel has separate scroll. Is that intentional?
orbitar-left-panel-subsites

Aivean:
There is some weirdness going on with subsite changes.
I installed `autorun` to log `site` and `siteInfo` values and (sometimes) I see the following picture:

image

Notice, how `subscribe` part is present only in the second log record.

---

Additionally, it seems that `siteInfo` in the periodic `status` update request is missing `id` field:
image
which triggers mobx listeners.

frontend/src/API/use/useFeed.ts
```diff
@@ -44,7 +44,7 @@ export function useFeed(id: string, feedType: FeedType, page: number, perpage: n
                 })
```
Aivean, line 44: Bug: when switching between feed types (in this case from  `watch-all` to `subsite`)  `useEffect` in `useFeed` is triggered twice:
imageAs a result, current posts could be set to the value that corresponds to the previous feed.

I'm not sure why this happens, it seems that `FeedPage` is called in some transition state, when `AppState.site` is still "main". Perhaps this is an artifact of having `useSearchParams` and `useMatch` inside the `FeedPage`.

Aivean: Ok, here is what is happening:
When location is changed, in the rendering phase the the body of the `FeedPage` is rendered (and `feedType` is calculated). After that, during the mounting phase the effect hooks in `SiteWatcher` and `useEffect` are fired. The problem is, even though `site` is already changed, when `useFeed` effect is fired, the `feedType` has it's previous value, before site was changed.

A quick workaround would be to move `useMatch` into the same `SiteWatcher` component and make them part of the `AppState`.

I'll try to think about a better approach.

Aivean: I found a better way to fix this issue:
https://github.com/spaceshelter/orbitar/commit/11a044609380218ec23d3806a578c76338bcbc6d

I used [a small library](https://github.com/superwf/mobx-react-router) that bridges react-router and mobx, removing intermediate steps with hooks and effects from synchronization of mobx state with actual location and history.

Note, how `site` and `siteInfo` are now directly derived from the underlying state.

frontend/src/Pages/FeedPage.tsx
Aivean: with https://github.com/spaceshelter/orbitar/commit/11a044609380218ec23d3806a578c76338bcbc6d `observer` is not strictly required, as `useSearchParams` and `useMatch` hooks trigger the render of `FeedPage` on location change anyway, but it may be needed in the future.

Aivean: Fixed in ad82828192517f0701ef3dc9b7a62e4a0cc0f3c6

frontend/src/API/use/useFeed.ts
Aivean: Fixed in ad82828192517f0701ef3dc9b7a62e4a0cc0f3c6

Aivean:
I also did some cosmetic work on `SiteSidebar` to prevent its flickering (quick opening and closing) on site changes in f0c7b4bdee26830aafd39dbdb4f3be5090a6e805

Aivean (approved):

Aivean (approved):

Aivean:
minor nit: number of subscribers is not updated when subscribing to a subsite:
orbitar-left-panel-subsites-subscribe

PR #79, opened 2022-05-26
Log user visit activity

Aivean:
As an important prerequisite for the senate elections and activity based rating/karma models, we need to track user activity.

This is the simplest possible implementation that records if user visited the site (feeds) in a particular 12 hour period.

### Implementation:

Table:
```
 create table user_activity
      (
          user_id    int      not null,
          visited_at datetime not null,
          constraint user_activity_pk
              primary key (user_id, visited_at)
      );
```
`visited_at` is truncated to a period (currently 12h) on the server before insertion. No additional indexes and foreign keys are added for insertion performance.

Insertion is performed in `FeedController` in async fire-and-forget manner. There is an in-memory cache on the server that limits the number of insertions.

### DIFF
--- a/backend/migrations/20220526092700-user-activity.js
+++ b/backend/migrations/20220526092700-user-activity.js
@@ -1,0 +1,32 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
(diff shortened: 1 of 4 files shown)


mugabe (approved):

mugabe:
I moved table to separate database to prevent main database flooding with non-content data.

PR #89, opened 2022-07-01
karma calculator v0

ArtCraft:


frontend/src/Components/Karma.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+    const positiveCommentsDivisor = 10; // positive comments are 10x cheaper than posts
+    const negativeCommentsDivisor = 2; // negative comments are only 2x cheaper than posts
+    //TODO extract the rest of coefficients to constants and comment them
+
+    useEffect(() => {
+        setU( clamp(profileVotingResult, -profileVotesCount, profileVotesCount));
+    },[profileVotesCount]);
+
+    useEffect(() => {
```
Aivean, line 23: Fyi, you don't need state and effects for every part of the formula.
State is necessary only for the slider values (that user can change).
Every drived value can be calculated using statements and constants (without the effects).

ArtCraft: fixed

frontend/src/Components/Karma.tsx
```diff
@@ -0,0 +1,71 @@
...
+export function Karma() {
+    const [allPostsValue, setP] = useState(0); // sum of votes for all posts
+    const [allCommentsValue, setC] = useState(0); // sum of votes for all comments
+    const [profileVotesCount, setUCount] = useState(0); //count of votes in profile
+    const [profileVotingResult, setU] = useState(0); // sum of votes in profile
+    const [punishment, setPunishment] = useState(0); // penalty set by moderator
+
+    useEffect(() => {
+         setU( clamp(profileVotingResult, -profileVotesCount, profileVotesCount));
+    },[profileVotesCount]);
```
Aivean, line 12: Linter is correct, `profileVotingResult` should be also present in the effect dependency list.

frontend/src/Components/Karma.tsx
Aivean: Looks good to me.

frontend/src/Components/Karma.tsx
ArtCraft: done

mugabe (approved):

PR #90, opened 2022-07-02
Fixes an issue when right top menu is shifted when navigating

pazoozoo42:
Fixes an issue when right top menu is shifted when navigating between pages with scroll and pages with no scroll

Before:

https://user-images.githubusercontent.com/57953386/177014210-f3c07dd3-bd3c-4c55-a486-abb367bc25ce.mov

After:

https://user-images.githubusercontent.com/57953386/177014216-5bf9e824-094f-4978-ba2b-4ab437564858.mov

The shift actually happens because of [this CSS rule](https://github.com/spaceshelter/orbitar/blob/d4bd77cf67aee9f3f29aeeb5f7782a4e64793948/frontend/src/index.scss#L141), if you remove it the effect would be the same, but I'm not sure why it was defined

### DIFF
--- a/frontend/src/index.scss
+++ b/frontend/src/index.scss
@@ -2,6 +2,10 @@
 
 // TODO add css reset
 
+html {
+  overflow-y: overlay;
+}
+
 html,body {
   height: 100%;
   margin: 0;


mugabe:
Page scrolls to top when media upload popup opens with this fix

https://user-images.githubusercontent.com/475344/177031939-d8a5938c-9827-465f-bd7e-76db559664c9.mov

pazoozoo42:
> Page scrolls to top when media upload popup opens with this fix

nice catch! I guess this won't work then

PR #92, opened 2022-07-08
Hide media content in posts and comments with low rating

Aivean:
Current threshold for "low rating" status is set to "less than -3".

Demo:
orb-hidden

Height autocut in comment with low rating:
image

### DIFF
--- a/frontend/src/Components/CommentComponent.tsx
+++ b/frontend/src/Components/CommentComponent.tsx
@@ -88,7 +88,7 @@ export default function CommentComponent(props: CommentProps) {
                             <HistoryComponent initial={{ content, date: created }} history={{ id: props.comment.id, type: 'comment' }} onClose={toggleHistory} />
                         :
                             <div className={styles.content}>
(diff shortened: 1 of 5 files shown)


pazoozoo42:
looks awesome. The only thing I would change: maybe it would make sense to extract `-3` somewhere since it's used more than one time

pazoozoo42:
> this change would take more time and bloat the scope of this change, so I decided to postpone it.

Makes sense. But even if you just extract it as a constant available to both `CommentComponent.tsx` and `PostComponent.tsx`  it would help to avoid touching two files instead of one if this value is to be adjusted in the future

Aivean:
> > this change would take more time and bloat the scope of this change, so I decided to postpone it.
> 
> Makes sense. But even if you just extract it as a constant available to both `CommentComponent.tsx` and `PostComponent.tsx` it would help to avoid touching two files instead of one if this value is to be adjusted in the future

Extracted constants into the separate config. 

Also, slightly increased the hidden media size and reduced the blur effect:
orb-hidden2

PR #97, opened 2022-07-16
Initial elasticsearch commit

pazoozoo42:
- add elasticsearch to dev composer JSON
 - updated README describing an option to generate dummy content with real words
 - added backend `package.json` scripts to set up the index and index current posts and comments from mysql
 - added backend script to search and see search responses

### DIFF
--- a/backend/elasticsearch/index-current-data.js
+++ b/backend/elasticsearch/index-current-data.js
@@ -1,0 +1,79 @@
+require('dotenv-flow').config();
+const mysql = require('mysql2/promise');
+const { Client: Client7 } = require('es7');
+
+const db_config = {
+  host: process.env.MYSQL_HOST || "localhost",
+  port: parseInt(process.env.MYSQL_PORT) || 3306,
+  user: "root",
+  password: process.env.MYSQL_ROOT_PASSWORD || "orbitar",
+  database: process.env.MYSQL_DATABASE || "orbitar_db",
+  "multipleStatements": true
+}
+
+const per_page = 100;
+
+const elastic_client = new Client7({
+  node: 'http://localhost:9200'
+});
+
+async function sleep(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+function create_dataset_item(post_or_comment, is_post) {
+  return {
+    id: is_post ? 'p' + post_or_comment.post_id : 'c' + post_or_comment.comment_id,
+    site_id: post_or_comment.site_id,
+    author_id: post_or_comment.author_id,
+    created_at: Date.parse(post_or_comment.created_at) / 1000,
+    ...(is_post && { post_id: post_or_comment.post_id, title: post_or_comment.title }),
+    ...(!is_post && { comment_id: post_or_comment.comment_id }),
+    source: post_or_comment.source
+  };
+}
+
+async function main() {
+  const connection = await mysql.createConnection(db_config);
+  connection.config.namedPlaceholders = true;
+
+  ['posts', 'comments'].map(async (item_type) => {
+    const items_count = (await connection.query(`SELECT COUNT(*) AS num FROM ${item_type}`))[0][0].num;
+    for (let i = 0; i < items_count; i += per_page) {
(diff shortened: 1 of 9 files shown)


backend/elasticsearch/index-current-data.js
```diff
@@ -0,0 +1,79 @@
...
+              error: action[operation].error,
+              operation: body[iii * 2],
+              document: body[iii * 2 + 1]
+            })
+          }
+        })
+        console.log(`errors: `, erroredDocuments);
+      }
+
+      await sleep(1000);
```
Aivean, line 73: Is sleep necessary here?

backend/elasticsearch/index-current-data.js
```diff
@@ -0,0 +1,79 @@
...
+    site_id: post_or_comment.site_id,
+    author_id: post_or_comment.author_id,
+    created_at: Date.parse(post_or_comment.created_at) / 1000,
+    ...(is_post && { post_id: post_or_comment.post_id, title: post_or_comment.title }),
+    ...(!is_post && { comment_id: post_or_comment.comment_id }),
+    source: post_or_comment.source
+  };
+}
+
+async function main() {
```
Aivean, line 36: I think `migrations` is a better place for this script.

backend/elasticsearch/index-current-data.js
pazoozoo42: with current data I think it's not, but it's a good practice to have it between bulk inserts to just ensure Elasticsearch indexing queue doesn't get stuck. With current amount of posts it won't slow down the process either way

backend/elasticsearch/index-current-data.js
pazoozoo42: I thought about it, it's tricky since you need to run it after mysql migrations and after dummy content is generated. What would you recommend to avoid double call of indexing script?

backend/elasticsearch/index-current-data.js
```diff
@@ -0,0 +1,79 @@
...
+    const items_count = (await connection.query(`SELECT COUNT(*) AS num FROM ${item_type}`))[0][0].num;
+    for (let i = 0; i < items_count; i += per_page) {
+      const query = `SELECT * FROM ${item_type} ORDER BY created_at LIMIT ${i}, ${per_page}`;
+      let items_batch = (await connection.query(query))[0];
+      let bulk_dataset = [];
+
+      for (let ii = 0; ii < items_batch.length; ii++) {
+        let item = items_batch[ii];
+        bulk_dataset.push(create_dataset_item(item, item_type === 'posts'));
+        await sleep(100);
```
Aivean, line 50: This `sleep` is confusing. If it's purpose is to scale sleep time with the `items_batch.length`, then it think it would be cleaner to do that explicitly (e.g. `sleep(items_batch.length * 100)` ) in the `sleep` invocation below ↓

backend/elasticsearch/index-current-data.js
Aivean: Good point. After some thought I agree that manual invocation of the reindexing is the way to go.

Currently, we have three cases (that I can think of):

1. Existing site update (prod/staging deploy, local update)
    In this scenario the content is already in the DB and it needs to be indexed for first time (assuming there will be another mechanism for incremental indexing of the new/updated entities)

2. Bootstrapping the (local/test) site from scratch and filling it with dummy content.
   In this scenario the content is added in bulk after the DB migration and may need to be indexed (depending on whether elasticsearch is required to be in sync, which is not always the case for test env)

3. Recovering the site from the MySQL dump.
   This scenario is similar to the previous one (the elasticsearch won't be included in the dump and all existing content needs to be indexed, but elasticsearch index may not be empty)


I think the best way to support all three cases is to modify proposed reindexing script in the following way:
1. by default the script checks if elasticsearch index is empty and simply exits if it's not
2. script has an optional flag to clear the elasticsearch index OR there is a readme on how to do that manually

This way all three use cases are covered:
* by default developers and automated systems can safely invoke the reindexing script as the last setup step
* in the rare case when restoring from the MySQL backup over existing elasticsearch, the manual clearing will be required before reindexing

PR #102, opened 2022-08-08
Increase site domain and name length limits

pazoozoo42:
- increase site domain max length from 10 to 15
 - increase site name max length from 15 to 20
 - adjust backend and sidebar styles accordingly


frontend/src/Conf.ts
```diff
@@ -0,0 +1,13 @@
...
+const Conf = {
+    POST_LOW_RATING_THRESHOLD: -3,
+    COMMENT_LOW_RATING_THRESHOLD: -3,
+    SITE_DOMAIN_MIN_LENGTH_CHARS: 3,
+    SITE_DOMAIN_MAX_LENGTH_CHARS: 15,
+    SITE_NAME_MIN_LENGTH_CHARS: 3,
+    SITE_NAME_MAX_LENGTH_CHARS: 20,
+};
+
+export default Conf;
```
pazoozoo42, line 13: this is copied from another PR which is not yet merged into main, so probably would create a conflict during merge but should be easily resolved

Aivean (approved):

PR #103, opened 2022-08-09
In the mobile version, the screen area shifts to the right because of the iframe video

thinkjazz:
### How it was fixed
1. Into TheParse.ts into 223 line was added `return <div class="video-container"><iframe width="500" height="282" src="${encodeURI(embed)}" allowfullscreen class="youtube" frameborder="0"></iframe></div>;`
2. Added a style patch with rules for div="video-container", which is taken from the database, which was put there by the parser **(!MPORTANT This will only work for new video embeds from YouTubez For older posts that do not have a video class in the iframe and do not have a wrapper `<div="video-container"><iframe .......> </iframe></div>`  )** 
_So, there will be a need for migration! I had a weird side effect with the video stretching, but I couldn't reproduce it._
3. So far I haven't come up with anything smarter than importing styles into the PostComponent.tsx component
video-convert-1660006128000
full_hd
iphone
iphone_h
pixel


frontend/src/Components/PostContentVideo.css
```diff
@@ -0,0 +1,13 @@
...
+  height: 0;
+  padding-bottom: 56.25%;
+}
+.youtube {
+  position: absolute;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+}
```
pazoozoo42, line 13: minior code style: I'd separate rules and add a new line at the end of the file to get rid of Github warning:
```css
.video-container {
  position: relative;
  width: 100%;
  height: 0;
  padding-bottom: 56.25%;
}

.youtube {
  position: absolute;
  top: 0;
  left: 0;
  width: 100%;
  height: 100%;
}
```

pazoozoo42 (commented):
Looks good and definitely an important improvement for mobile users!

My only concern is that 100% would make videos way too big on large screens, take a look: 

image

Maybe there's a way to apply max-width for it and keep it responsive (so that it could get smaller) for small screens?

frontend/src/Components/PostContentVideo.css
thinkjazz: 1)100% yes maybe it's make sense!  for big screens make a less width video frame. But reducing the width`(width:90%) `rule entails that the block with the video and the block with the `textarea  `will be unequal and unbalanced.  And I would like to video equal to the size of the field text area) 
2) Done
Screenshot 2022-08-11 102527
Screenshot 2022-08-11 102601

backend/src/parser/TheParser.ts
```diff
@@ -221,6 +221,7 @@ export default class TheParser {
 
         // noinspection HtmlDeprecatedAttribute
         return `<iframe width="500" height="282" src="${encodeURI(embed)}" allowfullscreen frameborder="0"></iframe>`;
+
```
pazoozoo42, line 224: let's remove this newline so that this file stays untouched

pazoozoo42 (approved):
looks good to me, thanks

backend/src/parser/TheParser.ts
thinkjazz: done

Aivean (approved):

pazoozoo42 (approved):

PR #104, opened 2022-08-10
More self-regulation restrictions

Aivean:
### Changes

* renamed profile page "karma" -> "self-regulation"
* removed "karma calculator" -> "self-regulation formula calculator"
* invites are disabled when issuer's "self-regulation score" <= 0
* when "self-regulation score" < -10 karma votes by the user are revoked (triggered on certain actions)
* display active restrictions in profile
* minor change in `VoteRepository.ts`: add `voterId` and `userId` for all vote-returning queries
   *  and remove now unnecessary call to `UserManager.getByUsername` in `UserManager`

### Example restrictions displayed:
image

### Network error when checking an invite of the user with low "self-regulation score":
image

### Calculator after renaming:
image

### Testing:
* tested locally on recent dump
* tested in debugger

### DIFF
--- a/backend/src/api/InviteController.ts
+++ b/backend/src/api/InviteController.ts
@@ -70,6 +70,13 @@ export default class InviteController {
                 return response.error('invalid-code', 'Invite code not found or already used');
             }
 
+            const issuerRestrictions = await this.userManager.getUserRestrictions(invite.issued_by);
+            if (!issuerRestrictions.canInvite) {
+                this.logger.warn(`Invite issuer ${invite.issued_by} has no permission to invite`,
+                    {invite: code});
+                return response.error('invalid-code', 'Invite code can\'t be used');
+            }
+
             return response.success({
                 code: invite.code,
                 inviter: invite.issuer
--- a/backend/src/managers/UserManager.ts
+++ b/backend/src/managers/UserManager.ts
@@ -272,8 +272,7 @@ export default class UserManager {
 
         const activeKarmaVotes: Record<string, number> = {};
         for (const vote of userVotes) {
-            const user = await this.getByUsername(vote.username);
-            if (await this.isUserActive(user.id)) {
+            if (await this.isUserActive(vote.voterId)) {
                 activeKarmaVotes[vote.username] = vote.vote;
             }
         }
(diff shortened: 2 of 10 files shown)


backend/src/managers/UserManager.ts
```diff
@@ -284,9 +283,29 @@ export default class UserManager {
...
+        }
+        await this.redis.set(`remove_votes_when_karma_is_low_${userId}`, 'true');
+        await this.redis.expire(`remove_votes_when_karma_is_low_${userId}`,
+            60 * 30 + Math.floor(Math.random() * 60 * 5) /* 30-35 minutes */);
+    }
+
     async getUserRestrictions(userId: number): Promise<UserRestrictions> {
         const effectiveKarma = await this.getUserEffectiveKarma(userId);
 
+        if (effectiveKarma < -10) {
```
pazoozoo42, line 305: magic number, maybe worth extracting into a class constant or add a TODO for the future refactoring

backend/src/managers/UserManager.ts
```diff
@@ -302,6 +321,7 @@ export default class UserManager {
                 Math.max(lastCommentTime.getTime() / 1000 - Date.now() / 1000 + commentSlowModeDelay, 0) : 0,
             restrictedToPostId: (effectiveKarma <= -1000 ? (lastOwnPost?.post_id || true) : false),
             canVote: effectiveKarma >= -10,
+            canInvite : effectiveKarma > 0,
```
pazoozoo42, line 324: same as above, this and other numbers above can be extracted

backend/src/managers/types/UserInfo.ts
```diff
@@ -42,4 +42,5 @@ export type UserRestrictions = {
                                                   if true - restriction active, but no posts,
                                                   if false - no restriction */
     canVote: boolean; /* if user can vote for posts, comments and karma */
+    canInvite: boolean; /* whether invites by this work */
```
pazoozoo42, line 45:
```suggestion
    canInvite: boolean; /* whether invites by this user should work */
```

pazoozoo42 (approved):
Code looks good, only minor suggestions. Tested locally, looks good, but I'd change a couple of things:

1) need to hide these sub-headers if corresponding lists are empty. On the screenshot `1` belongs to the comments ratings list, but "Posts" header is confusing: 

image

2) Do we want these micro fraction values? Maybe worth thinking about rounding it to some point: 

image

3) I would change this red background color or text color, it's barely readable especially when there's a link: 

image

backend/src/managers/UserManager.ts
Aivean: At this moment all “magic constants” are neatly contained inside a single method, if they were defined in a different place and used here (imo) it would only hurt the readability. But I agree that when (if) these constants start being used elsewhere, they should be moved into a centralized place or config.

backend/src/api/InviteController.ts
```diff
@@ -70,6 +70,13 @@ export default class InviteController {
                 return response.error('invalid-code', 'Invite code not found or already used');
             }
 
+            const issuerRestrictions = await this.userManager.getUserRestrictions(invite.issued_by);
+            if (!issuerRestrictions.canInvite) {
+                this.logger.warn(`Invite issuer ${invite.issued_by} has no permission to invite`,
+                    {invite: code});
+                return response.error('invalid-code', 'Invite code can\'t be used');
```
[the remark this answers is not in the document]
Aivean: I started addressing it and realized that for consistency all similar constant codes should be extracted as well, but it will introduce a bit of a scope creep. 

So I decided to create an issue so we don't forget about this: #105

Aivean:
@pazoozoo42 , I addressed the UX issues (although I tried to minimize the effort spent on that, as this page is mostly for internal usage and debug and will be thrown away in the future).

Changes:
image
image
image
image

pazoozoo42 (approved):

PR #110, opened 2022-08-14
Another bunch of self-regulation improvements

Aivean:
### UI:
* restyle the self-regulation page in user profile
* hide the detailed calculator by default

Own profile:
image

Other user's profile:
image

Noob:
image

First "tier" of restrictions:
image

Max restrictions:
image

### Added new restrictions:
* Can't create subsites when s.r.score <= 0
* Can't edit own content when s.r. is at its minimum (-1000)


Aivean:
As discussed, added following changes:
* user is considered "new" when registered less than 3 days ago
* separated permissions to vote for content and user
* can use invites only when s.r. score >= 2 and when user is new
* can vote for karma when s.r. score >=2 and not new (existing karma votes are revoked only when s.r. score < -10)

When karma voting disabled:
image

Noob:
image

Naughty boi:
image


Royal flush:
image

backend/src/api/PostController.ts
```diff
@@ -126,11 +126,21 @@ export default class PostController {
...
+            if (!restrictions.canEditOwnContent) {
+                post.canEdit = false;
+            }
 
             let comments: CommentEntity[] = [];
             if (!noComments) {
                 const rawComments = await this.postManager.getPostComments(postId, userId, format);
+
+                if (!restrictions.canEditOwnContent) {
+                    rawComments.forEach(comment => comment.canEdit = false);
```
pazoozoo42, line 141: `postManager.getPostComments` is already iterating comments, it might be more efficient to pass `canEdit` as an argument to this method and set this value there to avoid additional iteration

backend/src/managers/UserManager.ts
```diff
@@ -74,7 +74,7 @@ export default class UserManager {
         return user;
     }
 
-    async getFullProfile(username: string, forUserId: number): Promise<UserProfile | undefined> {
+    async getFullProfile(username: string, forUserId: number): Promise<UserInfo | undefined> {
```
pazoozoo42, line 77: minor: maybe rename it to `getFullUserInfo` if return type is changed from `UserProfile` to `UserInfo`?

frontend/src/Components/UserProfileKarma.tsx
```diff
@@ -60,23 +70,73 @@ export const UserProfileKarma = (props: UserProfileKarmaProps) => {
...
+            <div>
+                Нет права приглашать других людей на орбитар.
+                {restrictionsResult.effectiveKarma >= 0 && <>
+                   <p>Для новичков это нормально, они получают это право после того, как начитают писать хорошие посты и комменты.</p>
+                </>}
+            </div>
+        </>,
+
+        restrictionsResult.commentSlowModeWaitSec > 0 &&
+        <><span className={'i i-slow'}></span>Право комментировать ограничено одним комментарием в {formatTimeSec(restrictionsResult.commentSlowModeWaitSec)}.</>,
```
pazoozoo42, line 85: this probably won't happen in real life but `formatTimeSec` needs to be fixed to use another case:

image

frontend/src/Components/UserProfileKarma.tsx
```diff
@@ -60,23 +70,73 @@ export const UserProfileKarma = (props: UserProfileKarmaProps) => {
...
+                <div className={styles.beta}>BETA</div>
+                <p>Мы продолжаем работать над саморегуляцией.</p>
+                <p>Это самая первая версия, работающая по механизму,
+                <PostLink post={{id: 781, site: 'dev'}}> описанному тут</PostLink>.</p>
+                <p>Формула саморегуляции зависит от двух вещей — оценок другими людьми вас и вашего контента (постов и комментариев).
+                    Если коротко, ведите себя по-человечески, производите хороший контент, и все будет хорошо.</p>
+            </div>}
+
+            {!restrictionsResult && <div>Загрузка...</div>}
+            {!!restrictions?.length && <div>{restrictions.map((r, idx) =>
```
pazoozoo42, line 136: I think `!!` is not needed here unless I'm missing something

pazoozoo42 (dismissed):
Tested locally with latest backup,looks good! Only a couple of minor suggestions

backend/src/api/PostController.ts
Aivean: Iteration is cheap compared to DB and and json serialization. I wouldn't worry about that (don't think it's worth adding extra responsibility to `getPostComments`)

frontend/src/Components/UserProfileKarma.tsx
Aivean: React renders `{0}` as "0",  while `{false}` is not rendered.  `0 && something` evaluates to `0`.

frontend/src/Components/UserProfileKarma.tsx
Aivean: How to fix: https://github.com/moment/moment/blob/develop/locale/ru.js#L24

Aivean: image

backend/src/managers/UserManager.ts
Aivean: Renamed to `getByUsernameWithVote` (also fixed some caching issues).

Aivean:
Found a minor bug with karma voting restriction affecting post and comment voting (UI-only).

Aivean:
Bunch of minor fixes:
* fix the bug with disabled voting in UI
* rewording of inactive tooltip:
image
* restrict some invites API methods
*

Aivean:
Ok, I couldn't help myself and added the ability to set karma penalty (via redis entry).

How:
redis key: `karma_penalty_${userId}` value: `numeric_penalty`
```
127.0.0.1:6379> SET karma_penalty_11 2000
OK
```

Low penalty:
image


High penalty:
image

pazoozoo42 (approved):

PR #118, opened 2022-08-18
fix unique keys error

pazoozoo42:
Fixes runtime error on user's invites page
image

### DIFF
--- a/frontend/src/Components/UserProfileInvites.tsx
+++ b/frontend/src/Components/UserProfileInvites.tsx
@@ -46,17 +46,15 @@ export const UserProfileInvites = () => {
                 ?
                     <>
                         {activeInvites.map(invite =>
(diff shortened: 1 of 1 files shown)


frontend/src/Components/UserProfileInvites.tsx
```diff
@@ -46,8 +46,8 @@ export const UserProfileInvites = () => {
                 ?
                     <>
                         {activeInvites.map(invite =>
-                            <>
-                                <div key={invite.code} className='item'>
+                            <React.Fragment key={invite.code}>
```
Aivean, line 49: Why do we even wrap the div in the fragment? I think fragment can be removed and key added to the div.

pazoozoo42: makes sense, removed

Aivean (approved):

PR #119, opened 2022-08-19
Fix: local caching of the user profile prevented up-to-date karma value from being displayed

Aivean:
#110 broke the karma value displayed in user profile.

How to reproduce the bug:
* open user profile
* change karma
* refresh the page
* incorrect (old) karma value is displayed

After this change karma value is fetched directly from DB.


pazoozoo42 (approved):

Aivean:
superseded by #120

PR #122, opened 2022-08-24
chore/extract strings from the controllers

Alao-Abiodun:
### What does the PR do ?
- created a new constants files and extracted the repeated strings from the controllers `UserController.ts` and `InviteController.ts` into constants files as CONSTANT variable names in order to reuse them in any place of the codebase.


backend/package.json
```diff
@@ -4,10 +4,10 @@
   "description": "",
   "main": "dist/app.js",
   "scripts": {
-    "start:dev": "npx tsc && NODE_ENV=development node build/app.js",
+    "start:dev": "npx tsc && SET NODE_ENV=development node build/app.js",
```
Aivean, line 7: Please exclude changes to the `package.json` from the PR, it will break the linux workflow, and it's outside of the scope.

backend/src/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const ERROR_CODES = {
```
Aivean, line 1: I think the better location for this file could be:
`backend/src/api/utils/error-codes.ts`

backend/package-lock.json
```diff
@@ -1,8600 +1,8 @@
 {
   "name": "backend",
   "version": "1.0.0",
```
Aivean, line 3: Please don't commit `package-lock.json` (both for front and back), since dependencies were not changed.

Aivean (changes requested):
Thank you for your contribution!

Overall, the PR looks good, except a few minor things, please check the review comments.

backend/package.json
Alao-Abiodun: Alright, Thanks

> Please exclude changes to the `package.json` from the PR, it will break the linux workflow, and it's outside of the scope.

backend/src/constants.ts
Alao-Abiodun: Ok, I will refractor it to the path and change file name.

> I think the better location for this file could be: `backend/src/api/utils/error-codes.ts`



> I think the better location for this file could be: `backend/src/api/utils/error-codes.ts`

backend/package-lock.json
Alao-Abiodun: Ok, Thanks

Alao-Abiodun:
I have sent a new PR to the same branch. Thanks.

backend/package-lock.json
Aivean: Could you please make sure that `package-lock.json` is not modified? If you have trouble doing that I could do that myself tomorrow. Thanks.

Alao-Abiodun: Thanks, @Aivean So I noticed it was when I was doing the npm install setup it change the lockfileversion from 2 to 1. So I just reverted only the package-lock.json file for both frontend and backend back to the initial state.

Alao-Abiodun:
Hello @Aivean should I correct this error on the same branch that I am still working on? I just checked the codebase and I missed the `;` semi-colon after the close bracket.

Alao-Abiodun:
I have fixed the semi-colon error.

Aivean (approved):

Aivean:
@Alao-Abiodun , merged. Thank you for your contribution!

PR #123, opened 2022-08-24
Option to choose between feed sorting

pazoozoo42:
Option to choose between feed sorting:
LIVE (sorted by post last comment ts) or NEW (sorted by sort creation ts)


backend/src/api/types/requests/FeedSortingSave.ts
```diff
@@ -0,0 +1,8 @@
+import {FeedSorting} from '../../../db/types/FeedSortingSettings';
+
+export type FeedSortingSaveRequest = {
+    'site-id': number;
+    'feed-sorting': FeedSorting;
+};
+
+export type FeedSortingSaveResponse = {};
```
mugabe, line 8: `{}` in type declaration means any object

backend/src/db/repositories/PostRepository.ts
```diff
@@ -105,7 +106,7 @@ export default class PostRepository {
                 left join user_bookmarks b on (p.post_id = b.post_id and b.user_id=:user_id)
                 left join post_votes v on (v.post_id = p.post_id and v.voter_id=:user_id)
             order by
-                p.commented_at desc
+                p.${sortBy === FeedSorting.postCommentedAt ? 'commented_at' : 'created_at'} desc
```
mugabe, line 109: Please add index for `created_at desc`. Also I noticed that index for `commented_at` also missing 😱

backend/src/db/repositories/UserRepository.ts
```diff
@@ -37,6 +38,18 @@ export default class UserRepository {
         });
     }
 
+    async getFeedSortingSettingsByUserId(userId: number): Promise<FeedSortingSettingsRaw[] | undefined> {
+        return await this.db.query<FeedSortingSettingsRaw[]>('select * from feed_sorting_settings where user_id=:user_id', {user_id: userId});
+    }
+
+    async saveFeedSorting(siteId: number, feedSorting: FeedSorting, userId: number) {
+        await this.db.query(`replace into feed_sorting_settings (user_id, site_id, feed_sorting) values (:user_id, :site_id, :feed_sorting)`, {
```
mugabe, line 46: It's not that important here, just advice.
I personally prefer `insert ... on duplicate key update ...`.
`replace into` removes existing record instead of updating existing one, that is a bit slower.

frontend/src/API/FeedAPI.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export default class FeedAPI {
+    api: APIBase;
+
+    constructor(api: APIBase) {
+        this.api = api;
+    }
+
+    saveSorting(siteId: number, feedSorting: FeedSorting): Promise<unknown> {
+        return this.api.request<FeedSortingSaveRequest, unknown>('/feed/sorting', {
```
mugabe, line 17: I think we know that `unknown` 😀

mugabe (changes requested):

backend/src/api/types/requests/FeedSortingSave.ts
```diff
@@ -0,0 +1,8 @@
+import {FeedSorting} from '../../../db/types/FeedSortingSettings';
```
mugabe, line 1: Don't use non-API types in API types.
You should duplicate type or you can do reverse thing – use API type in non-API types.

Reason for this – I want to extract API types declarations into separate library to reuse it in backend and frontend.

backend/src/api/types/requests/FeedSortingSave.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export type FeedSortingSaveRequest = {
+    'site-id': number;
```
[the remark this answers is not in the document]
mugabe: Also I suggest to use subdomain name instead of id.

This is debatable question, but right now we use subdomain in other API methods.

pazoozoo42:
converting to draft because code review changes will take some time

backend/src/db/repositories/PostRepository.ts
pazoozoo42: both `commented_at` and `created_at` are used in multi-field indexes ([commented_at](https://github.com/spaceshelter/orbitar/blob/42ebfd368935144ed943f5fa327a4cc1efb5bb21/mysql/schema.sql#L97), [created_at](https://github.com/spaceshelter/orbitar/blob/c668cb29684c654a898ef4b84dcae3284d09e29e/backend/migrations/20220511160411-posts-add-author-created-index.js#L18)), which should work when sorting. I could be wrong though, but when I did explain it showed these indexes as possible to use

pazoozoo42:
@mugabe @Aivean ready for another look, I addressed all code review comments except [adding an index](https://github.com/spaceshelter/orbitar/pull/123#discussion_r953660486)

backend/src/api/FeedController.ts
```diff
@@ -48,6 +49,10 @@ export default class FeedController {
             perpage: Joi.number().min(1).max(50).default(10),
             format: joiFormat
         });
+        const feedSortingSchema = Joi.object<FeedSortingSaveRequest>({
+            site: Joi.string().regex(/^[a-z\d-]{3,10}$/i).required(),
```
mugabe, line 53: We can change it in the future, it would be better to move regexp to constant (see bottom of ApiMiddleware.ts).

backend/src/api/FeedController.ts
```diff
@@ -59,20 +64,19 @@ export default class FeedController {
         this.router.post('/feed/all', validate(feedSubscriptionsSchema), (req, res) => this.feedAll(req, res));
         this.router.post('/feed/posts', validate(feedPostsSchema), (req, res) => this.feedPosts(req, res));
         this.router.post('/feed/watch', validate(feedWatchSchema), (req, res) => this.feedWatch(req, res));
-        this.router.post('/feed/sorting', saveFeedSortingLimiter, (req, res) => this.saveFeedSorting(req, res));
+        this.router.post('/feed/sorting', saveFeedSortingLimiter, validate(feedSortingSchema), (req, res) => this.saveFeedSorting(req, res));
     }
 
-    async saveFeedSorting(request: APIRequest<FeedSortingSaveRequest>, response: APIResponse<FeedSortingSaveResponse>) {
+    async saveFeedSorting(request: APIRequest<FeedSortingSaveRequest>, response: APIResponse<Record<string, unknown>>) {
```
mugabe, line 70: Add declaration to `FeedSortingSave.ts`
`type FeedSortingSaveResponse = Record<string, unknown>`

backend/src/api/FeedController.ts
```diff
@@ -59,20 +64,19 @@ export default class FeedController {
...
         if (!request.session.data.userId) {
             return response.authRequired();
         }
 
         const userId = request.session.data.userId;
 
         try {
-            await this.userManager.saveFeedSorting(request.body['site-id'], request.body['feed-sorting'] as FeedSorting, userId);
+            await this.userManager.saveFeedSorting(request.body['site'], request.body['feedSorting'] as FeedSorting, userId);
```
mugabe, line 78: It's valid way to access field, but seems a bit odd to me, let's use
`request.body.site`, `request.body.feedSorting`

backend/src/db/utils/Feed.ts
```diff
@@ -0,0 +1,9 @@
+import {FeedSortingSettingsBySite} from '../types/FeedSortingSettings';
+import {FeedSorting} from '../../api/types/entities/common';
+
+export function getFeedSortingBySite(feedSortingSettings: FeedSortingSettingsBySite, site: string): FeedSorting {
```
mugabe, line 4: Still don't see any reason why it's global.

mugabe (changes requested):
Nothing critical, but some cosmetic changes required.

pazoozoo42:
per discussion with @mugabe we need to change the way feed sorting settings are provided to the UI, will need to move this from UserInfo to feed response

PR #94, opened 2022-07-10
Implement the backend for "effective user karma" calculation 

Aivean:
#### Summary of the changes:
* table `user_site_rating` that stores the rating of user's content (posts and comments) per subsite
* user activity calculation and caching in the redis
* `/user/karma` endpoint that returns user stats
* .../u/username/karma page that shows karma calculator with actual data
* user profile shows if user visited site in the previous week
* setVotes implementation that updates aggregate tables

#### Karma stats page (UI requires some polishing @ArtCraft)
image

#### Testing:
* manual testing for correctness
* concurrent stress test of the new `setVotes` (no deadlocks, but only 16-19rps locally)

PR #136, opened 2022-08-29
Restricted general access to invites

Aivean:
As discussed, the idea is to allow users to have general access to invites (with some restrictions). To prevent runaway effect all newly invited users will have the `ontrial` flag, that restricts their rights to karma voting and inviting new users, until several conditions are not met:
* at least three days passes since registration
* certain content rating threshold is met (s.r. score >= 2, see #94, #89)
* enough karma votes are accumulated (which directly and indirectly cover a significant percentage of the community, more on that later)

Invite accessibility rules:
* every user with full rights can have at least one invite
* for every active and "successful" "child" parent gets one additional invite per wait period (initially the period is one week)
* every "child" rejected by the community extends the waiting period by one week (i.e. if you invited three people who were rejected, you'll be getting 1 invite per month).

Initial implementation:
* add DB support
* add `ontrial` flag for the user object
* karma voting for the user with `ontrial` status removes ontrial status when the majority consensus is reached
* user profile shows progress to the full rights, invite reason and the list of trial voters


frontend/src/Components/Username.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  fill: var(--primary);
+  opacity: .7;
+}
+
+.username:hover svg {
+  color: var(--primaryHover);
+}
+
+.username:hover svg {
+  fill: var(--primaryHover);
```
pazoozoo42, line 21: I think it should be possible to simplify
```css
.username:hover svg {
  color: var(--primaryHover);
  fill: var(--primaryHover);
}
```

frontend/src/Pages/UserPage.tsx
```diff
@@ -60,57 +49,64 @@ export const UserPage = observer(() => {
...
+                        <div className={styles.registered}>#{user.id},
+                            {profile.invitedBy && <>
+                                <a href={`/u/${profile.invitedBy.username}/invites/#${user.username}`} title={'Детальный контекст приглашения'}>приглашен</a>
+                                <Username user={profile.invitedBy} /></> || <span>зарегистрирован</span>}
+                             <DateComponent date={user.registered} />
+                        </div>
                         {profile.invites.length > 0 && <div>
-                            По его приглашениям зарегистрированы: {profile.invites.map((user, idx) => {
-                                return <Username key={idx} user={user}/>;
+                            Пригласил: {profile.invites.map((user, idx) => {
```
pazoozoo42, line 100: is it possible to user user gender here for dynamic label?

pazoozoo42 (approved):
Just minor questions and suggestions.

I would also do the following:
On the `/profile/invites` page don't make invite URL a link, but replace it with copying on click functionality. And emphasize with CSS that invite is being regenerated and invite code is changed when they click "Отозвать" button

backend/src/managers/InviteManager.ts
```diff
@@ -58,4 +70,89 @@ export default class InviteManager {
...
+            };
+            if (!user.ontrial && await this.userManager.isUserActive(user.id) && (await getRestrictions()).canVoteKarma) {
+                activeIntegratedUsers++;
+            } else if (!(await getRestrictions()).canVote) {
+                rejectedUsers++;
+            }
+        }
+
+        /* each rejected user increases invite waiting period by 1 week */
+        let inviteWaitPeriodDays = 7 * (1 + rejectedUsers);
```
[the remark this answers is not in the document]
Aivean: I don't like extracting constants when they are not reused. This constant is already extracted in the scope it belongs to.

frontend/src/Pages/UserPage.tsx
Aivean: Done.

backend/src/managers/InviteManager.ts
```diff
@@ -58,4 +70,89 @@ export default class InviteManager {
...
+        let invitesPerPeriod = 1 + activeIntegratedUsers;
+
+        /* simplify fraction */
+        const gcdResult = gcd(1 + rejectedUsers, invitesPerPeriod);
+        inviteWaitPeriodDays /= gcdResult;
+        invitesPerPeriod /= gcdResult;
+
+        const msInAnHour = 60 * 60 * 1000;
```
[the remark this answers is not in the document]
Aivean: I usually extract repetitive code, even when it's self-explanatory.

Aivean:
@pazoozoo42 
> I would also do the following:
> On the `/profile/invites` page don't make invite URL a link, but replace it with copying on click functionality. 

Funny thing about that, that's how it would work on prod when the clipboard API is present. The bug was in `e.preventDefault();` call after the return if the clipboard API is not available.

> And emphasize with CSS that invite is being regenerated and invite code is changed when they click "Отозвать" button

Good idea, however, with the current implementation it's a headache to add. The outcome is not worth the time atm (IMO).

pazoozoo42 (dismissed):

Aivean:
Rebased onto dev to resolve conflicts.

PR #138, opened 2022-09-01
Fix development environment setup

Lexy2:
Override `REACT_APP_API_DOMAIN` var in `.env.development` so that it doesn't get set by default `.env`


pazoozoo42 (approved):
Thanks, this fixes the issue

PR #139, opened 2022-09-04
Feed sorting

pazoozoo42:
As discussed, this is changed to return feed sorting in the feed payload
[initial PR](https://github.com/spaceshelter/orbitar/pull/123) for the reference


backend/src/api/SiteController.ts
```diff
@@ -51,7 +51,7 @@ export default class SiteController {
             bookmarks: Joi.boolean().default(false)
         });
         const siteCreateSchema = Joi.object<SiteCreateRequest>({
-            site: Joi.string().regex(/^[a-z\d-]{3,10}$/i).required(),
+            site: joiSite.required(),
```
Aivean, line 54: Can reuse `joiSite` validator for `site` field in the other API methods (above).

backend/src/managers/FeedManager.ts
```diff
@@ -25,7 +26,7 @@ export default class FeedManager {
         this.redis = redis;
     }
 
-    async getSubscriptionFeed(forUserId: number, page: number, perpage: number, format: ContentFormat): Promise<PostInfo[]> {
+    async getSubscriptionFeed(forUserId: number, page: number, perpage: number, format: ContentFormat, sorting: FeedSorting): Promise<PostInfo[]> {
         const limitFrom = (page - 1) * perpage;
 
         // check if fanned out
```
Aivean, line 32: Need to add another flag, `subscriptions:${forUserId}:created_at:fanned` that corresponds to the completion of the fanning out of the new redis feed cache. The key that corresponds to the requested `sorting` method should be checked, and if it's not present, `await this.pullSubscriptionsFeed(forUserId);` should be called.

frontend/src/Pages/FeedPage.tsx
```diff
@@ -37,7 +39,9 @@ export default function FeedPage() {
     const perpage = 20;
     const page = parseInt(search.get('page') || '1');
 
-    const { posts, loading, pages, error, updatePost } = useFeed(site, feedType, page, perpage);
+    const [changeSorting, setChangeSorting] = useState(-1);
```
Aivean, line 42: Nit: seems that `-1` doesn't have any special connotation here, probably better to replace it with `undefined`.

frontend/src/Pages/FeedPage.tsx
```diff
@@ -53,12 +57,26 @@ export default function FeedPage() {
         document.title = docTitle;
     }, [siteInfo, feedType]);
 
+    const handleFeedSortingChange = async (e: React.MouseEvent) => {
+        e.preventDefault();
+        const feedSorting = (e.currentTarget as HTMLAnchorElement).dataset.feedSorting;
```
Aivean, line 62: Is there a reason to read DOM attributes here? Seems like an easier way would be to pass the sorting value as a parameter to the function:

```
const handleFeedSortingChange = (value: FeedSorting) => async (e: React.MouseEvent) => {
 ...
}

<a href='#' className={sorting === FeedSorting.postCommentedAt ? styles.active : ''} onClick={handleFeedSortingChange(FeedSorting.postCommentedAt)}>LIVE</a>
```

Aivean (changes requested):
Overall, looks good.

However, currently the feed with the new sorting method will be empty for existing users (that have their feed already materialized). Need to address that.

Aivean (approved):
I fixed my own nits and tested it locally.

@ArtCraft, please check it locally as well, but I think it's ready to be merged.

pazoozoo42:
@Aivean thank you, appreciate your help

Aivean (approved):
Reviewed and checked locally.

Btw, we can make this loading progress bar global (add it to mobx appState) and reuse it in other places.

PR #141, opened 2022-09-05
Reset password page

pazoozoo42:
Uses sendinblue.com API which is free for 300 emails per day
This is basic functionality only, I didn't encrypt email addresses in the db for now


backend/src/api/AuthController.ts
```diff
@@ -28,15 +36,87 @@ export default class AuthController {
             legacyHeaders: false
         });
 
+        // 2 attempts per minute should be enough
+        const resetPaswsordLimiter = rateLimit({
+            windowMs: 60 * 1000,
+            max: 100, // todo: put 2 instead of 100
```
Aivean, line 42: forgot TODO

pazoozoo42: changed to 5 requests per hour

pazoozoo42: also pushed some renamings

backend/src/db/repositories/UserRepository.ts
```diff
@@ -25,6 +26,53 @@ export default class UserRepository {
         return await this.db.fetchOne<UserRaw>('select * from users where username=:username', {username: username});
     }
 
+    // TODO: need to decide if we want to encrypt email addresses
+    async getUserByEmail(email: string): Promise<UserRaw | undefined> {
+        return await this.db.fetchOne<UserRaw>('select * from users where email=:email', {email});
+    }
+
+    async generateAndSavePasswordResetForUser(userId: number): Promise<string | undefined> {
+        const code = crypto.createHash('md5').update(Math.random().toString() + userId).digest('hex');
```
iamwavecut, line 35:
```suggestion
        const code = crypto.createHash('sha256').update(Math.random().toString() + userId).digest('hex');
```

Индустрия давно отказалась от md5 и гонит поганой метлой sha1, так что, во избежание будущих сломов предлагаю золотой стандарт текущего дня.

Aivean (changes requested):
One potential issue I see with this implementation is following: since codes are stored in raw format, if database leaks, the accounts that ever used password recovery would be compromized.

I suggest to address it in the following way:
1. instead of the raw code, store the encrypted hash of the code (use bcrypt with salt)
   * instead of the code send base64 encoded username+code
   * when resetting password, lookup code hashes by username and try all of them
2. remove expired codes from the DB before checking for existing codes and resetting password (to make sure expired coded cannot be used), and the expiration period could be much less than 1 day: I think two hours would be enough

pazoozoo42:
> ince codes are stored in raw format, if database leaks, the accounts that ever used password recovery would be compromized.

@Aivean not really, codes are stored for one hour only and also all user codes are cleared if one of them is used. I think this approach is safe

iamwavecut (approved):

iamwavecut:
> One potential issue I see with this implementation is following: since codes are stored in raw format, if database leaks, the accounts that ever used password recovery would be compromized.
> 
> I suggest to address it in the following way:
> 
> 1. instead of the raw code, store the encrypted hash of the code (use bcrypt with salt)
>    
>    * instead of the code send base64 encoded username+code
>    * when resetting password, lookup code hashes by username and try all of them
> 2. remove expired codes from the DB before checking for existing codes and resetting password (to make sure expired coded cannot be used), and the expiration period could be much less than 1 day: I think two hours would be enough

I think that most of it are already redundant security that overcomplicates things a lot. If an attacker has access to our hashes he would just generate a new reset request and fetch and exploit it.

iamwavecut (dismissed):

Aivean:
> If an attacker has access to our hashes he would just generate a new reset request and fetch and exploit it.

The access to what hashes? In the scheme I described even if attacker has realtime read-only access to the DB it would be impossible for them to use password reset to get access to an account, if they don’t have access to the account’s email.

However, that brings another security concern about the sessions ids that are stored in the DB. Switching to jwt will solve that, however before that we might need another way to ensure that sessions ids are not stored in DB in raw format.

backend/src/api/AuthController.ts
```diff
@@ -28,15 +36,87 @@ export default class AuthController {
             legacyHeaders: false
         });
 
+        // 5 requests per hour
+        const resetPaswsordLimiter = rateLimit({
```
Aivean, line 40: I think we need to limit by key (in this case — the email parameter). The rationale for that: currently a malicious party can DDOS this endpoint with reset requests containing arbitrary emails, and that will prevent legitimate users from resetting passwords.

Another, better solution would be to add limiter only for the failed requests (see sign-in for example), and fail request (in controller/manager) in the following cases:
* when requested email is not found in DB
* when new request comes too soon (we store the date of the last generated request)
* when too many active codes are stored in DB

Aivean:
@pazoozoo42 , I’m sorry for being picky, but this API is public and security-critical, and I believe we ought to make it reasonably secure right from the start.

pazoozoo42:
> @pazoozoo42 , I’m sorry for being picky, but this API is public and security-critical, and I believe we ought to make it reasonably secure right from the star

no hard feelings at all, we are trying to build something fast and secure at the same time. Thank you for you feedback

> Currently a malicious party can DDOS this endpoint with reset requests containing arbitrary emails, and that will prevent legitimate users from resetting passwords.

this is not the case: emails are only sent for those email addresses that exist in the db. Arbitrary email just won't work. To exploit the way you are describing this user should be sending emails to their own email address. This is easily detectable, but I agree that we can add a check for this. 
Since we store reset code generation time we can easily throttle successful requests for a user, say, 3 per day. Let me know what you think.

As a side note, I agree with you that security is important and one of the top priorities (after smooth user experience). At the same time, we should keep balance and avoid over-engineering to not to re-invent the wheel, this password resetting workflow has proven to be secure in many other projects

Aivean:
> this is not the case: emails are only sent for those email addresses that exist in the db. Arbitrary email just won't work. To exploit the way you are describing this user should be sending emails to their own email address. This is easily detectable, but I agree that we can add a check for this.

Currently the limiter limits all incoming requests. If you just ddos the endpoint with random emails, you saturate the limiter and endpoint will become inaccessible for legitimate users. I'm not talking about actually sending emails here.

Aivean (dismissed):
Tested locally, I think it's good enough as it is.

The only UX nit I have — perhaps it would be better to validate new password as it is being typed instead of onSubmit, as it could be frustrating for the user to enter illegal (weak) password twice only to discover that it doesn't pass validation.

Aivean:
Improved logging a bit.

```js
logger.error('Could not send email', {result});
```
produces:
```
[error  ] MAIL     Could not send email {"result":{"code":"unauthorized","message":"Key not found"}}
```

and

```js
try {
  throw new Error('Failed to send email');
} catch (e) {
  logger.error(e);
}
```
produces:
```
[error  ] MAIL     Failed to send email {"stack":"Error: Failed to send email\n    at /Users/Aivean/...Mailer.js:51:23\n    at Generator.next (<anonymous>)\n    at fulfilled (/Users...Mailer.js:5:58)\n    at processTicksAndRejections (node:internal/process/task_queues:94:5)"}
```

Aivean (approved):

backend/src/utils/Mailer.ts
```diff
@@ -28,7 +28,7 @@ export async function sendResetPasswordEmail(username: string, email: string, co
                     <p>Привет, ${username}!</p>
                     <p>Чтобы сбросить пароль, пройдите по этой ссылке: ${(siteConfig.http ? 'http://' : 'https://') + siteConfig.domain}/forgot-password/${code}</p>
                     <p>Ссылка активна только 3 часа</p>
-                    <p>Если ничего не получается, можете написать в <a href="https://t.me/orbitar_bot">телеграм боту</a></p>
+                    <p>Если ничего не получается, можете написать нам в <a href="https://discord.gg/j4tevpV9">Дискорд</a></p>
```
Aivean, line 31: Not sure if this is a good idea to post the discord invite on the (semi-) public page. Of course the invite can be invalidated, but still.

Aivean: Why not create and use a separate email, like support@orbitar.space ?

frontend/src/Pages/ResetPasswordPage.tsx
```diff
@@ -39,17 +44,18 @@ export default function ResetPasswordPage() {
             });
     }, [code, api]);
 
+    const onPasswordStrengthUpdate = useCallback((newStrength: PasswordStrength | undefined) => {
```
Aivean, line 47: What's the reason behind wrapping a state setter in useCallback?

Aivean:
Changed the confirmation wording a bit, and moved the confirmation checkbox into, what I believe is a more reasonable place:
image

Aivean:
And I think we still need to disallow the weakest passwords. It's not only the question of user safety, but also the question of the site safety. Why open ourselves to obvious exploits by allowing users to have ridiculous 1-letter passwords?

frontend/src/Pages/ResetPasswordPage.tsx
pazoozoo42: removed it, I was troubleshooting something and kept it like that, but it's redundant

Aivean (dismissed):

Aivean:
Changed hint a bit:
image

Aivean (approved):

Issue #131, opened 2022-08-27
Parser improvement — trim line breaks


Aivean:
When new post/comment is posted, it's often the case that there are extra line breaks at the beginning and at the end:

![orb_demo](https://user-images.githubusercontent.com/2865203/187047454-21647559-c748-4e30-a347-5a03dfe46e8c.gif)

Need to improve `TheParser` to trim empty lines.

PR #142, opened 2022-09-14
Parser improvements related to line break stripping

Aivean:
Changes
* strip line breaks at the beginning and the end of the text
* strip extra line breaks after the blockquote

Small demo:
orb-demo-bq

Testing:
* unit tests
* manual

### DIFF
--- a/backend/src/parser/TheParser.ts
+++ b/backend/src/parser/TheParser.ts
@@ -46,6 +46,9 @@ export default class TheParser {
     }
 
     parse(text: string): ParseResult {
+        // strip newlines from the beginning and end
+        text = text.replace(/^[\r\n]+|[\r\n]+$/g, '');
+
         const doc = this.parseDocument(text, {
             decodeEntities: false
         });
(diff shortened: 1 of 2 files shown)


Aivean:
Fix #131

backend/src/parser/TheParser.ts
```diff
@@ -54,14 +57,25 @@ export default class TheParser {
...
-        return doc.reduce((p, c) => {
-            const res = this.parseNode(c);
+        const p = {text: '', mentions: [], urls: [], images: []};
+        let prevBQ = false; // if previous node was blockquote
+        for (let node of doc) {
+            if (prevBQ && node.type === 'text') {
+                // remove a single newline after blockquote, allow only a single one if multiple were present
+                node = {
+                    ...node,
+                    data: node.data.replace(/^(\r?\n|[\r\n])(\r?\n|[\r\n])?(\r?\n|[\r\n])*/, '$2')
```
pazoozoo42, line 67: I think this *or* condition is redundant here: `|[\r\n]`, and if I change pattern to 
```js
data: node.data.replace(/^(\r?\n)(\r?\n)?(\r?\n)*/, '$2')
```
tests are passing too

pazoozoo42 (approved):
looks good, one minor suggestion to simplify the regex

backend/src/parser/TheParser.ts
Aivean: Sure, but it doesn't account for `\r` line break type (not sure if we need to support it, but we do in other places).
https://en.wikipedia.org/wiki/Newline

Aivean: And now that I looked into it, my regex doesn't account for `\n\r` 🤷‍♂️

megamk: > Sure, but it doesn't account for \r line break type

maybe add tests with `\r` then

pazoozoo42: > Sure, but it doesn't account for \r line break type 

maybe add tests with `\r` then

PR #143, opened 2022-09-14
fix #121 posts without read comments not showing in watch

### LINKED ISSUE #121: Inconsistency in the number of posts on the "watch" page
image

People are reporting "fire" showing a number of unread posts, while the "unread" page doesn't show anything. 
Probably an error in SQL. Need to reproduce.

Aivean:
Changes:
* made  `orbitar_db.user_bookmarks.read_comments` non-null

related: #100

Testing:
* manual:
   * used a fresh user to add a post with lots of comments to watched
   * observed the comments number appear near the "fire" and the post appear in "feed/watch/all",  but not in "feed/watch"
   * applied the fix
   * observed post appear also in the "feed/watch"

### DIFF
--- a/backend/migrations/20220913233142-user-bookmarks-read-comments-not-null.js
+++ b/backend/migrations/20220913233142-user-bookmarks-read-comments-not-null.js
@@ -1,0 +1,36 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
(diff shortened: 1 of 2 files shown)


pazoozoo42 (approved):
it just works 👍

PR #145, opened 2022-09-19
Feed improvements: subsite change and ability to show post on main

Aivean:
### Changed:

* allow to select and edit subsite of the post
* a checkbox to show post on main in addition to the selected subsite
* allow to subscribe and unsubscribe from main

### Implementation changes:
* improved performance of postFanOut a bit
* refactored

### Current look:
image

image

### TODO:
- [ ] add styles to the new components
- [ ] show subsite and checkbox status in the post history


Aivean:
This should be reworked a bit after #173 is merged (the implementation will be much simpler).

PR #149, opened 2022-09-24
set mysql to image instead of build context

Lexy2:
Replaced custom mysql context with a build script with MySQL image with an initdb script.

Starts faster, doesn't require build. Starting from scratch would require just to delete the _mysql_db_ volume, no need to rebuild the image with `--no-cache`.

Now tested with dev, properly publishes basic + MySQL X ports (3306 & 33060).

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -16,16 +16,17 @@ services:
 
   mysql:
     platform: linux/amd64
-    build:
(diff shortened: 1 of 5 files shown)


Aivean:
Tested locally (with and without volume removal).

Aivean (approved):

PR #153, opened 2022-09-25
Fix changing of the feed sorting on subsites

Aivean:
Issue description:
On the subsites that were not previously visited, it was impossible to change sorting (button didn't do anything).

The problem was in misalignment between two states and their default values. I simplified the whole thing to have a single state.


pazoozoo42 (approved):
thanks for digging into this!

PR #158, opened 2022-10-03
DB actions

Lexy2:
Added 2 workflows:
- Backup DB
- Restore Sanitized DB.

## Backup DB
Dumps the databases daily at 01:20 UTC daily using `mysqldump`, zips the dump with a password and uploads it as an artifact.

**Requires**:
- create _Production_ environment in _Settings > Environments_
- create two environment secrets:
  - `MYSQL_ROOT_PASSWORD` - obviously, the root password of the production MySQL server
  - `BACKUP_PASSWORD` - the password to encrypt the ZIP file with
- configure GitHub runner on the production server (pending from me & @Aivean)

## Restore Sanitized DB
Downloads the latest backup at 01:50 UTC every Monday created by the _Backup DB_ workflow, restores it on a random MySQL server, removes the PII, dumps the updated DB, and uploads it as an artifact, encrypted by a password.

**Requires:**
- Backup workflow to run at least once
- create the 3rd environment secret in _Production__:
  - STAGING_BACKUP_PASSWORD - the password to encrypt the sanitized ZIP for Staging environment
 - create two environment secrets in _Staging_:
  - `MYSQL_ROOT_PASSWORD` - obviously, the root password of the production MySQL server
  - `BACKUP_PASSWORD` - the password to decrypt the ZIP encrypted with `STAGING_BACKUP_PASSWORD`

### DIFF
--- a/.github/workflows/restore-sanitized-db.yml
+++ b/.github/workflows/restore-sanitized-db.yml
@@ -1,0 +1,102 @@
+name: Restore Sanitized DB
+
+concurrency: stage
+
+on:
+  workflow_dispatch:
+  workflow_call:
+  schedule:
+    - cron: '50 1 * * MON' # At 01:50 UTC every Monday
+
+jobs:
+  sanitize-db:
+    runs-on: ubuntu-latest
+    environment: Production
+
+    env:
+      MYSQL_ROOT_PASSWORD: root
+
+    services:
+      mysql:
+        image: mysql:5.7
+        ports:
+          - "3306:3306"
+        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=10s --health-retries=10
+
+    steps:
+      - name: Download DB backup
+        uses: dawidd6/action-download-artifact@v2
+        with:
+          workflow: backup-db.yml
+          workflow_conclusion: success
+          check_artifacts: true
+
+      - name: Unzip artifact & load DB
+        env:
+          BACKUP_PASSWORD: ${{ secrets.BACKUP_PASSWORD }}
+        run: |
+          cd `ls -t | head -n 1`
+          unzip -P $BACKUP_PASSWORD `ls -t | head -n 1`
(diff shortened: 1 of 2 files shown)


Aivean:
I suggest to switch "Backup DB" to staging as a first step to safely verify that the whole (backup -> restore) pipeline works.

Aivean:
Production environment created, two secrets added to both environments.

Aivean (approved):

.github/workflows/restore-sanitized-db.yml
```diff
@@ -0,0 +1,102 @@
...
+          BACKUP_PASSWORD: ${{ secrets.BACKUP_PASSWORD }}
+        run: |
+          cd `ls -t | head -n 1`
+          unzip -P $BACKUP_PASSWORD `ls -t | head -n 1`
+          cat `ls -t *.sql | head -n 1` | /usr/bin/mysql -h 127.0.0.1 -u root -p$MYSQL_ROOT_PASSWORD
+
+      - name: Remove PII
+        run: |
+          cat <<EOF | /usr/bin/mysql -h 127.0.0.1 -u root -p$MYSQL_ROOT_PASSWORD --database orbitar_db
+          delete from invites;
```
Aivean, line 45: I'll add a separate sql file to do that (it will be a bit more involved than that), but this works as a first version.

Lexy2:
> I suggest to switch "Backup DB" to staging as a first step to safely verify that the whole (backup -> restore) pipeline works.

I have tested this in my fork by assigning Production environment to the staging server runner.

.github/workflows/restore-sanitized-db.yml
Lexy2: Sounds good, why is it more involved?

Aivean: One example: ideally we don't want to remove invites, but rather randomize their hashes (since removing invites breaks some features).

Also, we need to clear `sessions` table as well. 

In the future, a standalone SQL file makes more sense from the organizational standpoint.

PR #159, opened 2022-10-06
Fix comment jumping off viewport when post has unsized images

4vanger:
I needed to sacrifice smooth scroll in order to keep anchor on desired comment.
Browser provides "overflow-anchor" property to control this behavior but algorythm is quite fragile and smooth scrolling was failing to work with it - in order to be anchored the element needs to be fully visible in a viewport.
Code will apply "focusing" class on the post which will disable scroll anchoring
Instead of center of the page focused comment will appear just below the topbar where it can be easily spotted.
Testing: 
* manually load the post with unsized random image in a post and comments body and verified that it is still gets anchored on selected comment on page refresh
* click "go to parent" link
* open post with `isNew=true`

Update 29.12.22: final version - code is very clean and organized. From my tests it works pretty good - the only case when it still misbehave is when parent's comment also have unsized image - in this case it will shift the layout but not more than one comment down so it is pretty easy to find. This is because parent comment is also visible in the viewport (because we need to shift page down so topbar won't cover the comment) and we cannot disable scroll-anchoring for parent element because it will also disable it for the comment we are trying to focus to.

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "dist/app.js",
   "scripts": {
-    "start:dev": "npx tsc && node build/app.js",
+    "start:dev": "npx tsc && NODE_ENV=development node build/app.js",
     "build": "npx tsc",
     "start:production": "NODE_ENV=production node build/app.js",
     "migration:dev": "NODE_ENV=development node migrations.js",
--- a/backend/package.json
+++ b/backend/package.json
@@ -4,7 +4,7 @@
   "description": "",
   "main": "dist/app.js",
   "scripts": {
-    "start:dev": "npx tsc && NODE_ENV=development node build/app.js",
+    "start:dev": "npx tsc && node build/app.js",
     "build": "npx tsc",
     "start:production": "NODE_ENV=production node build/app.js",
     "migration:dev": "NODE_ENV=development node migrations.js",
--- a/frontend/src/Components/CommentComponent.module.scss
+++ b/frontend/src/Components/CommentComponent.module.scss
@@ -1,4 +1,5 @@
 .comment {
+  overflow-anchor: none;
 
     // TODO fix alternative tree lines
     /*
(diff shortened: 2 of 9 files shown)


Aivean (changes requested):
Unfortunately, still doesn't work when images above the target comment are loaded. To reproduce: disable cache in Chrome, use the link to the comment on the very bottom of the post.

Additionally, the latest update broke the blinking effect for the selected comment. Now the whole tree blinks.

Aivean:
I'm not sure if you tried the following case:
* during the scroll:
* the target comment has the default `overflow-anchor` setting
* everything else has `overflow-anchor: none;`

If that doesn't work than I believe the only way to fix it currently is to retry the scroll if the viewport position of the target comment changed during the scroll (check it periodically).

frontend/src/Components/CommentComponent.module.scss
```diff
@@ -1,4 +1,5 @@
 .comment {
+  overflow-anchor: none;
```
Aivean, line 2: I don't think this is a good idea to make this a default. Instead, I'd suggest creating a separate style `.no-anchor` and adding and removing it as needed before and after scroll. The style could be applied even to some common ancestor of `.comment` and `.post`.

4vanger:
Oki, final version - really like it how it works - minimal async clutter and code duplication.
When focusing - attach a classname to Post which will disable anchoring. Wait until page is fully loaded before removing classname to reenable anchoring.
Tested it locally a lot - works pretty great for me.

backend/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "dist/app.js",
   "scripts": {
-    "start:dev": "npx tsc && NODE_ENV=development node build/app.js",
+    "start:dev": "npx tsc && node build/app.js",
```
Aivean, line 7: Is this intentional?

4vanger: sorry, accidentally commited - doesn't work on Windows otherwise. will revert

Aivean: > doesn't work on Windows otherwise
that's weird let's follow up

Aivean:
Preliminary testing went flawless. Will test a bit more tomorrow. Awesome job!

Aivean (approved):
LGTM

PR #160, opened 2022-10-06
Prevent RatingSwitch buttons from stealing focus

4vanger:
Problem:
1) click + or - button on post/comment
2) press "Space"
3) (even worse) scroll down few screens without clicking using mouse/touchpad
Expected behavior: page should scroll down one screen
Actual behavior: button is clicked again undoing the vote. Even more frustrating is button is not visible on a screen anymore

Fix: unfocus rating buttons after click

### DIFF
--- a/frontend/src/Components/RatingSwitch.tsx
+++ b/frontend/src/Components/RatingSwitch.tsx
@@ -89,33 +89,39 @@ export default function RatingSwitch(props: RatingSwitchProps) {
     }, [showPopup, ratingRef, popupRef, votes, state.vote, props.id, props.type]);
 
     const handleVote = (vote: number) => {
-        const prevState = { ...state };
-        if (state.vote === vote) {
-            vote = 0;
-        }
-
-        const newRating = state.rating - (state.vote || 0) + vote;
-
-        setState({ vote: vote, rating: newRating});
-        if (props.onVote) {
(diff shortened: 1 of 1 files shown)


frontend/src/Components/RatingSwitch.tsx
```diff
@@ -89,33 +89,39 @@ export default function RatingSwitch(props: RatingSwitchProps) {
...
-        if (props.onVote) {
-            props.onVote(newRating, vote);
-        }
-
-        api.voteAPI.vote(props.type, props.id, vote)
-            .then(result => {
-                setState({
-                    rating: result.rating,
-                    vote: result.vote
+        return (ev: React.MouseEvent) => {
```
Aivean, line 92: nit: could be simplified as  
`(vote: number) => (ev: React.MouseEvent) => { ... }`

Aivean (approved):
Checked locally, works as expected.

frontend/src/Components/RatingSwitch.tsx
4vanger: too hard to read to my taste

Aivean: Dunno what you are talking about, looks like idiomatic Haskell ;p

PR #161, opened 2022-10-07
Backup images and videos from comments and posts

pazoozoo42:
`npm run backup-media` will start the process. Script downloads images and videos from comments and posts and creates `index.csv` file with metadata


Lexy2 (dismissed):
Worth checking the checksum of the file contents rather than the URL, to prevent duplicate files from different sources.

backend/backup_media.js
```diff
@@ -0,0 +1,151 @@
...
+    loader.get(url, function(response) {
+      response.pipe(file);
+      file.on('finish', function() {
+        file.close(cb);
+      });
+    }).on('error', function(err) {
+      console.trace(err);
+      fs.unlinkSync(dest);
+      if (cb) cb(err.message);
+    });
```
Aivean, line 47: I think you need to call `.end()` to send the request (not sure if it's absence cause any actual issues).

backend/backup_media.js
```diff
@@ -0,0 +1,151 @@
...
+      if (fs.existsSync(outputPath)) {
+        console.log(outputPath, 'is already there');
+        saveEntry();
+        continue;
+      }
+      download(mediaSrc, outputPath, async function() {
+        saveEntry();
+        console.log('Downloaded:', itemType, itemId, mediaType, mediaSrc, config.dir + '/' + outputFilename, dimensions.width, 'x', dimensions.height);
+      });
+      await sleep(500);
```
Aivean, line 143: Do I understand correctly, that the main loop just spawns download jobs every 500 ms without waiting for their completion? Is that intentional?

backend/backup_media.js
```diff
@@ -0,0 +1,151 @@
...
+  database: config.database
+});
+
+const entries = [];
+
+connection.query(
+  'SELECT html, "comment" as type, comment_id as "item_id" FROM `comments` WHERE `html` like "%<img%" UNION ' +
+  'SELECT html, "comment" as type, comment_id as "item_id" FROM `comments` WHERE `html` like "%<video%" UNION ' +
+  'SELECT html, "post" as type, post_id as "item_id" FROM `posts` WHERE `html` like "%<img%" UNION ' +
+  'SELECT html, "post" as type, post_id as "item_id" FROM `posts` WHERE `html` like "%<source%"',
```
Aivean, line 74:
```suggestion
  'SELECT html, "post" as type, post_id as "item_id" FROM `posts` WHERE `html` like "%<video%"',
```

backend/backup_media.js
```diff
@@ -0,0 +1,151 @@
...
+const connection = mysql.createConnection({
+  host: config.host,
+  user: config.user,
+  password: config.password,
+  database: config.database
+});
+
+const entries = [];
+
+connection.query(
```
Aivean, line 70: Fetching everything each run is very expensive (especially using text-base queries).

To run this in production we need to store the update timestamp of the latest processed entity and start the next batch from there. Also fetch and process entities in chunks.

Aivean (changes requested):

Aivean (changes requested):
I'm also thinking about the way of how to run this periodically in production.

If invoked as runner workflow, it needs a way to either connect to the mysql docker container (currently port is not exposed in prod).

Alternatively this could be packed as a separate container in the docker-compose config and scheduled to invoke periodically.

backend/backup_media.js
pazoozoo42: > I think you need to call .end() to send the request (not sure if it's absence cause any actual issues).

Not sure I understand. Do you mean `http.get().send()`? If yes, then no, it's not needed

backend/backup_media.js
pazoozoo42: yes, but this is not on purpose. I'll make it synchronous

Aivean:
> > Alternatively this could be packed as a separate container in the docker-compose config and scheduled to invoke periodically.
> 
> I'll go that way, thanks

I'm not sure that that's the best way to go though.

It's relatively easy to implement, but it has disadvantages compared to the GHA workflow:
* more coupling to the server
* less visibility and control
* changing requires redeployment of the whole app

I think the GHA workflow route is better. DB queries could be run via docker mysql shell ([see this for example](https://github.com/spaceshelter/orbitar/actions/runs/3216795897/workflow#L99)), results exported as files.

pazoozoo42:
@Aivean as we discussed, I changed mysql entries to be fetched in batches and made downloading synchronous. 
I think there's no need to save last processed entry, because if image is already there it's not downloaded for the second time, only `index.csv` entry is added. At the same time if someone changes their comment or post, processing all entries would help to catch that and would download newer file

> Worth checking the checksum of the file contents rather than the URL, to prevent duplicate files from different sources.

@Lexy2 it could help but we cannot get the file checksum without downloading it, so I think it's not needed

backend/backup_media.js
```diff
@@ -0,0 +1,155 @@
...
+}
+
+(async () => {
+  const result = await connection.promise().query(
+    'select count(*) as total from `content_source` where `source` like "%<img%" or `source` like "%video%"'
+  );
+  console.log('Content sources with media: ', result[0][0].total);
+  console.log('Start processing');
+  await processBatch(0);
+  process.exit(0);
```
Aivean, line 151: Instead of terminating process, please add the `finally` block and call `await conn.end();` there.

backend/backup_media.js
```diff
@@ -0,0 +1,155 @@
...
+  for (let i = 0; i < result[0].length; i++) {
+    let entry = result[0][i];
+    let itemType = entry.ref_type;
+    let itemId = entry.ref_id;
+    let entrySource = entry.source;
+    const mediaItems = entrySource.match(/<(img|source).+src=['"]([^'"]+)['"]/g);
+    if (!mediaItems) {
+      continue;
+    }
+    for (let ii = 0; ii < mediaItems.length; ii++) {
```
Aivean, line 106: Since the index is not used, both for loops can be replaced with [`for of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of).

backend/backup_media.js
```diff
@@ -0,0 +1,155 @@
...
+          saveEntry(mediaType, outputPath, outputFilename, mediaSrc, itemType, itemId);
+          console.log('Downloaded: ', mediaSrc);
+        }
+      } catch (e) {
+        console.error('Failed to download: ', mediaSrc);
+      }
+    }
+  }
+  await sleep(SLEEP_BETWEEN_BATCHES_MS);
+  await processBatch(offset + BATCH_SIZE);
```
Aivean, line 141: I think it's more natural to return the number of processed items or a boolean "has next batch" flag from the `processBatch` method, and have the outer loop outside of the `processBatch` method.

megamk: debatable, this will require saving total number into a var and comparing it with a number of processed entries which would need to be compared with the total

pazoozoo42: debatable, this will require saving total number into a var and comparing it with a number of processed entries which would need to be compared with the total

Aivean: I mean, if `processBatch` returned `result[0].length` instead of calling `await processBatch(offset + BATCH_SIZE);` the "main" loop could look like this:

```js
let offset = 0
while (true) {
   const processed = await processBatch(offset);
   offset += processed
   if (processed < BATCH_SIZE) break;
}
```

The advantage of that is making `processBatch` have clearly defined responsibility: "process **batch**", instead of "process everything till the end".

pazoozoo42: this makes sense, I'll adjust

backend/backup_media.js
```diff
@@ -0,0 +1,171 @@
...
+      }
+    }
+  }
+  return result[0].length;
+};
+
+(async () => {
+  try {
+    const result = await connection.promise().query(
+      'select count(*) as total from `content_source` where (`source` like "%<img%" or `source` like "%video%") and date(created_at) > "' + START_DOWNLOADING_FROM_DATE + '" '
```
Aivean, line 150: `content_source.source` doesn't work here, it contains "source" text before parsing, meaning img and video could be present there as raw urls.

Traversing data in the order of  `content_source.content_source_id`  increase is a good idea, since the latest `content_source_id` could be saved as an offset for the next processing.

I'd suggest to rewrite the query as follows (pseudocode):
```sql

select
cs.content_source_id, coalesce(p.html, c.html) as html
( -- inner query:
   select countent_source_id from content_source 
   where  content_source_id > $OFFSET
   order by content_source_id asc
    limit $BATCH_SIZE 
) cs 
  left join posts p on content_source_id 
  left join comments c on content_source_id

where coalesce(p.html, c.html) like '%img/video%'
```

backend/backup_media.js
```diff
@@ -0,0 +1,171 @@
...
+      return false;
+    }
+    if (url.match(/^\/\//)) {
+      url = 'https:' + url;
+    }
+    try {
+      const response = await fetch(url, { signal: controller.signal });
+      const responseBinary = await response.buffer();
+      let writeResult = await fs.promises.writeFile(dest, responseBinary);
+      await sleep(1000);
```
Aivean, line 41: Why is sleep necessary here?

backend/backup_media.js
```diff
@@ -0,0 +1,171 @@
...
+  user: config.user,
+  password: config.password,
+  database: config.database
+});
+
+try {
+  fs.mkdirSync(config.dir);
+} catch (e) {}
+if (!fs.existsSync(config.dbFile)) {
+  fs.openSync(config.dbFile, 'w+');
```
Aivean, line 72: No need to open file when not using the returned handle.
```suggestion
```

backend/backup_media.js
```diff
@@ -0,0 +1,171 @@
...
+  fs.appendFileSync(config.dbFile, 'item_type,item_id,media_type,media_src,media_id,img_width,img_height  ' + '\n');
```
Aivean, line 73: In my local tests `appendFileSync` failed to create file when file doesn't exist (contrary to the documentation).
```suggestion
  fs.writeFileSync(config.dbFile, 'item_type,item_id,media_type,media_src,media_id,img_width,img_height  ' + '\n');
```

backend/backup_media.js
```diff
@@ -0,0 +1,171 @@
...
+} else {
+  fs.openSync(config.dbFile, 'w+');
+}
```
Aivean, line 76: No need to open file when not using the returned handle.
```suggestion
}
```

backend/backup_media.js
```diff
@@ -0,0 +1,171 @@
...
+      }
+      offset += processed;
+      await sleep(SLEEP_BETWEEN_BATCHES_MS);
+    }
+  } catch (e) {
+    console.log(`error happened`, e);
+  } finally {
+    console.log(`done`);
+    connection.end();
+    process.exit(0);
```
Aivean, line 169: Redundant exit.
```suggestion
```

Aivean (approved):
Looks good! Thanks a lot! Finally we can stop worrying about our media hosting going down.

I'll merge this and create a PR with a workflow wrapper.

PR #167, opened 2022-10-13
Use prettier for automatic code formatting

4vanger:
Changes in this PR:
1) Add root package.json and listed frontend and backend as workspaces. This should decrease number of files in FS and `npm install` will update both apps
2) added top `start` script which conceurrently running frontend and backend with single command
3) added import-sort and prettier for automatic code formatting.
4) automatically formatted PostPage.tsx as an example (too bad GH diff didn't pickup formatting changes) - no logic changes, just formatting.

What I encourage you to do:
1.  here is the list of [Prettier config options](https://prettier.io/docs/en/options.html)
1.  checkout branch locally
1.  install Prettier plugin for your IDE and enable running it on file save
1.  modify a single character and save file - Prettier will format the file
1. Check the formatting. If you don't quite like it - customize .prettierrc.yml (here is the list of [Prettier options](https://prettier.io/docs/en/options.html)). goto 4.
1. When satisfied - post your config on [Discord thread](https://discord.com/channels/964881133118578708/1017022629191884800) and let's discuss it.


Aivean (commented):
Let's postpone this until current major features are merged to avoid merge conflicts.

Alternatively you may consider configuring the new formatter to be more lenient and allow current formatting style.

frontend/src/Pages/PostPage.tsx
```diff
@@ -1,135 +1,182 @@
-import React, {useEffect, useState} from 'react';
-import styles from './PostPage.module.css';
-import {Link, useLocation, useParams, useSearchParams} from 'react-router-dom';
-import {CommentInfo, PostInfo, PostLinkInfo} from '../Types/PostInfo';
-import PostComponent from '../Components/PostComponent';
+import React, { useEffect, useState } from 'react';
```
Lexy2, line 1: I think this PR doesn't belong here

docker-compose.dev.yml
```diff
@@ -10,7 +10,7 @@ services:
       SERVER_DOMAIN: ${SERVER_DOMAIN}
       IMGUR_CLIENT_ID: ${IMGUR_CLIENT_ID}
     volumes:
-      - $PWD/caddy/Caddyfile.dev:/etc/caddy/Caddyfile
+      - ./caddy/Caddyfile.dev:/etc/caddy/Caddyfile
```
Lexy2, line 13: And this change doesn't have to be here either.

frontend/src/Pages/PostPage.tsx
4vanger: That's just an example of how formatted file will look like

docker-compose.dev.yml
4vanger: yeah - rigth

4vanger:
It will always produce conflicts so yes - we need to postpone it and migrate whole codebase at once

.prettierrc.yml
```diff
@@ -0,0 +1,5 @@
+printWidth: 100
+semi: true
+singleQuote: true
+trailingComma: all
+parser: typescript
```
Lexy2, line 5: I insist on tabWidth: 4

4vanger: why not use tab symbols and let everyone set preferred value in their IDE?
(let the holy war begin!)
With prettier it is rather simple - it will insert all the missing spaces. I don't like tabs because you can click "inside" of the tab and then you need to waste time moving the cursor

frontend/src/Pages/PostPage.tsx
Aivean: I'd suggest to use gists for this purpose. [Example.](https://gist.github.com/Aivean/c7e2c1a3037ab8b22e16714d2b73c27a/revisions)

.prettierrc.yml
Aivean: Let's use the current value of 4 spaces as a compromise.

.eslintrc
```diff
@@ -0,0 +1,24 @@
+{
```
Aivean, line 1: We already have .eslintrc in the `frontend` and `backend`. Repo root is not a js/ts project, so I think it makes more sense to keep per-module conf.

PR #170, opened 2022-10-27
Deepl.com translation for posts and comments

Aivean:
### Demo:
ezgif-1-f83d52e4ae

### Implementation:
* Translation using deepl.com
   * Note: root `.env` (`backend/.env.development`) now expects `DEEPL_API_KEY` 
* translation is cached per `content_source_id` in `translations` table
* source language of posts and comments is detected using [Fasttext library](https://fasttext.cc/) (wasm version)
* existing Ukrainian comments are marked as such based on the presence of `Є` and `Ї`

Additionally:
* fixed an existing bug when post title is not written to the history

TODO:
- [x] copy translation button styles from comment into post


backend/src/db/repositories/TranslationRepository.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export default class TranslationRepository {
+    private db: DB;
+
+    constructor(db: DB) {
+        this.db = db;
+    }
+
+    async getTranslation(content_source_id: number, language: string): Promise<TranslationRaw | undefined> {
+        return await this.db.fetchOne<TranslationRaw>('select * from translations where content_source_id=:content_source_id and language=:language', {
```
pazoozoo42, line 12: is it possible to have multiple translations for a given content source item? If not, comparing `language` is redundant

backend/src/managers/TranslationManager.ts
```diff
@@ -0,0 +1,137 @@
...
+import fetch from 'node-fetch';
+import PostRepository from '../db/repositories/PostRepository';
+import {ContentSourceRaw} from '../db/types/ContentSourceRaw';
+import TheParser from '../parser/TheParser';
+import {stripHtml} from 'string-strip-html';
+import {Logger} from 'winston';
+import {addOnPostRun, FastText, FastTextModel} from '../../langid/fasttext.js';
+import {urlRegex} from '../parser/urlregex';
+
+const fasttextModelProm: Promise<FastTextModel> = new Promise<FastText>((resolve) => {
```
pazoozoo42, line 12: minor, I'd rename to `fasttextModelPromise`

pazoozoo42 (dismissed):
Works and looks good! Couple of minor comments, and also I'd extract basic language value (`ru` in our case) into `.env` so that it would be easily configurable for users who might want to have another basic language in their installation

backend/src/db/repositories/TranslationRepository.ts
Aivean: Not at the moment, but the db design (composite key: content_source_id, language) allows to add this fairly easy. For example, if we decide to allow users to select their preferred translation language in the future.

But I agree, it's redundant at the moment, not sure if it's worth deleting though (somewhat invasive change).

Aivean:
@pazoozoo42, addressed review comments.

pazoozoo42 (approved):

Issue #44, opened 2022-05-12
Design change proposal:  don't use redis for feed generation


Aivean:
### Current design

* Redis zset is used to store materialized feed (post ids) for every user. 
* When a new comment is posted, all users that are subscribed to the corresponding subsite and have their feed materialized have their feed updated.
* When new user request the feed for the first time, or when they change their subscriptions, the feed has to be regenerated (basically, all posts must be fetched and reinserted into redis).

### Problems

* Posting a comment currently is a relatively expensive operation (proportional to the number of users, subscribed to the site).
* Changing subscription is also an expensive operation, proportional to the number of posts in all subsites user is subscribed to.
* Consistency is an issue:
  * if failure (network, process) occurs in the middle of the feed update, the feed will be corrupted
* If more criteria are required to build user feed (e.g. post rating filter), feed update for all subscribed users will have to be performed on every post vote

### Proposal

Use the mysql query to fetch user feed. With right indexes it's rather fast (actually, post data fetching is more expensive operation).

```sql
SELECT site_id FROM user_sites WHERE user_id = ?;
SELECT post_id from posts where site_id in (?) order by commented_at desc limit ?, ?;
```

Following indices are required for best performance:
```sql
create index user_sites_user_id_index on user_sites (user_id);
create index posts_commented_at_site_id_index on posts (commented_at desc, site_id asc);
```

### Load test

I created a simple tool that tests concurrent requests performance using the above queries #43.

Here are the results:

```
# With both indexes:

Running 100 concurrent users for 20 seconds
Limit: 100 posts per page
Max page: 8
Retrieve all: false
Total requests: 8059
Requests per second: 402.95

Running 100 concurrent users for 20 seconds
Limit: 100 posts per page
Max page: 8
Retrieve all: true
Total requests: 956
Requests per second: 47.8

Running 1 concurrent users for 20 seconds
Limit: 100 posts per page
Max page: 8
Retrieve all: false
Total requests: 5304
Requests per second: 265.2

Running 1 concurrent users for 20 seconds
Limit: 100 posts per page
Max page: 8
Retrieve all: true
Total requests: 947
Requests per second: 47.35

Running 100 concurrent users for 20 seconds
Limit: 100 posts per page
Max page: 0
Retrieve all: false
Total requests: 10984
Requests per second: 549.2


# without posts_commented_at_site_id_index index:

Running 100 concurrent users for 20 seconds
Limit: 100 posts per page
Max page: 8
Retrieve all: false
[...]

PR #173, opened 2022-11-07
Replace materialized redis feed cache per user with a dedicated in-memory index

Aivean:
### Rationale
As described in #44 , the current materialized feed implementation doesn't work well for our use case. It's inflexible (hard to add new feeds, change implementation around post-site relationships, etc), memory consuming, has consistency problems and issues with restoration from the backup.

Recently we've seen CPU spikes in prod, whenever new comment is added:
image

Presumably, initial feed creation (for new users) and subscription to subsites is even more CPU intensive.

### Solution
Create a dedicated in-memory index of posts per key (key roughly maps to subsite):

```
key1:  (time0, postA),          (time3, postX),  (time5, postZ)
key2:           (time2, postP),                              (time6, postO)
key3:                    (time2.5, postY)
```
Index allows efficient queries of the type:  give `limit` posts for the given set of keys/subsites with the given `offset`.

In other words, if you have  `Arr0`, `Arr1`, `Arr2` of posts from `Key0`, `Key`, `Key2`, where each post has `Timestamp` and `Id`, the job of the query can be described as:
* merge  `Arr0`, `Arr1`, `Arr2` into a single array `Tmp`
* sort `Tmp` by `Timestamp` of the posts
* skip `offset` largest items
* return `limit` items

Note: that's not the actual implementation, see below.

Additionally, index should allow efficient updates.

### Data structure/algorithm

Each index (per key) consists of three data structures:
* [BTree](https://en.wikipedia.org/wiki/B-tree) of tuples `(Timestamp, PostId)`
* [FenwickTree](https://en.wikipedia.org/wiki/Fenwick_tree) of post numbers per day
* Hash map of `PostId` -> `Timestamp`   (needed for updates)

#### Update:
* Add post to hashmap and btree
* increment frequency in the FenwickTree corresponding to the day of the post's timestamp

#### Query:
Reminder: query returns `limit` posts with maximal timestamp after `offset` for given set of keys/subsites.

* find the list of Indexes that correspond to the given keys
* given offset
   * binary search over the day range of the timestamps to find such a **day** that the total number of posts on and after that day for **all** keys is less or equal to `offset`
       * to calculate the total number of posts on and after the given day, iterate over list of indexes and do fenwick tree query for each index
   * after **day** is determined, find the range of keys before that day in Btree per key and use [k-way merge](https://en.wikipedia.org/wiki/K-way_merge_algorithm) (priority queue of B-tree iterators) to skip the remaining `offset` (if needed) and then return the needed `limit` posts

### Asymptotic complexity analysis:

* Query:
  `O(K log D +  limit * log(P) * log(K)`
   where:
    * `K` is number of keys/indexes in the query (number of subsites in the subscription) 
    * `D` is our data range in days  (3650  for 10 years)
    *  `limit` is limit (e.g. 20)
    *  `P` is the number of posts in each index/key

* Add/Update/Remove (per post and key):
  `O( log(D) + log(P) )`

### Implementation

The index is implemented as a separate REST microservice in Rust, packaged as Docker container.
Index is cleaned and re-populated every time on server startup (takes just a couple of seconds).

Considered alternatives: 
* Rust FFI or [Neon](https://neon-bindings.com/)
   * downsides: 
       * need to compile for each platform, need to have `rustc` installed on dev machines, complicates project setup for developers
       * more technically and operationally complex
* Writing everything in Typescript/Webassembly:
   * downsides: synchronous, slower, blocks main thread

The tradeoff of the current approach is added latency (the query itself is submillisecond, but the HTTP request and json ser/deser adds several **milliseconds** per query). However, this doesn't really matter, as HTTP requests are asynchronous and fully parallel.

### Performance

Tested on Apple M1.

Query with offset=0, limit=20, total number of posts =10M, date range =10years, query has 200 subsites/keys:
```
bench_zero_offset   ... bench:      28,242 ns/iter (+/- 426)
```

Query with random offset, limit=20, total number of posts =10M, date range =10years, query has **500** subsites/keys:
```
bench_random_offset ... bench:     491,429 ns/iter (+/- 60,715)
```

End-to-end http request to the microservice takes 6-9ms on my local setup.

### Additional changes

* Feed page: fetch posts from DB using a single request, instead of doing that one by one
* Update `user_bookmarks` with a single request (I believe old implementation was buggy)
   Happens when new comment is posted.

### TODO:
- [x] Ensure feed service is ready on backend start


backend/src/config.ts
```diff
@@ -5,6 +5,7 @@ type Config = {
     redis: RedisConfig;
```
Lexy2, line 5: Why are we still keeping this?

docker-compose.local.yml
```diff
@@ -26,7 +26,7 @@ services:
     volumes:
       - mysql_db:/var/lib/mysql
       - $PWD/mysql/schema.sql:/docker-entrypoint-initdb.d/schema.sql:ro
-  
+
   redis:
```
Lexy2, line 30: And this then?

feed/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# Tells docker to use the latest Rust official image
+FROM rust:latest as builder
+WORKDIR /feed
+# Copy the project files from your machine to the container
+COPY ./ ./
+# Build your application for release, inside the container
+RUN cargo build --release
+
+FROM debian:buster-slim
```
Lexy2, line 9: Why not alpine? Slimmer

Aivean: I'll try it.

backend/src/config.ts
Aivean: Only feed cache is replaced. Redis is still used as a semi-persistent in-memory cache for other things.

docker-compose.local.yml
Aivean: See above.

feed/Dockerfile
Aivean: I tried alpine, doesn't work, the issue is probably with the absence of the necessary libraries ([see this](https://stackoverflow.com/questions/49098753/unable-to-run-a-docker-image-with-a-rust-executable)). `debian:buster-slim` is already only 70mb, don't think it's worth trying to make it slimmer.

Aivean: Although I used slimer `rust:slim-buster` for the builder image instead of `rust:latest`.

Lexy2: Okay :-(

backend/package.json
```diff
@@ -54,6 +54,7 @@
     "redis": "^4.0.6",
     "url-parse": "^1.5.10",
     "web-push": "^3.5.0",
-    "winston": "^3.7.2"
+    "winston": "^3.7.2",
+    "node-fetch": "^2.6.7"
```
pazoozoo42, line 58: `node-fetch` is already there, but in `devDependencies`, do we need them in both sets?

pazoozoo42 (commented):
didn't review Rust code, otherwise only minor comment

backend/package.json
Aivean: We don't, probably an artifact from the merge of another changes. Fixed.

pazoozoo42 (approved):

PR #174, opened 2022-11-07
Backup media as a GH workflow

Aivean:


.github/workflows/backup-media.yml
```diff
@@ -0,0 +1,24 @@
...
+  backup-media:
+    environment: StagingRO
+    runs-on: [self-hosted, stage]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          path: media_backup_src
+      - name: Run backup
+        env:
+          MYSQL_ROOT_PASSWORD: root
```
Lexy2, line 21: Can we replace it with an environment secret?

backend/backup_media.sh
```diff
@@ -1,3 +1,13 @@
 #!/bin/bash
 MYSQL_CONTAINER=$(docker ps --format "{{.Names}}" | grep mysql | grep orbitar)
```
Lexy2, line 2: Can we make it
`docker-compose -f docker-compose.ssl.local.yml ps --format json | jq -r -M -c '.[] | select(.Service == "mysql") | .ID'`?

backend/backup_media.sh
```diff
@@ -1,3 +1,13 @@
...
+BACKUP_DIR=$1
+if [ -z "$BACKUP_DIR" ]; then
+    echo "No backup dir given"
+    exit 1
+fi
+
+# 30 minutes timeout
+docker run --rm -i --name=orbitar-backup-media -v "${PWD}":/app -v "$BACKUP_DIR":/orbitar_backup --network container:"$MYSQL_CONTAINER" \
+  -e "MYSQL_ROOT_PASSWORD=$MYSQL_ROOT_PASSWORD" \
+  --entrypoint='/bin/bash' node:17 -c "cd /app/ && npm install && node backup_media.js /orbitar_backup"
```
Lexy2, line 13: Quite cheesy, need to think about it in future

Lexy2 (changes requested):

.github/workflows/backup-media.yml
Aivean: Sure, but it's not very secret :)

backend/backup_media.sh
Aivean: What exactly is your concern here?

backend/backup_media.js
```diff
@@ -130,7 +150,7 @@ const processBatch = async (startWithId) => {
         console.log(outputPath, 'is already there');
         saveEntry(mediaType, outputPath, outputFilename, mediaSrc, itemType, itemId);
       } else {
-        console.log(`Processing media item `, mediaSrc);
+        console.log(`Processing media item `, hideUrl(mediaSrc));
```
pazoozoo42, line 153: I guess this is to hide real media URLs in the log so that they are not visible in the GH flow log? Maybe just remove this log entry then?

pazoozoo42 (approved):

backend/backup_media.js
Aivean: >  this is to hide real media URLs in the log so that they are not visible in the GH flow log

Correct. Wanted to have a compromise between the info logged and some degree of privacy/anonymity. I think logging  `content_id` and the domain of the media URL would be sufficient.

Long term solution would be to create another private repo in the `spaceshelter` org and run this workflow from there.

pazoozoo42: I see, makes sense. Also need to handle log entry on the line 39 in the `download` function:
```
      console.log(`Wrote ${responseBinary.length} bytes to ${dest} for ${url}`);
```

backend/backup_media.js
```diff
@@ -81,16 +81,37 @@ const saveEntry = (mediaType, outputPath, outputFilename, mediaSrc, itemType, it
...
-    lastPartParts[0] = lastPartParts[0].substr(0, 3) + '***';
-    lastPart = lastPartParts.join('.');
-    parts[parts.length - 1] = lastPart;
-    return parts.join('/');
+    try {
+        // parse url
+        const parsedUrl = new URL(url);
+
+        // replace second half of the characters in the string with asterisks
+        const hideChars = (str) => {
```
pazoozoo42, line 90: minor: I would define this in the global scope similar to other functions, up to you

backend/backup_media.js
```diff
@@ -81,16 +81,37 @@ const saveEntry = (mediaType, outputPath, outputFilename, mediaSrc, itemType, it
...
+            if (!str) {
+                return str;
+            }
+            const half = Math.ceil(str.length / 2);
+            return str.substr(0, half) + '*'.repeat(half);
```
pazoozoo42, line 95: `substr` is deprecated, `substring` can be used

pazoozoo42 (approved):
looks good, only minor suggestions

backend/backup_media.js
Aivean: I usually prefer to reduce the namespace scope as much as possible, and performance is not a concern here.

PR #175, opened 2022-11-07
Flush redis cache

Lexy2:
After restoration of the DB we restart the backend container but we don't flush the redis cache, and it stays populated with the old info.

This change resolves this nuisance.


.github/workflows/restore-sanitized-db.yml
```diff
@@ -100,3 +100,8 @@ jobs:
 
       - name: Restart backend
         run: docker-compose -f docker-compose.ssl.local.yml start backend
+
+      - name: Flush Redis cache
```
Aivean, line 104: I think this should go before the `Restart backend`.

Aivean (commented):
Please add some information re: how you tested it.

Lexy2:
> Please add some information re: how you tested it.

Ran the command locally on the staging server, setting the environment variable first.

Aivean (approved):
LGTM

PR #177, opened 2022-11-08
add animation to loader so it doesn't flash when page loads quickly

4vanger:
It is very annoying when page loads and robots banner is flashing for a moment before page loads.
Adding an animation (with 0.5s delay) and smooth border so it will show up only when backend takes a long time to respond.
Fast 3G:
Fast 3G

Slow 3G:
Slow 3G

Artificial 2s delay for API response
Artificial 2s delay for API response

### DIFF
--- a/frontend/src/Pages/LoadingPage.tsx
+++ b/frontend/src/Pages/LoadingPage.tsx
@@ -15,6 +15,8 @@ export default function LoadingPage() {
     });
 
     return (
-        <img className="loading" alt="Loading..." src="/img/robots.gif" />
+        <div className={styles.loading} style={{opacity: 0}}>
+            <img alt="Loading..." src="/img/robots.gif" />
+        </div>
     );
 }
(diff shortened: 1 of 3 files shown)


frontend/src/Pages/LoadingPage.tsx
```diff
@@ -15,6 +15,8 @@ export default function LoadingPage() {
     });
 
     return (
-        <img className="loading" alt="Loading..." src="/img/robots.gif" />
+        <div className={styles.loading} style={{opacity: 0}}>
```
4vanger, line 18: need it here in the styles so it won't flash if CSS is not loaded yet

Aivean:
Good idea, however, not sure about box-shadow. @ArtCraft , what's your opinion?

Also, on slow 3g, no cache:
orb-load

Aivean (approved):

PR #178, opened 2022-11-08
fix error when user is not found

4vanger:
If you try to login with non-existing username - you'll get 500 error because `userRaw` is undefined so it doesn't have `password` field. No existing security threat - just cleanup


Aivean:
Next level: call `bcrypt` even when user is not found to secure against timing attacks.

PR #179, opened 2022-11-13
Add "comment post" button which expands the post reply textarea

Aivean:
Demo:
orb-comm

The top button is hidden when there are no comments:
orb-comm2

### DIFF
--- a/frontend/src/Components/CommentComponent.module.scss
+++ b/frontend/src/Components/CommentComponent.module.scss
@@ -75,14 +75,15 @@
 
         font-weight: bold;
 
-        &.active {
+        &:active, &:hover {
             fill: var(--primary);
+            color: var(--primary);
             font-weight: bold;
         }
     }
     a {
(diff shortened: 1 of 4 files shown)


pazoozoo42 (commented):
Code is ok, but I don't understand which problem this PR is fixing?

Also, when the post is empty there are 3 similar labels about commenting and clicking the top one does nothing: 

image

Aivean:
> Code is ok, but I don't understand which problem this PR is fixing?

It addresses two problems:
1. Users often confuse the textarea in the bottom of the post with the comment reply textarea. That happens when the comment they are replying to is the last one in the post (happened to me once).
2. There was a request for a way to add a comment without manually scrolling to the bottom.

> Also, when the post is empty there are 3 similar labels about commenting and clicking the top one does nothing:

I also don't like the top button very much. Perhaps @ArtCraft has some comment on that?

PR #180, opened 2022-11-22
Save the draft post/comment in the local storage in case the page is reloaded

Aivean:
### Demo:
orb-ls

### Notes:
* saves draft in the local storage per key:
   * the key is parent comment id for posts, subsite for the new post
* removes local storage record on send or when text is cleared
* editing is not supported (the text is restored from the latest content version)

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -33,6 +33,7 @@
         "react-textarea-autosize": "^8.3.4",
         "react-toastify": "^8.2.0",
         "typescript": "^4.6.3",
+        "use-debounce": "^8.0.4",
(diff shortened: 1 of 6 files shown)


4vanger (commented):
LGTM at first read - will read more carefully later.
Should we save a data along with the comment content so we can purge outdated drafts from storage later?

pazoozoo42 (approved):
looks good and very useful, thanks!

Aivean:
> Should we save a data along with the comment content so we can purge outdated drafts from storage later?

Would be nice to have, could use something like that: https://www.npmjs.com/package/ttl-localstorage

PR #186, opened 2022-11-27
Fix for /upload url path that never worked in staging

Lexy2:
Production config is different from the one stored in the repository.

This PR updates the configs to match working configuration.

### DIFF
--- a/caddy/Caddyfile.dev
+++ b/caddy/Caddyfile.dev
@@ -16,7 +16,9 @@ http://{$SERVER_DOMAIN} {
     }
   }
 
-  reverse_proxy http://host.docker.internal:5000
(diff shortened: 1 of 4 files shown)


Aivean (approved):
Looks good.

The only caveat (unrelated to the change) is that caddy doesn't seem to restart on config changes (when using `docker-compose -p orbitar-dev -f docker-compose.local.yml up -d --build`).

Lexy2:
> Looks good.
> 
> The only caveat (unrelated to the change) is that caddy doesn't seem to restart on config changes (when using `docker-compose -p orbitar-dev -f docker-compose.local.yml up -d --build`).

Will create an issue.

PR #189, opened 2022-11-27
Added --watch to caddy's startup args

### LINKED ISSUE #188: Caddy doesn't restart on config changes
When debugging locally or in general, you might want to modify `Caddyfile`.

Caddy should watch changes in the file and apply them immediately instead of requiring a manual container restart.

Lexy2:
Fixes #188 for local development.

Watching `Caddyfile` for changes in production is [discouraged](https://caddy.community/t/why-is-caddy-run-watch-dangerous/8981).

The right way would be to [reload caddy gracefully](https://hub.docker.com/_/caddy) after the config changes are complete.

For us it would be something in the likes of
`docker-compose -f docker-compose.ssl.local.yml exec -w /etc/caddy caddy caddy reload`

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -13,6 +13,8 @@ services:
       - $PWD/caddy/Caddyfile.dev:/etc/caddy/Caddyfile
       - caddy_data:/data
       - caddy_config:/config
+    # Default caddy command borrowed from https://github.com/caddyserver/caddy-docker/blob/master/Dockerfile.tmpl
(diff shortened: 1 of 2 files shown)


Aivean:
What about other the two compose configurations?

Aivean:
> For us it would be something in the likes of
> docker-compose -f docker-compose.ssl.local.yml exec -w /etc/caddy caddy caddy reload

I think a more resilient option would be:
```
docker-compose -f docker-compose.ssl.local.yml restart caddy
```

PR #192, opened 2022-12-10
Basic search with Elasticsearch 7.17

pazoozoo42:
- Backend for basic and advanced search
- Frontend for basic search (open http://orbitar.local/search directly, since it's not referenced from anywhere)

Set up:

0) create elasticsearch container (see definitions in docker-compose files)
1) run `npm run elasticsearch:setup` in `backend/`
2) run `npm run elasticsearch:index` in `backend/`

Indexing is incremental similar to media backup. It's fast, so I guess we can run it every 10 -15 minutes

**Important:** simple UI was implemented just to show how search works. Simple search requires only `term` incoming parameter and searches both posts and comments

Backend supports advanced search, but UI for it is not implemented. Advanced search supports additional parameters:
```
export type SearchRequest = {
    term: string;
    scope?: 'post' | 'comment';
    author?: string;
    response_to?: string;
    site?: string;
    created_at_from: number;
    created_at_to: number;
    rating_from: number;
    rating_to: number;
    search_by_date: boolean;
    search_direction: 'asc' | 'desc';
}
```
Search returns 1000 entries maximum, and no pager is added because rendering of 1000 items is fast

image


backend/package.json
```diff
@@ -13,7 +13,9 @@
     "generate-dummy-content": "NODE_ENV=development node dummy_content_generator.js",
     "lint": "npx eslint src",
     "test": "jest --coverage",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
+    "elasticsearch:setup": ". search/index_elastic_setup.sh",
```
Aivean, line 17: Not a cross-platform solution.

backend/search/id.txt
```diff
@@ -0,0 +1 @@
+0
```
Aivean, line 1: Contrary to media backup (which is run on a different host), I'd suggest to add "indexed" flag to the `content_source` table. The backup script can fetch first X non-indexed records, index them and update status in the database.

Additionally, the indexing script can take a parameter that clears "indexed" flag as the first step (to allow reindexing from scratch).

Rationale for that:
1. In the future we'd want to switch to realtime indexing
2. storing `id.txt` in git introduces sync and update problems, let's avoid that

backend/search/index_data_to_elastic.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
Aivean, line 1: I believe we can restructure that a bit. There is no reason for `search` to reside inside the `backend`, it should be a separate folder/docker service. 

Scripts like `index_elastic_setup` should be executed as a part of the docker image build.

`index_data_to_elastic.js` could also reside inside the running elastic service (could be invoked by using `docker exec`. Since docker-compose provides the network connection between elastic service and mysql, reading and updating data won't be a problem.

Aivean: Additionally, `search` service can provide basic REST endpoint (batch indexing for the first version, incremental indexing in the future).

docker-compose.ssl.dev.yml
```diff
@@ -46,8 +46,27 @@ services:
...
+    image: elasticsearch:7.17.5
+    restart: always
+    ports:
+      - "9200:9200"
+      - "9300:9300"
+    environment:
+      - node.name=es-node01
+      - cluster.name=es-cluster
+      - discovery.type=single-node
+      - "ES_JAVA_OPTS=-Xms128m -Xmx128m"
```
Aivean, line 59: Isn't that too low for java?

docker-compose.ssl.local.yml
```diff
@@ -36,6 +36,24 @@ services:
     volumes:
       - redis_data:/data
 
+  elasticsearch:
+    image: elasticsearch:7.17.5
+    restart: always
+    ports:
+      - "9200:9200"
```
Aivean, line 43: We should not expose ports in the `local` (prod) configs unnecessarily.

frontend/src/API/SearchApi.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default class SearchApi {
+    private api: APIBase;
+
+    constructor(api: APIBase) {
+        this.api = api;
+    }
+
+    async search(term: string): Promise<any> {
+        return await this.api.request<any, any>('/search', {
```
Aivean, line 11: Linter nits.

docker-compose.ssl.dev.yml
Lexy2: I had experience that indexer can fail over time if memory limits were low. Starts okay but then fails.

docker-compose.ssl.local.yml
```diff
@@ -36,6 +36,24 @@ services:
...
+      - "9300:9300"
+    environment:
+      - node.name=es-node01
```
Aivean, line 46: We should probably consolidate all non-changing env variables into a single place (perhaps inside the folder for the new service).

docker-compose.ssl.dev.yml
Aivean: > I had experience that indexer can fail over time if memory limits were low. Starts okay but then fails.

I'd expect as much. As far as I know standard java runtime takes about ≈50mb by itself, and with how heap is organized, I'd recommend to give it at least 512 mb.

Aivean: Also, no reason to set min heap equal to max heap. It's a premature optimization that will unnecessarily hog memory.

pazoozoo42:
Converting to draft to solve issues, will re-open once ready

backend/search/index_data_to_elastic.sh
pazoozoo42: > Scripts like index_elastic_setup should be executed as a part of the docker image build.

this is tricky to implement. Elasticsearch docker image has no folder/file mapping to execute something after container and service is started

Aivean: Check [this SO answer](https://stackoverflow.com/questions/35526532/how-to-add-an-elasticsearch-index-during-docker-build) for some ideas. Let me and @Lexy2 know if you're stuck, we can help research the right approach.

docker-compose.ssl.dev.yml
pazoozoo42: changed to 512, and setting both values to it, I'm not an expert here and in all examples values are the same

Aivean: I'm an expert 8+ years java experience and I advise not to do it :)

https://stackoverflow.com/questions/43651167/is-there-any-advantage-in-setting-xms-and-xmx-to-the-same-value
We don't want to microoptimize the latency of the elasticsearch, and our memory is limited, so we better let JVM manage it.

pazoozoo42: They [recommend](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/heap-size.html) to use same value for both  and have it set up to 50% or server RAM (but not more than 32GB) but this is for dedicated host

Aivean: > They [recommend](https://www.elastic.co/guide/en/elasticsearch/reference/6.6/heap-size.html) to use same value for both and have it set up to 50% or server RAM (but not more than 32GB) but this is for dedicated host

That's for the case when the server's primary function is to host elasticsearch node. In this case it's reasonable to split memory between JVM and OS (file cache) and pre-allocate it to eliminate heap resizes. 

Our usecase is very different. Elasticsearch would be used only occasionally, and we don't really care about search latency. Having more memory available for the file cache (mysql performance) is more important for us.

Aivean:
@pazoozoo42 , thank you for working on this! I took a brief look, looks good overall, however want to try something regarding the docker image, I think it can be simplified. I'll create a separate PR based on this.

docker-compose.ssl.dev.yml
pazoozoo42: @Aivean I agree, could you please adjust these values in your PR that you are planning, thanks

Aivean:
Question: why not use the latest elasticsearch 8.5.x ?

Aivean:
Note: exception during the search (e.g. index doesn't exist) fails the whole backend.

pazoozoo42:
> Note: exception during the search (e.g. index doesn't exist) fails the whole backend.

Thanks, looking into this

docker-compose.ssl.dev.yml
Aivean: I discovered that elasticsearch explicitly checks that xms == xmx at startup and won't start if the are not equal. (sigh). Let's set them both to 512mb and see how that works.

pazoozoo42:
> Note: exception during the search (e.g. index doesn't exist) fails the whole backend.

@Aivean backend wasn't crashing, but it dumped whole exception in the log, I replaced `console.error` with `this.logger.error`

Aivean:
I think either some kind of `limit` or pagination is required, currently it's possible to fetch half of the content in a single request.

pazoozoo42:
> default limit is 1000 and it works fine. Problem is that total number is displayed showing that there's more than a thousand items found, I will fix that by showing something like "found: more than 1000"

@Aivean this is done

backend/migrations/20221212000000-search-indexed.js
```diff
@@ -0,0 +1,37 @@
...
+    dbm = options.dbmigrate;
+    type = dbm.dataType;
+    seed = seedLink;
+};
+
+exports.up = async function(db) {
+    await db.runSql(`
+    alter table content_source
+      add column indexed tinyint(1) not null default 0,
+      add index idx_indexed (indexed);
```
Aivean, line 21: The better index would be `alter table content_source add index idx_indexed_content_source_id (indexed, content_source_id);` (since both columns are used in the query).

Aivean: Fixed.

backend/src/api/SearchController.ts
```diff
@@ -0,0 +1,133 @@
...
+            site,
+            created_at_from,
+            created_at_to,
+            rating_from,
+            rating_to,
+            search_by_date,
+            search_direction
+        } = request.body;
+        try {
+            const result = await this.searchManager.search(
```
Aivean, line 113: Code smell: more than five params. Consider wrapping them into a dedicated request object.

backend/src/managers/SearchManager.ts
```diff
@@ -0,0 +1,230 @@
...
+        this.logger = logger;
+        this.siteManager = siteManager;
+        this.userManager = userManager;
+        const host = process.env.SEARCH_HOST || 'elasticsearch';
+        const port = parseInt(process.env.SEARCH_PORT, 10) || 'elasticsearch';
+        this.client = new Client({ node: `http://${host}:${port}` });
+    }
+
+    buildQuery(
+      term: string,
```
Aivean, line 23: Consider using a [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern).

Aivean:
Links are mangled sometimes, try searching for "dl=0".

Aivean (approved):
LGTM

Aivean (approved):

Aivean:
Hmm, noticed that results are sorted by date ascending. Not the best default searching mode (i'd say date desc or relevance are better).

Aivean (dismissed):

Aivean (changes requested):
Html injection in post title (and probably in the content) is possible.

pazoozoo42:
> Html injection in post title (and probably in the content) is possible.

tags were cleared for source only, I also added clearing for title too

Aivean (approved):

Aivean (approved):

PR #193, opened 2022-12-11
Create a universal deployment workflow

Aivean:
The goal is to have a single workflow that can deploy to both Staging and Prod using same steps.

Additionally, there are optional parameters that can be used for initial migration from the current scheme (volume deletion, restoration from the backup).

The plan is to test this on staging and once ready, run it on prod to deploy the newest version and migrate to the common structure.

### Testing:
* [successful restore-sanitized-db run](https://github.com/spaceshelter/orbitar/actions/runs/3667469296/jobs/6199967216)
* [successful staging deploy](https://github.com/spaceshelter/orbitar/actions/runs/3667612022/jobs/6200176805)
   * with everything: volume deletion, db restoration

### Manual changes potentially required on prod:
* `sudo chown -R github-runner:nogroup /orbitar`


.github/workflows/deploy-universal.yml
```diff
@@ -0,0 +1,102 @@
...
+          cd /orbitar
+
+      - name: Delete volumes
+        if : ${{ github.event.inputs.delete-volumes }}
+        run: docker-compose -f docker-compose.ssl.local.yml down -v
+
+      - name: Restore DB
+        if : ${{ github.event.inputs.restore-db }}
+        run: |
+          docker-compose -f docker-compose.ssl.local.yml stop backend
```
Lexy2, line 83: This one recreates all dependent containers. Thus, if they were deleted in the previous step, they may now exist prematurely.

.github/workflows/deploy-universal.yml
```diff
@@ -0,0 +1,102 @@
...
+          rm -rf /orbitar/*
+          cp -r ./* /orbitar/
+          cd /orbitar
+
+      - name: Delete volumes
+        if : ${{ github.event.inputs.delete-volumes }}
+        run: docker-compose -f docker-compose.ssl.local.yml down -v
+
+      - name: Restore DB
+        if : ${{ github.event.inputs.restore-db }}
```
Lexy2, line 81: 1. Space after `if`
2. [When you use expressions in an if conditional, you may omit the expression syntax (${{ }})](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif)
3. Workflow doesn't seem to respect the boolean flags, always restores DB, always deletes volumes, always clears Redis.

.github/workflows/deploy-universal.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: Build
+        env:
+          # replace None with an empty string
+          BUILD_FLAG: ${{ github.event.inputs.build-flag != 'None' && github.event.inputs.build-flag || '' }}
+        run: docker-compose -f docker-compose.ssl.local.yml build $BUILD_FLAG
+
+      - name: Download DB dump
+        id: get-db
+        if: ${{ github.event.inputs.restore-db }}
```
Lexy2, line 63: FYI:
```
##[debug]Evaluating condition for step: 'Flush redis'
##[debug]Evaluating: (success() && github.event.inputs.flush-redis)
...
##[debug]Expanded: (true && 'false')
##[debug]Result: 'false'
```
Seems like an actions bug that we'll have to mitigate.

```
$ echo '{"environment": "stage", "restore-db": "false", "flush-redis": false}' | gh workflow run 'Deployment (Universal)' --ref universal-deploy --json
could not parse provided JSON: json: cannot unmarshal bool into Go value of type string
$ echo '{"environment": "stage", "restore-db": false, "flush-redis": false}' | gh workflow run 'Deployment (Universal)' --ref universal-deploy --json
could not parse provided JSON: json: cannot unmarshal bool into Go value of type string
$ echo '{"environment": "stage", "restore-db": "False", "flush-redis": "False"}' | gh workflow run 'Deployment (Universal)' --ref universal-deploy --json
could not create workflow dispatch event: HTTP 422: Provided value 'False' for input 'restore-db' not in the list of allowed values (https://api.github.com/repos/spaceshelter/orbitar/actions/workflows/42779484/dispatches)

**Review by Lexy2 — CHANGES_REQUESTED** [review]

File: .github/workflows/deploy-universal.yml [hunk 4]
```diff
@@ -0,0 +1,102 @@
...
+    environment: ${{ github.event.inputs.environment == 'stage' && 'Staging' || 'Production' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Verify environment
+        run: |
+          (diff -wB <(grep -vE '^\s*#' /opt/deployment-specific-files/.env | cut -d '=' -f 1 | sort) \
+          <(grep -vE '^\s*#' .env.sample | cut -d '=' -f 1 | sort)) ||\
+          (echo "Environment variables mismatch" &&\
```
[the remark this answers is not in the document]
Aivean: It didn't work. The `diff` returns code 1 and the whole step fails before reaching the if.

.github/workflows/restore-sanitized-db.yml
```diff
@@ -26,26 +26,20 @@ jobs:
...
-        env:
-          BACKUP_PASSWORD: ${{ secrets.BACKUP_PASSWORD }}
+      - name: Load DB
         run: |
-          cd "$BACKUP_DIR"
-          cd `ls -t | head -n 1`
-          unzip -P $BACKUP_PASSWORD `ls -t | head -n 1`
-          cat `ls -t *.sql | head -n 1` | /usr/bin/mysql -h 127.0.0.1 -u root -p$MYSQL_ROOT_PASSWORD
+          cat ${{ steps.get-db.outputs.db-dump-file }} | /usr/bin/mysql -h 127.0.0.1 -u root -p$MYSQL_ROOT_PASSWORD
+          rm -rf ${{ steps.get-db.outputs.db-dump-path }}
```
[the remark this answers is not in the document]
Aivean: Didn't know that it's done at the end of the job. In this case, you're right, it's not necessary.

.github/workflows/restore-sanitized-db.yml
```diff
@@ -61,12 +55,10 @@ jobs:
             --add-drop-database --databases orbitar_db activity_db -u root -p$MYSQL_ROOT_PASSWORD > stage-backup.sql
 
       - name: Compress database
-        env:
-          STAGING_BACKUP_PASSWORD: ${{ secrets.STAGING_BACKUP_PASSWORD }}
         run: |
           pwd
           ls -la
-          zip -m -P STAGING_BACKUP_PASSWORD stage-backup.zip stage-backup.sql
```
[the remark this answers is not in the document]
Aivean: Yep.

.github/workflows/deploy-universal.yml
Aivean: I don't think it **re**creates them, just creates if they were previously deleted. And they will be used anyway in the latter step.

.github/workflows/deploy-universal.yml
Aivean: > Workflow doesn't seem to respect the boolean flags, always restores DB, always deletes volumes, always clears Redis.

Noticed that as well. Either the `gh workflow run` is broken and it will work once committed to `dev` and triggered via UI, or boolean params are not supported as an input type (weird).

.github/workflows/deploy-universal.yml
Aivean: Relevant: https://github.com/actions/runner/issues/1483

Aivean: The workaround seems to be `outputs.foo == 'true'`

.github/workflows/restore-sanitized-db.yml
```diff
@@ -61,12 +55,10 @@ jobs:
...
+          zip -m -P ${{ secrets.STAGING_BACKUP_PASSWORD }} stage-backup.zip stage-backup.sql
```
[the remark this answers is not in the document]
Lexy2: What if your password will contain a quote?
Not sure 100%, but it will also be exposed in the process startup command.

.github/workflows/deploy-universal.yml
Lexy2: Recreates but not restarts. If we were to do some stuff to the volumes before explicitly recreating them, this could be shooting yourself in the foot eventually. 
Not insisting on changing it though

.github/workflows/deploy-universal.yml
Lexy2: This bug is celebrating 1 year.

Lexy2:
> Manual changes potentially required on prod:
>   `sudo chown -R github-runner:nogroup /orbitar`

We have SGID set on `/orbitar`, but not SUID, is this the desired behaviour?

.github/workflows/deploy-universal.yml
```diff
@@ -0,0 +1,102 @@
...
+          zip-password: ${{ github.event.inputs.environment == 'stage' && secrets.STAGING_BACKUP_PASSWORD || secrets.BACKUP_PASSWORD }}
+
+      - name: Copy everything to /orbitar
+        run: |
+          mkdir -p /orbitar
+          rm -rf /orbitar/*
+          cp -r ./* /orbitar/
+          cd /orbitar
+
+      - name: Delete volumes
```
Lexy2, line 76: I think this step onwards should be executed in the context of `/orbitar` directory. One option is to set `$PWD` in `.env` to `/orbitar`

.github/workflows/deploy-universal.yml
Aivean: I'm not entirely sure what's the concern here is, but I'm open for suggestion.
My goal here was to ensure that backend is stopped before manipulating redis or mysql.

Aivean: I'm not sure it's going to work.

```
docker-compose -f docker-compose.ssl.local.yml stop backend
docker-compose -f docker-compose.ssl.local.yml up -d mysql --wait
```
The second command will recreate the volumes anyway.

Aivean: Just tried it locally:
```
docker-compose -f docker-compose.ssl.local.yml up -d mysql --wait
[+] Running 6/7
 ⠿ Network orbitar_default          Created                                                                                                                                                           0.0s
 ⠿ Volume "orbitar_caddy_data"      Created                                                                                                                                                           0.0s
 ⠿ Volume "orbitar_caddy_config"    Created                                                                                                                                                           0.0s
 ⠿ Volume "orbitar_mysql_db"        Created                                                                                                                                                           0.0s
 ⠿ Volume "orbitar_redis_data"      Created                                                                                                                                                           0.0s
 ⠿ Volume "orbitar_frontend_build"  Created                                                                                                                                                           0.0s
 ⠿ Container orbitar-mysql-1        Waiting                                                                                                                                                          13.6s
```

Aivean:
@Lexy2 , I addressed all review comments.
Here is the last successful deployment to staging: https://github.com/spaceshelter/orbitar/actions/runs/3673186332/jobs/6210036305

Run with:
```
 gh workflow run 'Deployment (Universal)'  --ref  universal-deploy -f environment=stage -f delete-volumes=true -f restore-db=true
 ```

.github/workflows/deploy-universal.yml
Lexy2: Yup, looks like it, sorry.

Lexy2 (approved):
Awesome workflow!

PR #201, opened 2022-12-14
Fix docker compose file used in prod db backup 2

Aivean:
oops, wrong syntax.

This time properly tested: https://github.com/spaceshelter/orbitar/actions/runs/3691512758/jobs/6249665513


Lexy2 (approved):
Cosmetic change, find as long as it builds.

PR #204, opened 2022-12-16
Option to filter posts and comments on user profile page

pazoozoo42:
This is related to search so I started from the `search` branch

https://user-images.githubusercontent.com/57953386/208183015-52ba7436-9e0d-46e5-ba03-80a2508d2895.mp4

### DIFF
--- a/backend/src/db/repositories/CommentRepository.ts
+++ b/backend/src/db/repositories/CommentRepository.ts
@@ -45,12 +45,12 @@ export default class CommentRepository {
 
     async getUserComments(userId: number, forUserId: number, filter: string, page: number, perPage: number): Promise<CommentRawWithUserData[]> {
         const limitFrom = (page - 1) * perPage;
-        let query = `
+        const query = `
                 select c.*, v.vote
                     from comments c
                       left join comment_votes v on (v.comment_id = c.comment_id and v.voter_id = :for_user_id)
                     where
-                      c.author_id = :user_id and c.deleted = 0
+                      c.author_id = :user_id and c.deleted = 0 ${filter ? ' and c.source like :filter ' : ''}
                     order by c.created_at desc
                     limit :limit_from, :limit_count
             `;
@@ -61,15 +61,6 @@ export default class CommentRepository {
               limit_count: perPage
           };
         if (filter) {
-            query = `
-                select c.*, v.vote
-                    from comments c
-                      left join comment_votes v on (v.comment_id = c.comment_id and v.voter_id = :for_user_id)
-                    where
-                      c.author_id = :user_id and c.deleted = 0 and c.source like :filter
-                    order by c.created_at desc
-                    limit :limit_from, :limit_count
-            `;
             params['filter'] = '%' + filter + '%';
         }
         return await this.db.query(query, params);
@@ -82,10 +73,9 @@ export default class CommentRepository {
     }
 
     async getUserCommentsTotal(userId: number, filter: string): Promise<number> {
-        let query = 'select count(*) cnt from comments where author_id = ? and deleted = 0';
+        const query = `select count(*) cnt from comments where author_id = ? and deleted = 0 ${filter ? '  and source like ? ' : ''}`;
         const params: Array<number | string> = [userId];
         if (filter) {
-            query = 'select count(*) cnt from comments where author_id = ? and deleted = 0 and source like ?';
             params.push('%' + filter + '%');
         }
         return this.db.fetchOne<{ cnt: string }>(query, params).then((res) => parseInt(res.cnt || '0'));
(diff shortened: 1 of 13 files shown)


backend/src/db/repositories/CommentRepository.ts
```diff
@@ -43,23 +43,36 @@ export default class CommentRepository {
...
-                limit_count: perPage
-            });
+            `;
+          const params = {
+              user_id: userId,
+              for_user_id: forUserId,
+              limit_from: limitFrom,
+              limit_count: perPage
+          };
+        if (filter) {
```
Aivean, line 63: The queries are mostly identical, should we merge them?

backend/src/db/repositories/CommentRepository.ts
```diff
@@ -43,23 +43,36 @@ export default class CommentRepository {
...
+            query = `
+                select c.*, v.vote
+                    from comments c
+                      left join comment_votes v on (v.comment_id = c.comment_id and v.voter_id = :for_user_id)
+                    where
+                      c.author_id = :user_id and c.deleted = 0 and c.source like :filter
```
Aivean, line 69: Let's maybe use full text index here?

backend/src/db/repositories/PostRepository.ts
```diff
@@ -83,28 +83,46 @@ export default class PostRepository {
...
-            });
+            `;
+        const params = {
+            user_id: userId,
+            for_user_id: forUserId,
+            limit_from: limitFrom,
+            limit_count: perPage
+        };
+        if (filter) {
+            query = `select p.*, v.vote, b.read_comments, b.bookmark, b.last_read_comment_id, b.watch
```
Aivean, line 104: Same as above, let's try to avoid the query duplication.

Aivean (commented):
Looks good overall! And it's a very useful feature, thank you for working on it.

UX related thoughts: perhaps the search icon will look better instead of the text? Or, alternatively, we can get rid of the button altogether, replacing it with a debounced request on input change.

backend/src/db/repositories/CommentRepository.ts
pazoozoo42: done

backend/src/db/repositories/PostRepository.ts
pazoozoo42: I am unsure whether we should use it here and I am debating with myself, but I tend to lean towards not using it. This would duplicate advanced search functionality once it's there

backend/src/db/repositories/CommentRepository.ts
pazoozoo42: I am unsure whether we should use it here and I am debating with myself, but I tend to lean towards not using it. This would duplicate advanced search functionality once it's there

Aivean: Ok, if that's just a temporary solution to be replaced with elasticsearch, I'm ok with it.

pazoozoo42:
> Or, alternatively, we can get rid of the button altogether, replacing it with a debounced request on input change.

implemented "search as you type", removed the button

backend/src/db/repositories/CommentRepository.ts
```diff
@@ -43,23 +43,27 @@ export default class CommentRepository {
...
-            });
+            `;
+          const params = {
+              user_id: userId,
+              for_user_id: forUserId,
+              limit_from: limitFrom,
+              limit_count: perPage
+          };
+        if (filter) {
+            params['filter'] = '%' + filter + '%';
```
Aivean, line 64: We should test this for potential abuse. E.g. super long queries with `%` in them.

backend/src/db/repositories/CommentRepository.ts
```diff
@@ -68,10 +72,13 @@ export default class CommentRepository {
...
     }
 
-    async getUserCommentsTotal(userId: number): Promise<number> {
-        return this.db.fetchOne<{ cnt: string }>('select count(*) cnt from comments where author_id = :user_id and deleted = 0', {
-            user_id: userId
-        }).then((res) => parseInt(res.cnt || '0'));
+    async getUserCommentsTotal(userId: number, filter: string): Promise<number> {
+        const query = `select count(*) cnt from comments where author_id = ? and deleted = 0 ${filter ? '  and source like ? ' : ''}`;
+        const params: Array<number | string> = [userId];
+        if (filter) {
```
Aivean, line 78: Regarding the conditional params. I just tested, and it seems that you can have unused params with `undefined` value:

```
{
              username: username,
              'test': undefined
          }
```
Should reduce the boilerplate.

Aivean (approved):
I addressed my review comments. Feel free to merge it.

One note though:  We need to address the current filter inefficiency before releasing this, either by switching to the full text index, or by using elasticsearch. Currently it's a performance concern.

backend/src/db/repositories/CommentRepository.ts
Aivean: Fixed.

backend/src/db/repositories/CommentRepository.ts
Aivean: Fixed.

PR #206, opened 2022-12-17
Speedup the feed build by using the correct toolchain and docker caching

Aivean:
Before the image had `stable` rust toolchain and had to re-download and recompile everything.

The fix ensure that docker cache layer has all dependencies pre-built and cached on the correct toolchain, so only local changes are recompiled.

### DIFF
--- a/feed/Dockerfile
+++ b/feed/Dockerfile
@@ -1,5 +1,5 @@
 # Tells docker to use the latest Rust official image
-FROM rust:1.64-slim-buster as builder
+FROM rustlang/rust:nightly-buster-slim as builder
 WORKDIR /feed
 # first create a cached layer with the dependencies
 RUN mkdir -p ./src && echo 'fn main() { panic!("Dummy Image Called!") }' > ./src/main.rs
@@ -14,8 +14,8 @@ RUN touch -a -m ./src/main.rs
 RUN cargo build --release
 
 FROM debian:buster-slim
-COPY --from=builder /feed /feed
+COPY --from=builder /feed/target/release/orbitar-index /feed/orbitar-index
 WORKDIR /feed
 EXPOSE 6767
 # Run the binary built inside the container
-CMD ["./target/release/orbitar-index"]
+CMD ["./orbitar-index"]
--- a/feed/Dockerfile
+++ b/feed/Dockerfile
@@ -1,9 +1,9 @@
 # Tells docker to use the latest Rust official image
-FROM rust:slim-buster as builder
+FROM rust:1.64-slim-buster as builder
 WORKDIR /feed
 # first create a cached layer with the dependencies
 RUN mkdir -p ./src && echo 'fn main() { panic!("Dummy Image Called!") }' > ./src/main.rs
-COPY ["Cargo.toml", "Cargo.lock",  "./"]
+COPY ["Cargo.toml", "Cargo.lock", "rust-toolchain.toml",  "./"]
 RUN cargo build --release
 RUN rm -rf ./src
 # Copy the project files from your machine to the container
--- a/feed/rust-toolchain
+++ b/feed/rust-toolchain
@@ -1,1 +1,0 @@
-nightly
--- a/feed/rust-toolchain.toml
+++ b/feed/rust-toolchain.toml
@@ -1,0 +1,2 @@
+[toolchain]
+channel = "nightly-2022-11-04"
--- a/feed/rust-toolchain.toml
+++ b/feed/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "nightly-2022-11-04"
+channel = "nightly"
--- a/feed/src/main.rs
+++ b/feed/src/main.rs
@@ -1,5 +1,6 @@
 #![feature(test)]
 #[macro_use]
+#[allow(unused_imports)]
 extern crate test;
 
 #[macro_use]


feed/Dockerfile
```diff
@@ -8,8 +8,6 @@ RUN cargo build --release
 RUN rm -rf ./src
 # Copy the project files from your machine to the container
 COPY ./ ./
-# Invalidate Cargo cache
-RUN touch -a -m ./src/main.rs
```
Lexy2, line 12: Removing this line always builds dummy image

Lexy2:
Can we also add
`#[allow(unused_imports)]`
to
`extern crate test;`?

The warning freaks me out.

Lexy2 (changes requested):

feed/Dockerfile
Aivean: Ok, I [found the reason](https://stackoverflow.com/questions/58473606/cache-rust-dependencies-with-docker-build#comment117943107_58474618) why I included it in the first place.
> If you're doing RUN echo "fn main() {}" > ./src/main.rs, then COPY ./src ./src, cargo may not rebuild your app. You'll need to update the last modified of the main.rs file to inform cargo to rebuild it. Before your 2nd cargo build, you need to run RUN touch -a -m ./src/main.rs to update the last modified of the file.

I'll keep this, and also think if I can switch to the stable rust and pin the base image version (currently nightly image updates frequently, which defies the caching attempts).

Lexy2: You have `rust-toolchain` file that stipulates `nightly` toolchain, and the reason you do so, because it supports `#![feature(test)]`

Aivean: > You have `rust-toolchain` file that stipulates `nightly` toolchain, and the reason you do so, because it supports `#![feature(test)]`

You're right, stable rust is a no-go.

Aivean:
Ok, attempt number 2, @Lexy2 

I pinned an older version of the `slim-buster` for the `builder` (it shouldn't be updated as frequently).
Also I pinned a specific `rust-toolchain.toml` ("nightly-2022-11-04") which, together with `Cargo.lock` should reduce docker cache invalidation.

Also, silenced unused imports warning.

Lexy2:
Hey @Aivean,

I got to the bottom of the issue. 

### Problem
Because you select a specific toolchain in the `rust-toolchain.toml`, rust image will **always** call `rustup` to update it:
```
Status: Downloaded newer image for rust:1.64-slim-buster
 ---> 085572de3c6e
...
Step 5/14 : RUN cargo build --release
 ---> Running in 51de0818b423
info: syncing channel updates for 'nightly-2022-11-04-x86_64-unknown-linux-gnu'
info: latest update on 2022-11-04, rust version 1.67.0-nightly (215e3cd21 2022-11-03)
info: downloading component 'cargo'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: installing component 'cargo'
info: installing component 'rust-std'
info: installing component 'rustc'
    Updating crates.io index
 Downloading crates ...
 ```
 
 ### Solution:
As you need nightly toolchain to use the unstable test feature, modify the Dockerfile to use 
`FROM rustlang/rust:nightly-slim as builder`
but don't copy and actually remove `rust-toolchain.toml`:
`COPY ["Cargo.toml", "Cargo.lock",  "./"]`
Thus, the image will use the nightly toolchain, but not spend time updating it at each build.

Docker will not pull the updated nightly image from the registry if it is already cached, and thus you'll have a stable build until you run `docker pull`, or execute the job on an agent where your `rustlang/rust:nightly-slim` image does not exist.
 
P.S.: If you don't want to run tests during regular builds (although running tests before deployments is a good thing, I suppose), you could split tests and main code as described [here (StackOverflow)](https://stackoverflow.com/a/37679386/15862075). The benefit is that the main code will compile and run on a `stable` rust version that you can even fix to a specific version (but without `rust-toolchain.toml` file).

**Comment by Aivean** [comment]
@Lexy2 , I cannot remove `rust-toolchain.toml`, because it's needed for local development (if you want to run `cargo build` without docker).

Do you see any problems with the current (updated) solution? Rustup should be cached in in the docker layer, unless Cargo.toml, lock or toolchain are changed.

**Comment by Aivean** [comment]
Btw, test feature seems to be stable, the issue is specifically with benchmarks, I'll look into moving them into a separate optional module.

**Comment by Lexy2** [comment]
> @Lexy2 , I cannot remove `rust-toolchain.toml`, because it's needed for local development (if you want to run `cargo build` without docker).
Then just remove it from the `Dockerfile`

> Do you see any problems with the current (updated) solution? `rustup` should be cached in in the docker layer, unless Cargo.toml, lock or toolchain are changed.
I don't see benefits over the original version without this.

**Comment by Aivean** [comment]
> > @Lexy2 , I cannot remove `rust-toolchain.toml`, because it's needed for local development (if you want to run `cargo build` without docker).
> 
> Then just remove it from the `Dockerfile`
> 
> > Do you see any problems with the current (updated) solution? `rustup` should be cached in in the docker layer, unless Cargo.toml, lock or toolchain are changed.
> 
> I don't see benefits over the original version without this.

The benefit over the original version is following:

Previously, the initial caching compilation of dependencies was mistakenly performed on the stable toolchain (as the `rust-toolchain.toml` was not copied over). And the compilation of the main app triggered another compilation.

Currently, with the toolchain properly pinned, the deps would compile only once in the worst case (when base image is updated, or when dependencies or the toolchain actually changes in the code). When there are code-only changes, only the app is recompiled (almost instantly).

**Comment by Lexy2** [comment]
> Previously, the initial caching compilation of dependencies was mistakenly performed on the stable toolchain (as the `rust-toolchain.toml` was not copied over). And the compilation of the main app triggered another compilation.

Makes total sense!

> However, I agree, that pinning the builder image version might be not necessary.

Let's add this change and base the image on `rustlang/rust:nightly-slim` to speed up the build even more, or separate the nightly features from the stable features, up to you.

**Comment by Aivean** [comment]
> Let's add this change and base the image on `rustlang/rust:nightly-slim` to speed up the build even more, or separate the nightly features from the stable features, up to you.

I just tested, switching to `rustlang/rust:nightly-slim` doesn't speed up the initial build, it still tries to update the rust version and the crate index:
```
=> [builder 3/9] RUN mkdir -p ./src && echo 'fn main() { panic!("Dummy Image Called!") }' > ./src/main.rs                                                                                            0.1s
 => [builder 4/9] COPY [Cargo.toml, Cargo.lock, rust-toolchain.toml,  ./]                                                                                                                             0.0s
 => [builder 5/9] RUN cargo build --release                                                                                                                                                          69.6s
 => => # info: installing component 'clippy'
 => => # info: installing component 'rust-docs'
 => => # info: installing component 'rust-std'
 => => # info: installing component 'rustc'
 => => # info: installing component 'rustfmt'
 => => #     Updating crates.io index
```

With the proper caching of initial build, I don't think that there is any difference. I think `rust:1.64-slim-buster` or `rust:slim-buster` are better, as they support more platforms:
image

Lexy2:
> I just tested, switching to `rustlang/rust:nightly-slim` doesn't speed up the initial build, it still tries to update the rust version and the crate index:

If you don't copy the `rust-toolchain.toml`, it won't try to update the rust version and thus will compile faster. crate index will be updated in any case.

> With the proper caching of initial build, I don't think that there is any difference. I think `rust:1.64-slim-buster` or `rust:slim-buster` are better, as they support more platforms: image

This makes sense.

Lexy2 (approved):

Lexy2 (changes requested):
linux/arm64 is supported

Lexy2 (approved):

Lexy2 (approved):

PR #208, opened 2022-12-17
Try render comment tree as a flat dom list

Aivean:
Currently comment tree is represented as nested structure inside the dom (and we suspect that it has detrimental impact on the layout calculation speed). This PR changes dom representation into a simplified flat structure, where leftMargin is calculated beforehand.


4vanger:
Btw this will help solving scrolling-problem that #159 is trying to fix - because of nested elements top comments may still impact the scroll and we cannot do anything with that. Having flat structure will make it possible to fix that case

PR #209, opened 2022-12-18
User profile updates

pazoozoo42:
- option to show and edit profile bio
 - move theme toggle to the user profile page

Own profile:
image

image

Any other user profile:
image

image

### DIFF
--- a/backend/src/api/types/requests/UserProfile.ts
+++ b/backend/src/api/types/requests/UserProfile.ts
@@ -46,3 +46,11 @@ export type UserRestrictionsResponse = {
     canEditOwnContent: boolean;
     canCreateSubsites: boolean;
 };
+
+export type UserSaveBioRequest = {
+    bio: string;
+};
+
+export type UserSaveBioResponse = {
+    bio: string | boolean;
+};
--- a/frontend/src/Components/CreateCommentComponent.tsx
+++ b/frontend/src/Components/CreateCommentComponent.tsx
@@ -24,8 +24,9 @@ interface CreateCommentProps {
     post?: PostLinkInfo;
     text?: string;
     storageKey?: string;
-
-    onAnswer: (text: string, post?: PostLinkInfo, comment?: CommentInfo) => Promise<CommentInfo | undefined>;
+    defaultPreview?: string;
+    staticEditor?: boolean;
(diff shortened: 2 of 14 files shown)


frontend/src/Types/UserInfo.ts
```diff
@@ -8,6 +8,8 @@ export type UserBaseInfo = {
     id: number;
     username: string;
     gender: UserGender;
+    bio_source: string;
```
Aivean, line 11: Perhaps `UserProfileInfo` is a better place for these fields?

frontend/src/Components/CreateCommentComponent.tsx
```diff
@@ -205,16 +206,22 @@ export default function CreateCommentComponent(props: CreateCommentProps) {
...
+                        setPreviewing(response);
+                    }
+                }
             })
             .catch(error => {
                 console.log('onAnswer ERR', error);
             })
             .finally(() => {
-                setPreviewing(null);
+                //setPreviewing(null);
```
Aivean, line 224: Why is this commented out?

pazoozoo42: debug leftovers, thanks for catching this

Aivean (changes requested):
Overall, looks good. 

One thing that I'm not certain of, is how the preview/editing is implemented in user profile UX-wise.

image



I have several concerns about current implementation:
* Your profile looks different for you compared to how others see it. 
    this is a problem if we consider profile to be a "vanity space"
* Editing of the profile info happens pretty rarely compared to visiting the profile
    currently lots of unnecessary or outright not working UI elements (are shown: editor, editor buttons, "send" button) when own profile is visited
* The UX for profile editing is different from the editing of posts/comments (when the "final" version is shown after "submit" button is pressed). Is there a reason for that?

I'd suggest to make the profile editing UX same as it is for comments/posts. Display the final html in the profile and a small "edit" button, which, when pressed, displays the editor. This should also simplify the code for the `CreateCommentComponent`.

backend/src/api/types/requests/UserProfile.ts
```diff
@@ -45,4 +45,12 @@ export type UserRestrictionsResponse = {
...
-};
\ No newline at end of file
+};
+
+export type UserSaveBioRequest = {
+    bio: string;
+};
+
+export type UserSaveBioResponse = {
+    bio: string | boolean;
```
Aivean, line 55: Please don't do this. Error code is already available on the http level.

PR #210, opened 2022-12-18
spoiler tag

maximt:
- used css blur filter to hide spoiler content
- mouse hover to show spoiler content
- click to show spoiler content permanently


Aivean (approved):
LGTM,

I change the onclick behavior to remove spoiler instead of togging it, I believe it should work better if spoiler contains interactive elements: images, links, etc.

PR #212, opened 2022-12-19
Added missing .env options to .env.development

Lexy2:
Quick & dirty, added necessary options to fix local development.


backend/.env.development
```diff
@@ -7,4 +7,7 @@ SITE_DOMAIN=orbitar.local
 DEFAULT_TL_LANGUAGE=ru
 
 SENDING_BLUE_API_KEY=
-DEEPL_API_KEY=
\ No newline at end of file
+DEEPL_API_KEY=
+
+MYSQL_USER=orbitar
```
Aivean, line 12: I think these are also taken from `.env` for the `NODE_ENV=development` configurations.

Lexy2:
The original issue turned out to be not in environment settings, but rather an internal MYSQL issue, fixed in #213.

PR #215, opened 2022-12-21
Add Settings tab to the user profile page

pazoozoo42:
- move Exit and theme toggle to the user profile settings tab
 - show number of posts, comments and available invites
 - fix a couple of lint warnings

### DIFF
--- a/frontend/src/Components/UserProfileInvites.tsx
+++ b/frontend/src/Components/UserProfileInvites.tsx
@@ -21,7 +21,8 @@ import classNames from 'classnames';
 import DateComponent from './DateComponent'; // Import css
 
 type UserProfileInvitesProps = {
-    username: string
+    username: string;
+    onInvitesChange: () => void;
 };
 
 export const UserProfileInvites = observer((props: UserProfileInvitesProps) => {
@@ -63,6 +64,7 @@ export const UserProfileInvites = observer((props: UserProfileInvitesProps) => {
             const result = await api.inviteAPI.create(reason);
             console.log('CREATE', result);
             forceRefresh();
+            props.onInvitesChange();
             // eslint-disable-next-line
         } catch (error: any) {
             console.log('CREATE ERR', error);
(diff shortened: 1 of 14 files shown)


Aivean:
One useful usecase that we've lost here is changing the theme to see how post preview looks like with different background. Perhaps we should add a small change "theme" button near the "preview" button of the content editor?

frontend/src/Components/UserProfileInvites.tsx
```diff
@@ -63,6 +64,8 @@ export const UserProfileInvites = observer((props: UserProfileInvitesProps) => {
             const result = await api.inviteAPI.create(reason);
             console.log('CREATE', result);
             forceRefresh();
+            props.onInvitesChange();
+            // eslint-disable-next-line
```
Aivean, line 68: Why are these necessary?

Aivean (approved):
LGTM

frontend/src/Components/UserProfileInvites.tsx
pazoozoo42: you cannot type exception callback argument (typescript limitation) but lint is complaining about it

pazoozoo42:
> One useful usecase that we've lost here is changing the theme to see how post preview looks like with different background. Perhaps we should add a small change "theme" button near the "preview" button of the content editor?

we can but this will switch theme permanently and might cause some irritation if they forget to switch it back after post is created and they will need to go to settings to change it back

Aivean:
> > One useful usecase that we've lost here is changing the theme to see how post preview looks like with different background. Perhaps we should add a small change "theme" button near the "preview" button of the content editor?
> 
> we can but this will switch theme permanently and might cause some irritation if they forget to switch it back after post is created and they will need to go to settings to change it back

We can remember the original settings and switch it back automatically when reply is posted.

pazoozoo42:
> We can remember the original settings and switch it back automatically when reply is posted.

@Aivean https://github.com/spaceshelter/orbitar/pull/217

PR #229, opened 2022-12-27
Feed docker build: pin slim-buster image version to allow better caching

Aivean:
Pinned to rust:1.64-slim-buster
image

https://hub.docker.com/layers/library/rust/1.64-slim-buster/images/sha256-c761809aae5fe07769dab99f4404ca5d71a954934a74484e48eb06c0260ef484?context=explore


Lexy2 (approved):
**LGWW**

PR #231, opened 2022-12-27
Status request improvements

Aivean:
Performance/refactoring:

* Remove unnecessary request from status (moved to separate requests, invoked only when needed):
  * site information
  * subscription
* Cache status (`UserStats`) per user
  cache is invalidated when:
      * user visits post, watches/unwatches
      * comment is posted
* since status requests are now very cheap, increase status poll frequency to once per 30 seconds (was 1 minute)

Note: for ease of review this PR is split into the self-contained commits.

### DIFF
--- a/backend/src/managers/UserManager.ts
+++ b/backend/src/managers/UserManager.ts
@@ -141,7 +141,7 @@ export default class UserManager {
                 comments: unreadComments,
                 posts: 0
             }
-        }
+        };
 
         this.userStatsCache.set(forUserId, stats);
         return stats;
--- a/backend/src/managers/UserManager.ts
+++ b/backend/src/managers/UserManager.ts
@@ -37,6 +37,7 @@ export default class UserManager {
     private readonly redis: RedisClientType;
     private siteConfig: SiteConfig;
     private cacheLastVisit: Record<number, Date> = {};
+    private userStatsCache = new Map<number, UserStats>();
     private readonly logger: Logger;
     private readonly mailLogger: Logger;
     private readonly parser: TheParser;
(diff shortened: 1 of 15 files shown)


frontend/src/API/SiteAPIHelper.ts
```diff
@@ -29,8 +29,23 @@ export default class SiteAPIHelper {
...
+
     async list(page: number, perpage: number): Promise<SiteWithUserInfo[]> {
         const result = await this.api.list(page, perpage);
         return result.sites;
     }
+
+    async updateSiteInfo() {
+        if (this.appState.site) {
+            const result = await this.api.site(this.appState.site);
+            if (result && result.site.site === this.appState.site) {
```
pazoozoo42, line 46: `result.site.site === this.appState.site`
is this check needed? It's getting site info by a site name so not quite clear when some other site could be returned here

frontend/src/Components/SiteSidebar.tsx
```diff
@@ -36,6 +36,22 @@ export const SiteSidebar = observer((props: SidebarProps) => {
...
+        if (!siteInfo) {
+            api.site.updateSiteInfo().then().catch(
+                () => toast.error('Не удалось обновить информацию о сайте!'),
+            );
+        }
+    }, [siteInfo, api.site]);
+
+    useEffect(() => {
+        if (subscriptions === undefined) {
+            api.site.subscriptons().then().catch(
```
pazoozoo42, line 49: this is being called multiple times in a row when main page is opened, is this expected?

backend/src/managers/UserManager.ts
```diff
@@ -127,16 +128,30 @@ export default class UserManager {
...
         };
+
+        this.userStatsCache.set(forUserId, stats);
+        return stats;
+    }
+
+    deleteUserStatsCache(forUserId: number) {
+        this.userStatsCache.delete(forUserId);
+    }
+    clearUserStatsCache() {
```
pazoozoo42, line 153: this is called on each comment posted on site, not sure if having this cache would be beneficial

frontend/src/API/SiteAPIHelper.ts
Aivean: `this.api.site` is an async call. While it's in progress, `this.appState.site` could have been changed.

frontend/src/Components/SiteSidebar.tsx
Aivean: Probably not, lemme check.

backend/src/managers/UserManager.ts
Aivean: It would. The cache is cheap. Currently the **average** time interval between comments is 1.5 minutes, meaning ≈2 cache hits per user (assuming 30s status update), more in practice.

One future improvement would be to have an intelligent cache eviction behavior that depends on whether the comment is posted in a post with low number of comments/watches. In such case only affected users could be evicted from the cache.

frontend/src/Components/SiteSidebar.tsx
Aivean: This effect is called once (can be verified by adding console.log). The `subscription` requests that you see are different (originated from `useFeed`).

There might be a problem with `useFeed`, but it's not a regression introduced in this PR.

Aivean: Please see #235 (a regression from the sorting feature).

pazoozoo42 (approved):

PR #232, opened 2022-12-28
Populate user cache with active users on server startup

Aivean:
Speeds up server startup by reading recent users in one batch to populate user cache. Before users were read one by one on demand, which was slow.

### DIFF
--- a/backend/src/db/repositories/UserRepository.ts
+++ b/backend/src/db/repositories/UserRepository.ts
@@ -195,6 +195,16 @@ export default class UserRepository {
         `, {user_id: userId, visited_at: dateToInsert});
     }
 
+    getLastActiveUsers(): Promise<UserRaw[]> {
+        return this.db.fetchAll(`
+            select * from users
+            where user_id in (select distinct user_id
+                              from activity_db.user_activity
+                              where visited_at > subdate(now(), interval 30 day)
+                              ) limit 1000
+        `);
+    }
+
     /**
      * Returns the number of active users that are not on trial.
      */
--- a/backend/src/managers/UserCache.ts
+++ b/backend/src/managers/UserCache.ts
@@ -4,7 +4,13 @@ import {UserRaw} from '../db/types/UserRaw';
 
 export class UserCache {
     private userRepository: UserRepository;
-    private initialized: Promise<void> | true | false = false;
+    /**
+     * Cache initialized state:
+     * - false: not initialized
+     * - Promise<void>: initializing, wait for it
+     * - true: initialized
+     */
+    private initializedState: Promise<void> | boolean = false;
     private cacheId: Record<number, UserInfo> = {};
     private cacheUsername: Record<string, UserInfo> = {};
     private cachedUserParents: Record<number, number | undefined | false> = {};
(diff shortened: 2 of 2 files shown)


backend/src/db/repositories/UserRepository.ts
```diff
@@ -195,6 +195,16 @@ export default class UserRepository {
...
     }
 
+    getLastActiveUsers(): Promise<UserRaw[]> {
+        return this.db.fetchAll(`
+            select * from users
+            where user_id in (select distinct user_id
+                              from activity_db.user_activity
+                              where visited_at > subdate(now(), interval 30 day)
+                              ) limit 1000
+        `);
```
pazoozoo42, line 205: is this fast? Join instead of subquery would be probably faster

backend/src/managers/UserCache.ts
```diff
@@ -4,6 +4,7 @@ import {UserRaw} from '../db/types/UserRaw';
 
 export class UserCache {
     private userRepository: UserRepository;
+    private initialized: Promise<void> | true | false = false;
```
pazoozoo42, line 7: will `private initialized: Promise<void> | boolean = false;` work here?

backend/src/managers/UserCache.ts
```diff
@@ -12,7 +13,27 @@ export class UserCache {
...
+        if (this.initialized) {
+            return this.initialized;
+        }
+        this.initialized = (async () => {
+            const users = await this.userRepository.getLastActiveUsers();
+            for (const user of users) {
+                this.cache(this.mapUserRaw(user));
+            }
+            this.initialized = true;
+        })();
```
pazoozoo42, line 29: it works but it feels a bit hacky. Having two properties, one for the action and another one for status would probably make the code more readable

pazoozoo42 (dismissed):

backend/src/db/repositories/UserRepository.ts
Aivean: Should be fast enough (the query is executed only once at server start). Join is trickier to write (need to group by all fields in the `user` table?).

backend/src/managers/UserCache.ts
Aivean: Good catch.

backend/src/managers/UserCache.ts
Aivean: I think of this `initialized` property as a state machine:
`false` → (awaitable `Promise`) → `true`

I think spitting it into two separate properties introduces the possibility of having unwanted states, and the readability should be addressed in another way. I'll try giving this prop another name and adding some comments.

backend/src/db/repositories/UserRepository.ts
Aivean: And to clarify, the improvement is from batching reads (previously users were fetched one by one, each in a separate transaction), queries are fast enough in either case.

backend/src/managers/UserCache.ts
Aivean: Please check if now it's more clear.

pazoozoo42 (approved):

PR #235, opened 2022-12-29
Fix `useFeed` being called two times in a row

Aivean:
The problem was in the value of `sorting` being used as effect dependency.

Initially `sorting` was `undefined` and it was initialized to the stored sorting value returned by the feed request. Which in turn resulted feed request performed two times: first for the "undefined" value, second time for the concrete value.

### DIFF
--- a/frontend/src/API/use/useConditional.ts
+++ b/frontend/src/API/use/useConditional.ts
@@ -1,0 +1,12 @@
+import {useRef} from 'react';
+
+/**
+ * Bumps the value of a variable to the next value if condition is true.
+ */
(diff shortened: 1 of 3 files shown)


pazoozoo42 (approved):
Can we use `react-use` library instead of self-written hook? This library has `usePrevious` and also many other useful hooks we could use in the future

https://github.com/streamich/react-use

Aivean:
> Can we use `react-use` library instead of self-written hook? This library has `usePrevious` and also many other useful hooks we could use in the future
> 
> https://github.com/streamich/react-use

I've created a separate issue for that : #236

PR #237, opened 2022-12-30
hide popup when clicking links inside

4vanger:
Bug: clicking link on a popup will follow the link but won't close the popup.
Before:

After:


Aivean (approved):
LGTM

PR #230, opened 2022-12-27
Video tag poster for idiod.video

Aivean:
1. Parser updated to determine image preview from idiod.video and dump.video `mp4` links and set preload to none, instead adding `poster`:
```
<video  preload="none" poster="https://idiod.video/preview/8feuw2.mp4" controls="" width="500"><source src="https://idiod.video/8feuw2.mp4" type="video/mp4"></video>
```

---

2. Youtube embeds updated to be lightweight image previews until clicked:
yt-demo

memory usage before:
image

after:
image

----

3. Introduced versioning mechanism for the Parser. When parser version is bumped, html is re-parsed and saved (on demand), so users always see up-to-date parser results. 

Allows to introduce breaking parser changes in the future (embeds, mention links, etc).

PR #243, opened 2022-12-31
Lightweight video embed component for the supported mp4 video hostings

Aivean:
Demo:
v-demo

Implementation is similar to #230

### DIFF
--- a/frontend/src/Components/ContentComponent.tsx
+++ b/frontend/src/Components/ContentComponent.tsx
@@ -42,9 +42,10 @@ function updateVideo(video: HTMLVideoElement) {
     });
 }
 
-function updateImg(img: HTMLImageElement) {
+function processYtEmbed(img: HTMLImageElement) {
     // if has class youtube-embed convert to iframe on click
     const ytUrl = img.dataset.youtube;
+
     if (ytUrl && !img.classList.contains('youtube-embed-processed')) {
         img.classList.add('youtube-embed-processed');
         img.addEventListener('click', (e) => {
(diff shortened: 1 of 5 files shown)


frontend/src/Components/ContentComponent.tsx
```diff
@@ -59,7 +60,34 @@ function updateImg(img: HTMLImageElement) {
...
+    const videoUrl = img.dataset.video;
+
+    if (videoUrl && !img.classList.contains('video-embed-processed')) {
+        img.classList.add('video-embed-processed');
+        img.addEventListener('click', (e) => {
+            e.preventDefault();
+            const video = document.createElement('video');
+            video.src = videoUrl;
+            video.controls = true;
+            video.autoplay = true;
```
4vanger, line 79: JFYI: this often won't play video automatically - browser learns user's habit and if user often plays video on this website - it will allow autoplay

frontend/src/Components/ContentComponent.tsx
```diff
@@ -59,7 +60,34 @@ function updateImg(img: HTMLImageElement) {
...
+            video.loop = !!img.dataset.loop;
+            video.style.width = img.width.toString() + 'px';
```
4vanger, line 81: Won't work if poster is absent or generic placeholder was used

4vanger (dismissed):
Some notes but approving to unblock

frontend/src/Components/ContentComponent.tsx
Aivean: This logic expects `placeholder` from TheParser.

PR #246, opened 2023-01-02
Fix duplicate search request

pazoozoo42:
No hook dependency is needed for initial search on page load

### DIFF
--- a/frontend/src/Pages/SearchPage.tsx
+++ b/frontend/src/Pages/SearchPage.tsx
@@ -1,4 +1,4 @@
-import React, {useCallback, useEffect, useMemo, useState} from 'react';
+import React, {useEffect, useRef, useState} from 'react';
 import styles from './SearchPage.module.scss';
 import feedStyles from './FeedPage.module.scss';
 import {useForm, SubmitHandler} from 'react-hook-form';
@@ -64,8 +64,6 @@ function SearchResult(props: {
     </>;
 }
 
-let latestSearchTerm = '';
-
 export default function SearchPage() {
     const api = useAPI();
     const [isSearching, setSearching] = useState(false);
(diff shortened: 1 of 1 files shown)


frontend/src/Pages/SearchPage.tsx
```diff
@@ -101,7 +101,7 @@ export default function SearchPage() {
         } else {
             setFocus('term');
         }
-    }, [defaultSearch]);
+    }, []);
```
Aivean: I think this is a backward way of fixing the issue. Generally `useEffect` should have all variables, that are used inside, as dependencies.

Instead of removing `defaultSearch` from the dependencies list, I suggest calling `setSearchParams({term});` on form submit, instead of calling `doSearch` directly.

pazoozoo42: ~No, I'm pretty sure removing dependency is the right fix for this use case. The logic should be applied only once on the page load (physical page load, not component render)~
on the second thought I think you are right, I will adjust the code

pazoozoo42:
Restructured the component which also fixed this issue

frontend/src/Pages/SearchPage.tsx
```diff
@@ -64,62 +64,61 @@ function SearchResult(props: {
     </>;
 }
 
+let latestSearchTerm = '';
```
Aivean: This is not an idiomatic approach, as it would not work if we use two instances of the SearchPage component (probably won't happen in our case, but still).

The correct approach would be to utilize [`useRef`](https://beta.reactjs.org/reference/react/useRef) hook, or our wrapper [`usePrevious`](https://github.com/spaceshelter/orbitar/blob/0d0deb35837330217a4e3c523332a17d47b5190f/frontend/src/API/use/usePrevious.ts#L7)

Aivean (commented):
Overall, LGTM, but please address the lint warnings and replace `let` with the hook.

Aivean (dismissed):

Aivean (approved):

PR #252, opened 2023-01-13
Add pictshare instance for media self-hosting

Aivean:
Adds [`pictshare`](https://github.com/HaschekSolutions/pictshare/) docker-compose service.

The uploaded media is accessible via https://orbitar.domain/media/hash.ext

### DIFF
--- a/backend/src/parser/TheParser.ts
+++ b/backend/src/parser/TheParser.ts
@@ -28,8 +28,10 @@ export default class TheParser {
     // Bump this version when introducing breaking changes to the parser.
     // Content will be re-parsed and saved on access when this version changes.
     static readonly VERSION = 2;
+    readonly domain;
 
-    constructor() {
+    constructor(domain = 'orbitar.local') {
+        this.domain = domain;
         this.allowedTags = {
             a: (node) => this.parseA(node),
             img: (node) => this.parseImg(node),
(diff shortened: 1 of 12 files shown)


caddy/Caddyfile.local
```diff
@@ -25,6 +25,16 @@ http://{$SERVER_DOMAIN} {
     }
   }
 
+  handle_path /media/* {
```
4vanger, line 28: shorten to `/m/` to save few pennies on traffic in long run

caddy/Caddyfile.ssl.dev
```diff
@@ -17,6 +17,16 @@
     }
   }
 
+  handle_path /media/* {
```
4vanger, line 20: same note about shorter path

caddy/Caddyfile.ssl.local
```diff
@@ -27,6 +27,16 @@
     }
   }
 
+  handle_path /media/* {
```
4vanger, line 30: same note about shorter pass

backend/src/parser/TheParser.ts
```diff
@@ -28,8 +28,10 @@ export default class TheParser {
     // Bump this version when introducing breaking changes to the parser.
     // Content will be re-parsed and saved on access when this version changes.
     static readonly VERSION = 2;
+    readonly domain;
 
-    constructor() {
+    constructor(domain = 'orbitar.local') {
```
4vanger, line 33: We should use prod URL by default. Otherwise in case someone will miss arguments during dev - they won't notice it in dev but it will be failing in production. It should be opposite.
Or maybe load HOST_URL from env variables

Aivean:
Converting to draft, as another approach — separate instance of hosting is explored.

4vanger (approved):
LGTM at first pass. One minor note - shorten url from `/media/` to `/m/` to save a few pennies over the years

tucnak (approved):
Сlean and reasonable although Caddy stuff is hard to get through by untrained eye.

backend/src/parser/TheParser.ts
Aivean: I think a better way is to remove the default value here altogether.

Aivean:
Superseded by #254

PR #253, opened 2023-01-13
Run restore-db job from /orbitar folder to reuse the currently deployed version

Aivean:


Lexy2 (approved):

Lexy2:
@Aivean missed this one, don't see any issues, but double-check if merge would not break anything.

PR #256, opened 2023-01-17
Extract metadata from self-hosted media urls and add it to images and videos

Aivean:
To recap, the basic idea is to make pictshare encode the metadata (dimensions) into the returned urls.

Backend decrypts such urls and adds metadata to `img` and `video`.

Caveat: `max-width` and `max-height` in image dimensions no longer work as expected.


Aivean:
Looks like I managed to fix the styles, but there are still some issues with layout shifts.

PR #258, opened 2023-01-22
Calculate self-regulation "contentRating" part based on the unique users

Aivean:
To recap, the final self-regulation score is composed of the `contentRating` and `userRating` (each in the range 0..1):

     ((contentRating + 1) * userRating - 1) * 1000

Before, the `contentRating` part was based on the sum (absolute value) of the other's votes for the user's content:

```sql
select sum(vote) from votes where user_id = ...
```

This could lead to abuse, as any individual user can potentially vote for every possible content entity of another user, creating biased distribution.

To mitigate the issue above, this PR proposes to first group votes by users, then normalize them (using a non-linear function) into a fixed range, and only after that do the final sum:

```sql
select sum(rating) from (
   select non_linear_transform( sum(votes) ) 
    from votes where user_id = ...
    group by voter_id
)
```

This way the contribution to the `contentRating` from each individual voter is bounded.

Specifically, the actual implementation uses scaled bipolar sigmoid function as a non-linear transform (final range from -2 to 2):
```sql
select sum(rating) as rating, count(*) as voters
   from (SELECT (4 / (1 + EXP(LEAST(50, -0.1 * sum(comment_rating + post_rating))))) - 2 AS rating
   FROM user_user_rating
     WHERE user_user_rating.user_id = :user_id
       and exists(select 1 from user_karma where user_karma.voter_id = user_user_rating.voter_id and vote != 0)
         group by voter_id
   ) t
```
Additionally, only votes from the users with full rights are considered.

### DIFF
--- a/frontend/src/Components/UserProfileKarma.tsx
+++ b/frontend/src/Components/UserProfileKarma.tsx
@@ -47,10 +47,6 @@ export const UserProfileKarma = (props: UserProfileKarmaProps) => {
         Object.keys(karmaResult.postRatingBySubsite)
             .reduce((acc, key) => acc + karmaResult.postRatingBySubsite[key], 0);
 
-    const sumCommentRating = !karmaResult ? 0 :
-        Object.keys(karmaResult.commentRatingBySubsite)
-            .reduce((acc, key) => acc + karmaResult.commentRatingBySubsite[key], 0);
-
     const activeKarmaVotesSum = !karmaResult ? 0 :
         Object.keys(karmaResult.activeKarmaVotes)
             .reduce((acc, key) => acc + karmaResult.activeKarmaVotes[key], 0);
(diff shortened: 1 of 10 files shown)


frontend/src/Components/UserProfileKarma.tsx
```diff
@@ -173,14 +169,28 @@ export const UserProfileKarma = (props: UserProfileKarmaProps) => {
...
+                    <h3>Прогресс прав:</h3>
                     <pre>
                         {JSON.stringify(karmaResult.trialProgress, null, 2)}
                     </pre>
+                    </div>
                 </div>
 
                 <div className={styles.container}>
-                    <Karma commentsSumRating={sumCommentRating} postsSumRating={sumPostRating}
+                    <Karma contentSumRating={karmaResult.totalNormalizedContentRating} postsSumRating={sumPostRating}
```
pazoozoo42: `postsSumRating` is no longer used, maybe remove it to avoid confustion

pazoozoo42 (dismissed):
Code looks fine and works locally with no issues. Got one minor comment about getting rid of unused pieces of code. I have no strong opinion on the idea of the changes, so I suggest you wait for at least one more approval

PR #263, opened 2023-01-28
Barmalini improvements

Aivean:
Barmalini is a local meme — an anonymous user that anyone (well, some) has access to.

This PR solidifies this in the codebase, adding some sensible restrictions:
1. Any user with full rights has access to barmalini.
2. There is no single barmalini password
3. Users can generate a temporary (1 hour) barmalini password
4. Any Barmalini session is capped at 1 hour
5. Barmalini rights are capped below first negative tier (10 minutes slow mode, no invites, no voting)
6. It's impossible to reset barmalini password

Mini demo:
barmalini


backend/src/managers/UserManager.ts
```diff
@@ -111,6 +112,10 @@ export default class UserManager {
 
     async checkPassword(username: string, password: string): Promise<UserInfo | false> {
         const user = await this.getByUsername(username);
+        if (config.barmalini.userId && user?.id === config.barmalini.userId) {
```
4vanger: this and line 256 and 423 are the same - worth separating in separate method `isValidBarmalini` for consistency

backend/src/session/Session.ts
```diff
@@ -61,6 +71,16 @@ export default class Session {
         this.data = new SessionData('');
     }
 
+    public getAgeMillis() {
+        if (!this.created) return 0;
+        const now = new Date();
+        return now.getTime() - this.created.getTime();
+    }
+
+    public isBarmalini() {
```
4vanger, line 80: Use `isValidBarmalini` from `UserManager` - can't do the opposite because session doesn't exist there yet

backend/src/session/Session.ts
```diff
@@ -138,14 +158,19 @@ export class SessionData {
 export function session(db: DB, logger: Logger): RequestHandler {
     return (req, res, next) => {
         req.session = new Session(db, logger, req, res);
-        req.session.restore()
-            .then(() => {
+        return (async () => {
```
4vanger: this is really hard to read - too much async calls.
Are you sure you need to return async function from current function? It wasn't returning it previously.
If `session` can return async function this could be simplified a lot by throwing away return statement

frontend/src/Components/UserProfileSettings.tsx
```diff
@@ -51,17 +54,81 @@ export default function UserProfileSettings(props: UserProfileSettingsProps) {
...
+        if (selection) {
+            selection.removeAllRanges();
+            selection.addRange(range);
+        }
+    };
+
+    return (
+        <div className={styles.barmalini}>
+            {access && <div>
+                <div><span className={styles.label}>Логин:</span> {access.login}</div>
```
4vanger: It is worth noting that password is valid for only one hour.
```suggestion
                <div>
                Счастливого бармаления. Пароль истекает через час.<br />
                <span className={styles.label}>Логин:</span> {access.login}</div>
```

backend/src/api/UserController.ts
```diff
@@ -338,4 +341,33 @@ export default class UserController {
             return response.error('error', `Could not update bio`, 500);
         }
     }
+
+    async barmaliniPassword(request: APIRequest<BarmaliniPasswordRequest>, response: APIResponse<BarmaliniPasswordResponse>) {
```
4vanger, line 344: Add a check that current user is already a barmalini.

4vanger (dismissed):
Few notes here and there but generally ok
lgtm

backend/src/api/UserController.ts
Aivean: It's covered by user restrictions, but won't hurt to add an explicit check.

backend/src/session/Session.ts
Aivean: Its not returning an async function — its returning a promise — a result of async function call.

Previously it was using promis API - then/catch, but the async logic got expanded with more async calls, so I rewrote it by introducing an async block `async () => {...}` which is immediately called (returning the promise).

I understand the readability concern, probably worth adding a comment, but I don't think that alternative (e.g. using promise API, as before) would be more readable.

backend/src/session/Session.ts
Aivean: Unfortunately `Session` doesn't have a dependency on the `UserManager`, no worth the effort to introduce one over this.

Aivean:
@4vanger, thanks for the review. Addressed review comments, please take another look if you have time, otherwise I'll just merge it.

4vanger (approved):

PR #264, opened 2023-01-29
Fix a self-regulation bug when user-user rating was not counted for the new users

Aivean:
The bug — forgot to add the creation of the **new** rows in the `user_user_rating` table.

The fix is simple:
* add the missing `insert ignore` during voting
* add migration to repopulate `user_user_rating` table

### DIFF
--- a/backend/migrations/20230121673962-normalized-sr-fix.js
+++ b/backend/migrations/20230121673962-normalized-sr-fix.js
@@ -1,0 +1,48 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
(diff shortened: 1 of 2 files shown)


backend/migrations/20230121673962-normalized-sr-fix.js
```diff
@@ -0,0 +1,48 @@
...
+    // repopulate user_user_rating from post_votes and comment_votes
+    await db.runSql(`delete from user_user_rating;`);
+
+    await db.runSql(`
+        insert into user_user_rating (user_id, voter_id, comment_rating, post_rating)
+        select user_id,
+               voter_id,
+               sum(cr),
+               sum(pr)
+        from (select author_id as user_id, voter_id, sum(vote) as cr, 0 as pr
```
4vanger, line 27: So `pr` is always zero?

4vanger (approved):
LGTM

PR #265, opened 2023-01-30
Move backup media into a daemon container (wip)

pazoozoo42:


backend/src/parser/TheParser.ts
```diff
@@ -299,7 +304,7 @@ export default class TheParser {
             return this.parseDisallowedTag(node);
         }
         const text = this.renderVideoTag(url, node.attribs['loop'] !== undefined);
-        return { text, mentions: [], urls: [], images: [url] };
+        return { text, mentions: [], urls: [], mediaUrls: [url], images: [url] };
```
Aivean, line 307: don't forget `processImage` and `processVideo`

media/run-from-scratch.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+docker build -t orbitar-media . && \
```
Aivean, line 2: This is not needed in the final version, since the container should be run via docker-compose.

backend/package.json
```diff
@@ -34,7 +34,8 @@
     "ts-jest": "^28.0.3",
     "@types/jest": "^27.5.1",
     "@types/crypto-js": "^4.1.1",
-    "image-size": "^1.0.2"
+    "image-size": "^1.0.2",
```
Aivean, line 37: Should not be needed in the final version.

media/src/app.ts
```diff
@@ -0,0 +1,276 @@
...
+}
+
+interface IContentSourceItem extends RowDataPacket {
+  content_source_id: number;
+  html: string;
+  ref_type: ItemType.COMMENT | ItemType.POST;
+  ref_id: number;
+}
+
+if (!fs.existsSync(config.dbFile)) {
```
Aivean, line 28: In the final version I suggest to replace files with a single `tinyint` flag in the `content_source`, similar to `indexed`.

However, it's worth to have a local in-memory cache of the processed urls.

media/src/app.ts
```diff
@@ -0,0 +1,276 @@
...
+    const [rows] = await connection.query<IContentSourceItem[]>(
+      `select
+        cs.content_source_id,
+        coalesce(p.html, c.html) as html,
+        coalesce(p.source, c.source) as source,
+        cs.ref_type,
+        cs.ref_id
+      from
+          content_source cs
+      left join comments c on c.content_source_id = cs.content_source_id
```
Aivean, line 140: A reminder to remove this join. We want to parse all `content_source` entries, even those from the history (that don't have associated post or comment).

media/src/app.ts
```diff
@@ -0,0 +1,276 @@
...
+
+    let itemId = startWithId;
+    for (const entry of rows) {
+      itemId = entry.content_source_id;
+      console.log(`${Date.now()} working with item ${itemId}`);
+      let itemType = entry.ref_type as ItemType;
+      let content = entry.source;
+
+      try {
+        http.globalAgent.maxSockets = 1024;
```
Aivean, line 158: A keepalive agent is worth adding here:
```
const agent = new http.Agent({
    keepAlive: true
})
```

media/src/app.ts
```diff
@@ -0,0 +1,276 @@
...
+        const parserResponse = await fetch('http://host.docker.internal:5001/api/v1/post/preview', {
```
Aivean, line 159: A keepalive `fetch` agent is worth adding here:
```
const agent = new http.Agent({
    keepAlive: true
})
```

backend/src/api/PostController.ts
```diff
@@ -260,13 +262,20 @@ export default class PostController {
 
     preview(request: APIRequest<PostPreviewRequest>, response: APIResponse<PostPreviewResponse>) {
         const userId = request.session.data.userId;
-        if (!userId) {
+        // allow access from private IPs with no session id
+        if (!userId && !isPrivate(request.ip)) {
```
Aivean, line 266: Please be extra careful here. Verify that IP comes from the actual user/CF and NOT from caddy. 

I'd suggest to add a special authorization token specifically for this case. 

Alternatively, as `media` container has DB access, you can create/use an actual session ID.

backend/src/api/PostController.ts
```diff
@@ -260,13 +262,20 @@ export default class PostController {
...
             return response.authRequired();
         }
         const content = request.body.content;
+        const detailedResponseRequested = request.body.details;
         try {
-            const result = this.postManager.preview(content);
-            response.success({ content : result });
+            let result;
+            if (detailedResponseRequested) {
```
Aivean, line 273: I'd say merge them for simplicity. The returned extra info is not a secret.

PR #270, opened 2023-02-10
Simple pager

pazoozoo42:
As we discussed this new pager looks like this: 

image

image

image


Aivean:
Refactored and improved a bit.

Aivean (approved):

PR #271, opened 2023-02-11
Anonymous posts

Aivean:
Add the ability to mark some posts as anonymous (done manually via the new table `anonymous_posts`). All comments, posted in such posts have autor overridden by the specified user.

### Quick demo:
anon-demo
In this demo I'm logged in as `Aivean`, but comments in that particular post are posted from `Aivean2`. This is achieved by having this row in `anonymous_posts`:

| post\_id | user\_id |
| :--- | :--- |
| 7819 | 253 |


backend/src/api/PostController.ts
```diff
@@ -309,6 +316,7 @@ export default class PostController {
         }
         catch (err) {
             this.logger.error('Comment create failed', { error: err, user_id: userId, format, content, post_id: postId });
+            this.logger.error(error);
```
pazoozoo42: is this needed? `error` is not defined anywhere

frontend/src/API/PostAPIHelper.ts
```diff
@@ -64,7 +65,8 @@ export default class PostAPIHelper {
             post: post,
             comments: comments,
             site: siteInfo,
-            lastCommentId: lastCommentId
+            lastCommentId: lastCommentId,
+            anonymousUser: response.anonymousUser && this.appState.cache.setUser(response.anonymousUser)
```
pazoozoo42: I'm probably missing something but I cannot seem to find where this cache is used, `APICache` has no method for that and there's no direct access in the code

pazoozoo42 (dismissed):

backend/src/api/PostController.ts
Aivean: It's a typo. good catch.

frontend/src/API/PostAPIHelper.ts
Aivean: You're right. We can probably drop it altogether.

PR #272, opened 2023-02-13
spoiler blur with sharp edges

maximt:
using svg filter to get blur effect with sharp edges

https://user-images.githubusercontent.com/230060/218472518-c9b7f1fb-de05-4c1f-b954-1ea00431d9d0.mp4

### DIFF
--- a/frontend/src/index.scss
+++ b/frontend/src/index.scss
@@ -62,28 +62,38 @@ a {
   color: var(--danger);
 }
 
-.spoiler {
-  background-color: var(--fgHard);  
-  opacity: 0.2;
-  /* blur to text */
-  filter: blur(4px);
-  -webkit-filter: blur(4px);
+svg#spoiler-mask {
+  /* hide svg */
+  width: 0;
+  height: 0;
+  overflow: hidden;
+  visibility: hidden;
+  /* Firefox fix */
+  border: 0;
+  clip: rect(0 0 0 0);
+  height: 1px;
+  margin: -1px;
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  width: 1px;
+}
 
+.spoiler {
+  background-color: var(--fg);
+  filter: blur(2px);
+  opacity: .8;
   /* more blur to img video etc */
-  & > img, video {
-    filter: blur(16px);
-    -webkit-filter: blur(16px);
+  & > * {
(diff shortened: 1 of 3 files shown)


frontend/src/Components/ContentComponent.tsx
```diff
@@ -128,12 +126,53 @@ function updateImg(img: HTMLImageElement) {
...
-    const spoilerOnClickHandler = () => {
-        spoiler.classList.remove('spoiler');
-        spoiler.removeEventListener('click', spoilerOnClickHandler);
+function updateSpoilers(div: HTMLDivElement) {
+    const spoilers = div.querySelectorAll('span.spoiler');
+    if (spoilers.length === 0) {
+        return;
+    }
+
+    const svg = document.getElementById('sharpBlurSVG');
```
Aivean: Perhaps we can have this SVG in file form in `Assets/` folder and always added to body?


e.g. import svg like this:
https://github.com/spaceshelter/orbitar/blob/1ed9e95a34678ac7f02620c9923a901a7eff2fa0/frontend/src/Components/Topbar.tsx#L9

and add it somewhere here: 
https://github.com/spaceshelter/orbitar/blob/5580fd1f62a2f3d55e10aa51e5da0d3f70fa644a/frontend/src/App.tsx#L110

maximt: Yes, this was the first version, but this svg is not a image, Im not sure Assets/ is the correct place, and decided to create the svg on demand. But if its ok, I can move svg-filter to assets/

Aivean: The definition of "asset" is wider than just an image. I think it would be more readable to have this svg the explicit form.

Aivean (approved):
Concise and elegant. Good job!

frontend/src/index.scss
```diff
@@ -62,28 +62,38 @@ a {
...
+  width: 0;
+  height: 0;
+  overflow: hidden;
+  visibility: hidden;
+  /* Firefox fix */
+  border: 0;
+  clip: rect(0 0 0 0);
+  height: 1px;
+  margin: -1px;
+  overflow: hidden;
```
Aivean, line 76: Overlooked properties duplication.

PR #273, opened 2023-02-20
username suggestions in textarea when typing @ symbol

pazoozoo42:
When `@` symbol is typed and then any letter after it -- usernames suggestions list is opened at the cursor position. You can select a username from the list with mouse or arrow up and arrow down keys, clicking or pressing Enter on the selected item will paste it to the cursor position

https://user-images.githubusercontent.com/57953386/219991886-cd98d908-edfb-46db-b2d3-d0f8f03dec6e.mov


backend/src/api/UserController.ts
```diff
@@ -369,4 +376,18 @@ export default class UserController {
...
+    async suggestUsername(request: APIRequest<SuggestUsernameRequest>, response: APIResponse<SuggestUsernameResponse>) {
+        if (!request.session.data.userId) {
+            return response.authRequired();
+        }
+        const {start} = request.body;
+        try {
+            const usernames = await this.userManager.getUsernameSuggestions(start);
+            return response.success({usernames});
+        } catch (error) {
+            this.logger.error('Could not update user bio', { error });
```
iamwavecut:
```suggestion
            this.logger.error('Could not get usernames suggestions', { error });
```

backend/src/api/UserController.ts
```diff
@@ -369,4 +376,18 @@ export default class UserController {
...
+            return response.error('error', `Could not update bio`, 500);
```
iamwavecut:
```suggestion
            return response.error('error', `Could not get usernames suggestions`, 500);
```

backend/src/api/UserController.ts
```diff
@@ -77,6 +78,11 @@ export default class UserController {
             keyGenerator: (req) => String(req.session.data?.userId)
         });
 
+        const suggestUsernameLimiter = rateLimit({
```
Aivean, line 81: I'd suggest to make the limiter per user and use a larger window. Current settings don't prevent from DOS attck by constantly sending 10 rps from a single account.

backend/src/db/repositories/UserRepository.ts
```diff
@@ -261,4 +262,11 @@ export default class UserRepository {
             gender
         });
     }
+
+    async getUsernameSuggestions(start: string): Promise<string[]> {
+        return await this.db.fetchAll<string>(
```
Aivean: That probably works as a first version, but it's better to cache all usernames in memory, ideally in [a Trie](https://en.wikipedia.org/wiki/Trie).

backend/src/managers/UserManager.ts
```diff
@@ -659,4 +659,13 @@ export default class UserManager {
             return false;
         }
     }
+
+    async getUsernameSuggestions(startsWith: string): Promise<string[]> {
+        const cached = this.userCache.getUsernameSuggestion(startsWith);
```
Aivean: Note: this variant of the cache can grow without limit. It's possible to exhaust server's RAM by generating queries with random username prefixes.

I'd suggest to either use LRU cache instead of map, or, better, switch to the Trie (see comment above).

frontend/package.json
```diff
@@ -15,24 +18,24 @@
...
     "react-app-rewired": "^2.2.1",
+    "react-circular-progressbar": "^2.1.0",
+    "react-confirm-alert": "^3.0.6",
     "react-dom": "^18.0.0",
     "react-hook-form": "^7.29.0",
     "react-icons": "^4.3.1",
     "react-router-dom": "^6.3.0",
     "react-scripts": "5.0.0",
+    "react-textarea-autosize": "^8.3.4",
+    "react-textarea-with-suggest": "pazoozoo42/react-textarea-with-suggest#support-forwarded-ref",
```
Aivean:
```suggestion
    "react-textarea-with-suggest": "pazoozoo42/react-textarea-with-suggest#7168b69",
```
Let's use the commit hash here, to ensure that unintended changes don't get accidentally deployed.

frontend/package.json
```diff
@@ -15,24 +18,24 @@
...
     "react": "^18.0.0",
     "react-app-rewired": "^2.2.1",
+    "react-circular-progressbar": "^2.1.0",
+    "react-confirm-alert": "^3.0.6",
     "react-dom": "^18.0.0",
     "react-hook-form": "^7.29.0",
     "react-icons": "^4.3.1",
     "react-router-dom": "^6.3.0",
     "react-scripts": "5.0.0",
+    "react-textarea-autosize": "^8.3.4",
```
Aivean:
```suggestion
```

Aivean (commented):
I think UX could be improved, if the first result in the suggestion list is selected by default. This way, if it's a single result, autocompletion could be triggered with a single "Enter".

Not sure if it's easy to implement though.

Aivean:
Another suggestion (out of scope of this PR, something to think about in the future):

There are too many usernames returned by the 0-1 character queries, so suggestions are not very useful. Instead of showing the top of the global list, we can show more relevant usernames, for example, participants of the current comment thread.

Aivean:
Found a bug/weirdness. When comment text is restored from local storage, it triggers autocompletion, but with a last character?
username-suggest-demo
To reproduce:
* create a comment reply with partial username mention
* refresh the post page
* open that reply

---
Also, sometimes, when entering a post and immediately starting to type the reply, autocompletion doesn't trigger. Not sure how to reproduce reliably.

---

Also, note, all test here are performed with added 500ms latency in dev tools.

frontend/package.json
Aivean: Actually, your PR got accepted (https://github.com/marylorian/react-textarea-with-suggest/pull/8), we can use the npm version.

iamwavecut:
> Another suggestion (out of scope of this PR, something to think about in the future):
> 
> 
> 
> There are too many usernames returned by the 0-1 character queries, so suggestions are not very useful. Instead of showing the top of the global list, we can show more relevant usernames, for example, participants of the current comment thread.

+1 on this one. 

The only comment is that we should not, at all, request this local list, as we already have it in the FE storage. 

Would reduce amount of requests to the BE drastically and possibly will improve UX.

Aivean:
Another idea for the future: how about we sort results (that match the query) by the number of user's comments? The rationale is that users that comment more frequently are more likely to be mentioned.

pazoozoo42:
- changed cache to `trie` 
- Implemented automated highlighting if there's only one item in the list
- changed rate limit

> Also, sometimes, when entering a post and immediately starting to type the reply, autocompletion doesn't trigger. Not sure how to reproduce reliably.

- decreased debounce callback time from 100ms to 50ms, it should help

> Found a bug/weirdness. When comment text is restored from local storage, it triggers autocompletion, but with a last character?

was able to reproduce, but cannot figure what would be the fix for this yet

Aivean:
@pazoozoo42 
> Implemented automated highlighting if there's only one item in the list

Not sure that it works as intended. Right now sometimes the first element is highlighted even there are more items in the list.  My guess is that happens, when you type something, get results, select one, then erase what you typed to get a different list.

But anyway, I'd say, we should always highlight the first element (unless other element is already highlighted), even if there are more than one result. Rationale:
1. Saves one keypress
2. No reasons against that
3. Consistency

Aivean:
Found another bug. Currently for Russian language the component requests completions only for the last character.

Aivean (changes requested):
I think we need to address the bug with Russian usernames, other issues/improvements can be fixed later.

pazoozoo42:
> Not sure that it works as intended. Right now sometimes the first element is highlighted even there are more items in the list. My guess is that happens, when you type something, get results, select one, then erase what you typed to get a different list.

it's most probably that it is highlighted by the mouse hover

Aivean:
> it's most probably that it is highlighted by the mouse hover

No, I can consistently reproduce it by:
* starting to type (no selection)
* selecting one of the results
* erasing everything
* starting to type again (selection is already there)

pazoozoo42:
> No, I can consistently reproduce it by:
    starting to type (no selection)
    selecting one of the results
    erasing everything
    starting to type again (selection is already there)

Is this a blocker?

Aivean:
> Is this a blocker?

No, I don't think so.

Aivean (approved):
Looks good. Thanks again for this amazing feature, @pazoozoo42 !

PR #276, opened 2023-02-27
Fix for auto suggest removing text after a newline

pazoozoo42:
the component was fixed, this just changes where to load it from


Aivean:
Superseded by #277

PR #278, opened 2023-03-02
Add the ability to run manual sql interventions during the autodeploy

Aivean:
image

Example test run:
success: https://github.com/spaceshelter/orbitar/actions/runs/4319844489
non existing file: https://github.com/spaceshelter/orbitar/actions/runs/4319904822/jobs/7539594971

### DIFF
--- a/.github/workflows/deploy-universal.yml
+++ b/.github/workflows/deploy-universal.yml
@@ -17,6 +17,10 @@ on:
         description: 'Restore DB?'
         type: boolean
         default: false
+      manual-sql:
+        description: 'SQL intervention file path'
+        type: string
+        default: ''
       reindex-elastic:
         description: 'Reindex elasticsearch?'
         type: boolean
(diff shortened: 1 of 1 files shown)


Lexy2:
Assuming those SQL interventions are relatively small, otherwise, they would be database migrations, why request a file name instead of a raw SQL?

Aivean:
> Assuming those SQL interventions are relatively small, otherwise, they would be database migrations, why request a file name instead of a raw SQL?

To prepare the file beforehand and verify it on staging, reduces the chance of human error due to copy-pasting, etc.

Also, we probably don't want to show the content of these files in logs.

Lexy2:
Looks okay to me with one note:
Supplying file with bad contents effectively renders the site unusable as it doesn't restart the backend. But maybe it's good. You wouldn't want the backend to break the whole database if intervention script couldn't finish successfully.

Lexy2 (approved):

Lexy2:
I would rather make a separate wf to apply these things safely than risk breaking something during deployment when you can be unsure what caused the breakage.
There is also no history of changes when you apply those small scripts, they aren't saved anywhere.

Aivean:
> I would rather make a separate wf to apply these things safely than risk breaking something during deployment when you can be unsure what caused the breakage. There is also no history of changes when you apply those small scripts, they aren't saved anywhere.

That is all true and valid. However, this is intended as an improvement over the current approach of running scripts manually. 

I don't think we'd rely on this heavily in the future. If we do, we'll design a better solution.

PR #279, opened 2023-03-05
stop other videos

maximt:
Only one video at a time is allowed to play, other videos will be paused. 
Works with video and YouTube.
YouTube Iframe API is used for playback events.

Edit:

Added IntersectionObserver wrapper to track element's visibility and stop invisible videos that are scrolled out of the viewport.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -16,6 +16,7 @@
         "@types/react-dom": "^17.0.14",
         "@types/textarea-caret": "^3.0.1",
         "@types/webscopeio__react-textarea-autocomplete": "^4.7.2",
+        "@types/youtube": "^0.0.47",
         "@webscopeio/react-textarea-autocomplete": "^4.9.2",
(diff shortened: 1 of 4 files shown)


Aivean:
Sorry, didn't get the chance to review this (yet). Will do this week.

pazoozoo42 (dismissed):
This is a great improvement, thank you! Tested locally and it works as expected

Aivean:
One thing I'm thinking of is reverting the videos to "lightweight embed" form instead of stopping (for the videos that have this available, e.g. all youtube videos and some mp4) to save the RAM.

maximt:
> One thing I'm thinking of is reverting the videos to "lightweight embed" form instead of stopping (for the videos that have this available, e.g. all youtube videos and some mp4) to save the RAM.

Actually, videos are paused but not stopped, and you can resume playback from the last position, I think its the most expected behavior.

Aivean (approved):

PR #283, opened 2023-03-11
Fix orbitarMediaPoster for url with /raw at the end

maximt:
Currently, urls like "https://site/123.mp4" and "https://site/123.mp4/raw" will be rendered differently.
Small patch to fix this behavior.


Aivean (approved):

Aivean:
Found a minor bug:

Works:
```
<video src="https://host/hash.mp4/raw"/>
```

Doesn't:
```
https://host/hash.mp4/raw
```

Need also to accept `/raw` suffix here:
https://github.com/spaceshelter/orbitar/blob/6c901ec9df8cdfeea94c0b51ea52e7f9f4b787b4/backend/src/parser/TheParser.ts#L213

PR #284, opened 2023-03-14
Seeing settings page of another user does not make sense, this disables it

pazoozoo42:
API endpoints patching is not needed

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -129,6 +129,7 @@ const Ready = observer(() => {
                     <Route path="u/:username">
                         <Route path="" element={<UserPage />} />
                         <Route path=":page" element={<UserPage />} />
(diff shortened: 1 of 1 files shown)


Aivean:
I think we should disable it in the corresponding API controller (just return status code when requesting user doesn't match requested user).

This is better, because:
1. Currently for your own user  /u/page and /profile/page are synonyms. Your PR will break that.
2. Your change leaves API access unrestricted.

Aivean (approved):

Aivean:
After some consideration, it's better to disable this page in the View: 
https://github.com/spaceshelter/orbitar/blob/d520df112c846aea6044e231cf953794d9b378b1/frontend/src/Pages/UserPage.tsx#L34

e.g. 
https://stage.orbitar.space/u/orbitar/settings
https://stage.orbitar.space/u/orbitar/invites1

Should have identical result.

PR #286, opened 2023-03-20
Deduplicate notifications when mentioning user multiple times

Aivean:
Tested three cases:
* multiple mentions in the post
* multiple mentions in a standalone comment
* multiple mentions in the reply comment
* no mentions with all the above
* unrelated mentions with all the above

### DIFF
--- a/backend/src/managers/PostManager.ts
+++ b/backend/src/managers/PostManager.ts
@@ -15,6 +15,7 @@ import {CommentInfoWithPostData} from './types/CommentInfo';
 import {SiteInfo} from './types/SiteInfo';
 import {HistoryInfo} from './types/HistoryInfo';
 import TranslationManager from './TranslationManager';
+import {UserInfo} from './types/UserInfo';
 
 export default class PostManager {
     private bookmarkRepository: BookmarkRepository;
@@ -82,7 +83,7 @@ export default class PostManager {
 
         await this.bookmarkRepository.setWatch(postRaw.post_id, userId, true);
 
-        for (const mention of parseResult.mentions) {
+        for (const mention of new Set(parseResult.mentions)) {
             await this.notificationManager.sendMentionNotify(mention, userId, postRaw.post_id);
         }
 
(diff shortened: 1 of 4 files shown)


backend/src/managers/PostManager.ts
```diff
@@ -82,7 +83,7 @@ export default class PostManager {
 
         await this.bookmarkRepository.setWatch(postRaw.post_id, userId, true);
 
-        for (const mention of parseResult.mentions) {
+        for (const mention of new Set(parseResult.mentions)) {
```
4vanger: Wouldn't it be better to dedup `parseResults.mentions`?

backend/src/managers/PostManager.ts
```diff
@@ -244,12 +245,20 @@ export default class PostManager {
...
+            if (mention.toLowerCase() === parentAuthor?.username.toLowerCase()) {
+                continue;
+            }
             await this.notificationManager.sendMentionNotify(mention, userId, postId, commentRaw.comment_id);
         }
 
-        if (parentCommentId) {
-            await this.notificationManager.sendAnswerNotify(parentCommentId, userId, postId, commentRaw.comment_id);
+        if (parentAuthor) {
+            await this.notificationManager.sendAnswerNotify(parentAuthor.id, userId, postId, commentRaw.comment_id);
```
4vanger, line 263: Previously this was commentId and now we are sending userId - is that correct?

backend/src/managers/PostManager.ts
```diff
@@ -244,12 +245,20 @@ export default class PostManager {
...
         this.userManager.clearUserRestrictionsCache(userId);
 
-        for (const mention of parseResult.mentions) {
+        let parentAuthor: UserInfo | undefined;
+        if (parentCommentId) {
+            const parentComment = await this.commentRepository.getComment(parentCommentId);
+            parentAuthor = await this.userManager.getById(parentComment.author_id);
+        }
+        for (const mention of new Set(parseResult.mentions)) {
+            if (mention.toLowerCase() === parentAuthor?.username.toLowerCase()) {
```
4vanger:
```suggestion
            // if author of parent comment/post was mentioned - do not send notifications - they are already notified about answer to their comment
            if (mention.toLowerCase() === parentAuthor?.username.toLowerCase()) {
```

backend/src/managers/PostManager.ts
```diff
@@ -244,12 +245,20 @@ export default class PostManager {
         const commentRaw = await this.commentRepository.createComment(userId, postId, parentCommentId, content, language, parseResult.text);
         this.userManager.clearUserRestrictionsCache(userId);
 
-        for (const mention of parseResult.mentions) {
+        let parentAuthor: UserInfo | undefined;
+        if (parentCommentId) {
```
4vanger, line 249: When there is no `parentComment` - this is a top-level comment. Should we use post's author as parentAuthor to avoid duplicate mentions?

4vanger (commented):
few notes

backend/src/managers/PostManager.ts
Aivean: Yep, the signature has changed (part of the logic was extracted outside).

backend/src/managers/PostManager.ts
Aivean: Post author currently doesn't receive mentions for top level comments.

Aivean:
@4vanger, thanks for review, addressed the comments.

4vanger (approved):

PR #293, opened 2023-04-06
add Vimeo embed support

maximt:
Added simple iframe player and “stop-other-videos” feature.

### DIFF
--- a/frontend/src/Components/ContentComponent.tsx
+++ b/frontend/src/Components/ContentComponent.tsx
@@ -2,6 +2,7 @@ import React, {useEffect, useRef, useState} from 'react';
 import classNames from 'classnames';
 import styles from './ContentComponent.module.scss';
 import {createObserverService, observeOnHidden} from '../Services/ObserverService';
+import * as Vimeo from '@vimeo/player';
 
 interface ContentComponentProps extends React.ComponentPropsWithRef<'div'> {
     content: string;
@@ -11,6 +12,7 @@ interface ContentComponentProps extends React.ComponentPropsWithRef<'div'> {
 
 declare global {
     interface Window {
+        Vimeo: typeof Vimeo;
         YT: typeof YT;
         onYouTubeIframeAPIReady: () => void;
     }
@@ -30,6 +32,10 @@ function updateContent(div: HTMLDivElement) {
 
     div.querySelectorAll('video').forEach(video => {
         updateVideo(video);
+    });
+
+    div.querySelectorAll('iframe.vimeo-embed').forEach(iframe => {
+        updateVimeo(iframe as HTMLIFrameElement);
     });
 
     div.querySelectorAll('span.spoiler').forEach(spoiler => {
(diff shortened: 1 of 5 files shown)


frontend/src/Components/ContentComponent.tsx
```diff
@@ -128,6 +134,31 @@ function processVideoEmbed(img: HTMLImageElement) {
...
+
+function loadVimeoPlayer(iframe: HTMLIFrameElement) {
+    const attachVimeoPlayer = () => {
+        const player = new window.Vimeo.Player(iframe);
+        player.on('play', function() {
+            stopInnerVideos(document.body, iframe);
+        });
+    };
+
+    if (window.Vimeo) {
```
4vanger, line 158: See comment above - let's always load it async

frontend/src/Components/ContentComponent.tsx
```diff
@@ -2,6 +2,7 @@ import React, {useEffect, useRef, useState} from 'react';
 import classNames from 'classnames';
 import styles from './ContentComponent.module.scss';
 import {createObserverService, observeOnHidden} from '../Services/ObserverService';
+import * as Vimeo from '@vimeo/player';
```
4vanger: https://bundlephobia.com/package/@vimeo/player@2.18.0
21Kb gzipped - not horrible but so far we only had [1 vimeo link](https://orbitar.space/search?term=vimeo) so don't need to include it by default - if we can load it async - let's do that

frontend/src/Components/ContentComponent.tsx
```diff
@@ -213,6 +250,11 @@ function stopInnerVideos(el: Element, exept?: HTMLVideoElement | HTMLIFrameEleme
             return;
         stopVideo(iframe as HTMLIFrameElement);
     });
+    el.querySelectorAll('iframe.vimeo-embed').forEach(iframe => {
```
4vanger, line 266: 1) it will be the best to implement "stop when other video playing" functionality in separate PR and also don't make it Vimeo specific - we should look for all `video` elements and check if any of them is [currently playing](https://stackoverflow.com/questions/6877403/how-to-tell-if-a-video-element-is-currently-playing)

4vanger (commented):
I know it is WIP - just my 2 cents

frontend/src/Components/ContentComponent.tsx
maximt: Its just @types/vimeo__player for type safe compiling. Player is loaded on-demand.

frontend/src/Components/ContentComponent.tsx
maximt: done - PR #279 
HTMLVideoElement and YT are already supported.

frontend/src/Components/ContentComponent.tsx
maximt: Do you mean tag.async = true? Good idea, thanks.

frontend/src/Components/ContentComponent.tsx
4vanger: It may have unexpected effects of including `@vimeo/player` into bundle if we ever include it. You can explicitely instruct compiler that you only need types by using `import type`

frontend/src/Components/ContentComponent.tsx
4vanger: Shouldn't matter much at this point but will still help.
I incorrectly assumed initially that `window.Vimeo` object is defined in included `@vimeo/player` package

frontend/src/Components/ContentComponent.tsx
```diff
@@ -128,6 +134,31 @@ function processVideoEmbed(img: HTMLImageElement) {
...
+        const player = new window.Vimeo.Player(iframe);
+        player.on('play', function() {
+            stopInnerVideos(document.body, iframe);
+        });
+    };
+
+    if (window.Vimeo) {
+        attachVimeoPlayer();
+    } else {
+        const tag = document.createElement('script');
```
4vanger: to prevent duplicate script inclusion in slow network situation
```suggestion
    } else if(!document.getElementById('vimeo-script')){
        const tag = document.createElement('script');
        tag.id = 'vimeo-script';
```

Aivean (approved):

PR #294, opened 2023-04-08
Fix autocut not triggering, trigger blur and autocut for the personally downvoted posts and comments

Aivean:
Changes:
* fix bug when autocut was not triggered
* when YOU downvote the content, it's automatically blurred and cut (same behaviour as when it has rating <= -3)
* posts that YOU downvoted in the feed are collapsed to 200px height (as an alternative for the ignore on the post level)
* fixed bug in recreation of the observer maps: https://github.com/spaceshelter/orbitar/blob/d665eb841670f32b064e7f1c0fda70405af62bb3/frontend/src/Services/ObserverService.ts#L35

Collapsed post in the feed:
image

Downvoted comment:
image

### DIFF
--- a/frontend/src/Components/ContentComponent.tsx
+++ b/frontend/src/Components/ContentComponent.tsx
@@ -1,11 +1,11 @@
 import React, {useEffect, useRef, useState} from 'react';
 import classNames from 'classnames';
 import styles from './ContentComponent.module.scss';
-import {createObserverService, observeOnHidden} from '../Services/ObserverService';
+import {observeOnHidden} from '../Services/ObserverService';
 
 interface ContentComponentProps extends React.ComponentPropsWithRef<'div'> {
     content: string;
-    autoCut?: boolean;
+    autoCut?: number;
     lowRating?: boolean;
 }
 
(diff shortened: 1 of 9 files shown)


frontend/src/Components/ContentComponent.tsx
```diff
@@ -225,14 +228,25 @@ export default function ContentComponent(props: ContentComponentProps) {
...
+                        setCut(true);
+                        resizeObserver?.disconnect();
+                        resizeObserver = null;
+                    }
+                }
             }
+        };
+
+        if (props.autoCut) {
+            resizeObserver = new ResizeObserver(handleResize);
```
4vanger: What if content is just super long block of text without images? It will never resize so it won't collapse.
Need to force autocut first and if not cut- schedule `ResizeObserver`

Aivean: Lemmy test this.

Aivean: Seems to work just fine as is.

4vanger: Hm, unless there is autocut implemented somewhere else - I don't see how it may be working

Aivean: Ok, I'll rewrite this just in case.

Aivean: Fixed this and some other corner cases (mainly the interaction with the low rating blur).

4vanger (approved):

PR #296, opened 2023-04-08
Fix scroll jumping when answering

Aivean:
A bit hacky, but works.

Demo:
scroll2

### DIFF
--- a/frontend/src/Pages/PostPage.tsx
+++ b/frontend/src/Pages/PostPage.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect, useRef} from 'react';
+import React, {useEffect, useRef, useState} from 'react';
 import styles from './PostPage.module.css';
 import {Link, useLocation, useParams, useSearchParams} from 'react-router-dom';
 import {CommentInfo, PostInfo, PostLinkInfo} from '../Types/PostInfo';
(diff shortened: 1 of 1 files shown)


frontend/src/Pages/PostPage.tsx
```diff
@@ -54,16 +55,20 @@ export default function PostPage() {
...
         }
         else if (unreadOnly) {
             // find first new comment
              scrollToComment = document.querySelector<HTMLDivElement>(`.isNew`);
+             commentId = scrollToComment?.dataset.commentId ? parseInt(scrollToComment.dataset.commentId) : undefined;
         }
         // do nothing if element is focused already
-        if(!scrollToComment || scrollToComment.className.indexOf(styles.focusing) >= 0) {
+        if (!scrollToComment || scrollToComment.className.indexOf(styles.focusing) >= 0 ||
+            (scrolledToComment && scrolledToComment.postId === postId && scrolledToComment.commentId === commentId)
```
4vanger, line 70: `scrolledToComment` is a non-null here alread
```suggestion
            scrolledToComment.postId === postId && scrolledToComment.commentId === commentId
```

4vanger (approved):
Ok as a fix but easier fix will be removing `comments` dependency in line 101.
Adding a comment updates the array and triggers effect execution.
Other possible fix is when posting new comment - update hash to that comment ID and then it will scroll to newly posted comment
Approving to unblock with one small note

frontend/src/Pages/PostPage.tsx
Aivean: Oops, missed that. If you have time, please refactor it.

PR #299, opened 2023-04-09
Fix error when first time subscribing to a subsite

Aivean:
A trivial bug with 'undefined' access, when previous subscription record does not exist.

### DIFF
--- a/backend/src/api/SiteController.ts
+++ b/backend/src/api/SiteController.ts
@@ -107,6 +107,7 @@ export default class SiteController {
         }
         catch (error) {
             this.logger.error('Could not change subscription', { site, main, bookmarks, error });
+            this.logger.error(error);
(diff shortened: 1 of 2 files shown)


backend/src/api/SiteController.ts
```diff
@@ -107,6 +107,7 @@ export default class SiteController {
         }
         catch (error) {
             this.logger.error('Could not change subscription', { site, main, bookmarks, error });
+            this.logger.error(error);
```
4vanger, line 110: Forgot to remove debug? Double logging `error`

Aivean: Nope, this is intentional. The first line doesn't print the stacktrace, the second does.

4vanger (approved):

Issue #300, opened 2023-04-09
Add hotkeys to the editor formatting buttons


Aivean:
### Some standard hotkeys:
cmd + b -> bold
cmd + i -> italic
cmd + u -> underline
cmd+Shift+X -> strike-through
cmd + k -> link

PR #302, opened 2023-04-09
Editor hotKeys

eyesplitter:
Added all hotkeys from #300 

All hotkeys work with `ctrl` key for the people who love windows

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -31,6 +31,7 @@
     "react-confirm-alert": "^3.0.6",
     "react-dom": "^18.0.0",
     "react-hook-form": "^7.29.0",
+    "react-hotkeys-hook": "^4.3.8",
     "react-icons": "^4.3.1",
     "react-router-dom": "^6.3.0",
     "react-scripts": "5.0.0",
--- a/frontend/src/Components/CreateCommentComponent.tsx
+++ b/frontend/src/Components/CreateCommentComponent.tsx
@@ -25,7 +25,6 @@ import ReactTextareaAutocomplete from '@webscopeio/react-textarea-autocomplete';
 import TextareaAutosize from 'react-textarea-autosize';
 import debouncePromise from 'debounce-promise';
 import {useHotkeys} from 'react-hotkeys-hook';
-import {HotkeysEvent} from 'react-hotkeys-hook/dist/types';
 
 interface CreateCommentProps {
     open: boolean;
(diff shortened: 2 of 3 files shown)


Aivean:
@eyesplitter  Thank you for the contribution! The code looks good overall, however I found a bug when testing it locally. If a hotkey (say, `CMD-b`) is pressed when the form is not in focus it somehow gets "memorized" and triggered when the form gets focus and only the CMD key is pressed.

E.g. to reproduce:
1. Open reply textarea
2. Click outside it to make to lose focus
3. Press CMD-b
4. Click inside the textarea to make it gain focus
5. Press CMD  (just cmd)
6. Observe that `CMD-b` event is fired

Demo:
ezgif-1-abee070e90

Frankly I'm not sure what's going on, it maybe the bug in the `react-hotkeys-hook` library. Perhaps it would be easier to just extend the previous `handleKeyDown` approach?

frontend/package.json
```diff
@@ -31,6 +31,7 @@
     "react-confirm-alert": "^3.0.6",
     "react-dom": "^18.0.0",
     "react-hook-form": "^7.29.0",
+    "react-hotkeys-hook": "^4.3.8",
```
4vanger, line 34: 7.4Kb gzipped minifed. Not too small
https://bundlephobia.com/package/react-hotkeys-hook@4.3.8

frontend/src/Components/CreateCommentComponent.tsx
```diff
@@ -79,6 +96,7 @@ export default function CreateCommentComponent(props: CreateCommentProps) {
     const [isPosting, setPosting] = useState(false);
     const [previewing, setPreviewing] = useState<string | null>(null);
     const [mediaUploaderOpen, setMediaUploaderOpen] = useState(false);
+    const hotKeysRef = useHotkeys<HTMLTextAreaElement>(allowedKeys.join(','), (_, e) => handleHotKey(e), {enableOnFormTags: true});
```
4vanger: Why global catch VS recommended pattern from documentation:
```
useHotkeys<HTMLTextAreaElement>('ctrl+enter,meta+enter', handleAnswer, {enableOnFormTags: true});
...
```
Afair it is more readable and extendable this way

Aivean: I think it's related to the focus:
https://github.com/JohannesKlauss/react-hotkeys-hook#focus-trap

eyesplitter:
@Aivean 
I suppose that was a bug in the `react-hotkeys-hook` module.
This behavior works only with the meta key.
So I switched the handler from `HotkeysEvent` to `KeyboardEvent` and now it works as expected.

frontend/src/Components/CreateCommentComponent.tsx
eyesplitter: > I think it's related to the focus: https://github.com/JohannesKlauss/react-hotkeys-hook#focus-trap

Yes, it is.

frontend/package.json
eyesplitter: but the codebase looks better.

Aivean (approved):
LGTM

PR #305, opened 2023-04-11
Fix editor hotkeys stop working after preview

eyesplitter:
Moved the hotkeys ref element upper cuz we losing ref on conditional rendering


frontend/src/Components/CreateCommentComponent.tsx
```diff
@@ -355,27 +355,29 @@ export default function CreateCommentComponent(props: CreateCommentProps) {
...
-                            movePopupAsYouType={true}
-                            trigger={suggestTrigger}
-                        />
-                    </div>
-                :  <div className={classNames(commentStyles.content, styles.preview, postStyles.preview)} onClick={handleClosePreview}><ContentComponent content={previewing} /></div>
-            }
+            <div ref={hotKeysRef}>
+                {
+                    (previewing === null )
+                    ?  <div className={styles.editor} ref={containerRef}>
```
Aivean: Why not reuse `containerRef`?

Aivean (approved):

PR #306, opened 2023-04-12
Pre button in editor

LazyKarlson:
Button for <pre> HTML tag added to editor.
My VS Code add some spaces to code for readability thats why many changes in commit.


Aivean (changes requested):
Please get rid of the unrelated formatting changes. They do no match the formatting style of the codebase and potentially may conflict with PRs that other people are working on.

LazyKarlson:
> Please get rid of the unrelated formatting changes.

@Aivean Done!

Aivean:
@LazyKarlson, did you test it?

I think the parser support should be more involved, as right now it will still perform all replacements inside the `<pre>` tag (like videos and mentions).

Additionally, regex is not the right tool to do html escaping [here](https://github.com/spaceshelter/orbitar/pull/306/files#diff-d5aa2b4b04dd52bc19cbe923491046ba2bbc9f4ff6be8a40376fee71f846600fR212). Also, this logic shouldn't happen on the client. The user should see clear readable text inside the `<pre>` tag when writing/editing.

maximt:
I think [code] tag will be more expected and more well-known by users.

PR #307, opened 2023-04-12
Fix rating popup position behaviour

eyesplitter:
Instead of positioning the popup relative to the page, we position it relative to the rating block

Fixed:
1. bug when popup stays still on scrolling page
2. bug when popup moves on page resize (old bug)
3. bug when popup moves and renders incorrectly on changing the page font size (old bug)

Before:
[popup_before.webm](https://user-images.githubusercontent.com/3425688/231578629-59abe6f8-c81f-4a51-aab1-e18f38eff9f1.webm)

After:
[popup_after.webm](https://user-images.githubusercontent.com/3425688/231578655-42bf945c-7085-46b9-9d05-df2c87447c85.webm)

### DIFF
--- a/frontend/src/Components/RatingSwitch.tsx
+++ b/frontend/src/Components/RatingSwitch.tsx
@@ -225,7 +225,7 @@ const RatingList = React.forwardRef((props: RatingListProps, ref: ForwardedRef<H
     }
 
     return (
-        <div ref={ref} className={styles.list} onMouseDown={popupMouseDownHandler}>
+        <div ref={ref} className={` ${styles.list} ratingPopup`} onMouseDown={popupMouseDownHandler}>
             <div className={styles.listUp}>
                 <div className={listStyles.join(' ')}>{props.rating}</div>
                 <div className={styles.listDetails}>
--- a/frontend/src/Components/RatingSwitch.tsx
+++ b/frontend/src/Components/RatingSwitch.tsx
@@ -36,7 +36,6 @@ export default function RatingSwitch(props: RatingSwitchProps) {
         if (!ratingRef.current || !popupRef.current) {
             return;
         }
-        const [popupEl, ratingEl] = [popupRef.current, ratingRef.current];
         if (!showPopup) {
             return;
         }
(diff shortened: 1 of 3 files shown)


Aivean:
Looks pretty good overall, and definitely an improvement over what we had, however, there is a weird interaction with transparency in the "new comments" mode:
popup_transp

eyesplitter:
@Aivean fixed the opacity issue.
Now if the rating popup is opened we should remove the opacity from the comment.
This solution possibly [not working](https://developer.mozilla.org/en-US/docs/Web/CSS/:has#browser_compatibility) in Firefox

frontend/src/Components/RatingSwitch.tsx
```diff
@@ -52,27 +51,17 @@ export default function RatingSwitch(props: RatingSwitchProps) {
...
-        }
+        const rh = ratingEl.clientHeight;
+        const ph = popupEl.clientHeight;
 
-        popupEl.style.left = (nx) + 'px';
-        popupEl.style.top = (ny) + 'px';
+        const isTotalHeightMoreThanPageHeight  = y + rh + ph > document.documentElement.scrollHeight;
+        isTotalHeightMoreThanPageHeight
+            ? popupEl.style.bottom = 30 + 'px'
+            : popupEl.style.top = 30 + 'px';
```
4vanger: nit
```suggestion
        if(isTotalHeightMoreThanPageHeight){
            popupEl.style.bottom = '30px'
        } else {
            popupEl.style.top = '30px';
        }
```

eyesplitter: @4vanger should we add `no-ternary` to eslint?

eyesplitter: Oh, I see. `30 + '30px'`. 

 It is allowed to use ternary?

eyesplitter: After these changes, we have a huge gap. Is it ok?

Screenshot 2023-04-13 at 19 23 27

4vanger: > Oh, I see. `30 + '30px'`.
nope, it was my mistype. fixed.

> It is allowed to use ternary?

Ternary is perfectly fine - just assignment in ternary looks weird as for me. We don't have strict rules about codestyles unfortunately.

Aivean (approved):
Works pretty well! Thanks!

PR #303, opened 2023-04-10
Fix nav shift

eyesplitter:
Fix navigation shift after applying `overflow-y: scroll` to `body`

Before:
Screenshot 2023-04-11 at 00 59 50

After:
Screenshot 2023-04-11 at 00 59 35

PR #308, opened 2023-04-14
Revert latest scroll changes

Aivean:
Basically, this PR:
* reverts #303 and #304 
* slightly changes #296  (sets `overflow-y: scroll;` to `html` only)

Basically, this is a compromise. The slight gap in the Topbar (mentioned in #303) will be present, but layout shift are gone.

Most importantly, this fixes three regressions:
* scroll position when using "back" navigation
* Home/End gestures/hotkeys
* Scrollbar style

### DIFF
--- a/frontend/src/Components/UserProfilePosts.tsx
+++ b/frontend/src/Components/UserProfilePosts.tsx
@@ -47,8 +47,7 @@ export default function UserProfilePosts(props: UserProfilePostsProps) {
 
   const { posts, loading, pages, error, updatePost } = useFeed(props.username, 'user-profile', page, perpage, undefined, undefined, filter || '');
     useEffect(() => {
-        const element = document.querySelector('#app');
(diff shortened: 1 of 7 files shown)


frontend/src/Components/UserProfilePosts.tsx
```diff
@@ -47,8 +47,7 @@ export default function UserProfilePosts(props: UserProfilePostsProps) {
 
   const { posts, loading, pages, error, updatePost } = useFeed(props.username, 'user-profile', page, perpage, undefined, undefined, filter || '');
     useEffect(() => {
-        const element = document.querySelector('#app');
-        element && element.scrollIntoView({block: 'start'});
```
4vanger, line 51: `scrollIntoView` does nothing if element is already in view even when partially

4vanger (approved):

PR #312, opened 2023-04-17
Catch exceptions when storing cache to localStorage 

Aivean:
Should help with `quota exceeded` exception some users are facing:
image


4vanger (approved):
This won't make it worse but wouldn't this break any existing functionality relying on cache data?

Aivean:
> This won't make it worse but wouldn't this break any existing functionality relying on cache data?

It shouldn't as there is already a local in-memory cache in addition to the local storage.

PR #313, opened 2023-04-19
Fix scroll jumping when MediaUploader form is shown

Aivean:
Before:
scroll_before

After:
scroll_after


eyesplitter (approved):
Looks good

PR #314, opened 2023-05-01
Tighten restrictions for "exiled" users: allow to see only their own content

Aivean:
Now when user loses all permissions, their access to the content is limited only by their own posts:
* Can still see the list of subsites
* All feeds (including feeds by subsite) show only user's posts (every feed is the same)
* Feed sorting still works
* "Fire" works, but is limited to user's own posts
* Cannot see other people's profiles
* Doesn't have access to other people's posts, even by direct links

### DIFF
--- a/backend/src/api/FeedController.ts
+++ b/backend/src/api/FeedController.ts
@@ -1,7 +1,7 @@
 import FeedManager from '../managers/FeedManager';
 import {Logger} from 'winston';
 import {Router} from 'express';
-import {APIRequest, APIResponse, joiFormat, validate, joiSite} from './ApiMiddleware';
+import {APIRequest, APIResponse, joiFormat, joiSite, validate} from './ApiMiddleware';
 import Joi from 'joi';
 import UserManager from '../managers/UserManager';
 import SiteManager from '../managers/SiteManager';
@@ -70,11 +70,15 @@ export default class FeedController {
 
         try {
             const sorting = await this.userManager.getFeedSorting(userId, 1);
-            // const total = await this.feedManager.getSubscriptionsTotal(userId);
+            const restrictedPosts = await this.feedManager.getRestrictedPosts(userId, page, perPage, format, sorting);
+
             const {
                 total,
                 posts: rawPosts
-            } = await this.feedManager.getSubscriptionFeed(userId, page, perPage, format, sorting);
+            } = restrictedPosts ? {
+                total: await this.postManager.getPostsByUserTotal(userId),
+                posts: restrictedPosts
+            } : await this.feedManager.getSubscriptionFeed(userId, page, perPage, format, sorting);
             const { posts, users, sites } = await this.enricher.enrichRawPosts(rawPosts);
 
             response.success({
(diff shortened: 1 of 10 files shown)


backend/src/api/FeedController.ts
```diff
@@ -70,11 +70,15 @@ export default class FeedController {
...
+            const restrictedPosts = await this.feedManager.getRestrictedPosts(userId, page, perPage, format, sorting);
+
             const {
                 total,
                 posts: rawPosts
-            } = await this.feedManager.getSubscriptionFeed(userId, page, perPage, format, sorting);
+            } = restrictedPosts ? {
+                total: await this.postManager.getPostsByUserTotal(userId),
+                posts: restrictedPosts
+            } : await this.feedManager.getSubscriptionFeed(userId, page, perPage, format, sorting);
```
4vanger, line 81: General question for this PR - what is instead explicitely branching for restricte user content we embed restrictions into all manager methods? In this way we can be sure that there is no case missed that still fetches all the posts without restrictions. 
For example `getAllPostsTotal` will perform permission check inside and will return only user's posts

Aivean: Technically possible, although not consistent with the current approach (restrictions are already checked in controllers), also requires a bit more effort/refactoring, e.g. pair methods like `FeedManager.getAllPosts` / `FeedManager.getAllPostsTotal` would have to either calculate restrictions twice, or be joined into a single method.

Don't think it's worth the effort atm / could be refactored later if there is a valid reason to do so.

backend/src/db/repositories/PostRepository.ts
```diff
@@ -93,7 +95,7 @@ export default class PostRepository {
                      left join user_bookmarks b on (b.post_id = p.post_id and b.user_id = :for_user_id)
             where p.author_id = :user_id
                 ${filter ? ' and (p.source like :filter or p.title like :filter) ' : ''}
-            order by created_at desc
+            order by ${sorting === FeedSorting.postCommentedAt ? 'commented_at' : 'created_at'} desc
```
4vanger, line 98: bad readability - needed to reach to find `FeedSorting` declaration to find that this is enum. Also it will stop working as expected if more values will be added there. How about add a map with values:
```
const orderByMap: Record<keyof typeof FeedSorting, string> = {
  [FeedSorting.postCommentedAt]: 'commented_at',
  [FeedSorting.postCreatedAt]: 'created_at'
}
```

backend/src/api/UserController.ts
```diff
@@ -111,6 +111,10 @@ export default class UserController {
                 return response.error(ERROR_CODES.NOT_FOUND, 'User not found', 404);
             }
 
+            if (await this.userIsRestrictedToOwnContent(userId, profileInfo.id)) {
+                return response.error(ERROR_CODES.NOT_FOUND, 'User not found', 404);
```
4vanger: `403 Forbidden` will be more appropriate here imho

backend/src/api/FeedController.ts
4vanger: I see your point but it is still easy to miss something with this approach - for example `Post#translate` API is not limited and can be used to retrieve any comment or post currently

4vanger (commented):
First pass - still need to check controllers logic

backend/src/db/repositories/PostRepository.ts
Aivean: We already use this approach in other places, refactoring it is outside of the scope of this PR.

Aivean:
@4vanger, addressed (some) review comments, please take a look.

4vanger (approved):
Looks good but I haven't tried running it locally

PR #315, opened 2023-05-05
Prevent post bumping by exhiled users

Aivean:
When user has minimum rights, they comments won't bump the post (their last post), and mentions won't work (to prevent their abuse). The post will still show unread comments and appear in the "watch" feed.


4vanger:
Previously we were talking about not doing that to allow user to fix their karma. Now they will need someone else to find and up their posts which is probably too much to ask. Instead I think we should limit comments to 1 per day at some point

4vanger (approved):
It seems like people are preferring to have this - latest case: https://orbitar.space/p13077?new

Aivean:
I'm going to deploy this, as people abuse mentions.

PR #317, opened 2023-05-06
Update restrictions wording

Aivean:
Related to #314
image
image


Lexy2 (approved):
Rubber-stamped, just a text change.

4vanger:
@Aivean "Доступ ограничен только своими постами" should not be shown for the case when user has no posts

Aivean:
> @Aivean "Доступ ограничен только своими постами" should not be shown for the case when user has no posts

Debatable. The restriction is still there, even if user doesn't have any posts.

PR #318, opened 2023-05-07
Add coub support via lightweight embed

Aivean:
Demo:
coub-demo

### Implementation details:

1. Added a simple support for fetching coub previews to `orbitar-media`
    See [this commit](https://github.com/spaceshelter/orbitar-media/commit/918e2df7a9b1280854cdd9ebc9ec32f25faa3c7a)
    Urls like this: https://orbitar.media/coub/3808rb  redirect to the preview image

2. Added lightweight embed for coubs (works for links like https://coub.com/view/3808rb)
    When coub iframe goes out of screen, it's replaced back with the embed.


maximt:
Works fine, but maybe need to add support for stopVideo/stopInnerVideos and call it on click event?

Aivean:
> Works fine, but maybe need to add support for stopVideo/stopInnerVideos and call it on click event?

@maximt, good point, fixed, see the latest commit.

maximt (approved):

PR #320, opened 2023-05-09
Detect light/dark theme automatically from system settings for new users

Aivean:
When theme settings are not present in localStorage, tries to infer the default theme from system settings.

### Testing

Open orbitar.local in incognito tab, observe the dark theme.


maximt (approved):
tested in firefox and chromium, works fine

PR #330, opened 2023-05-29
Always trigger reload when home "monster" icon is clicked

Aivean:
### Changes

* Adds a mechanism to force-remount page component.
* "Monster", "Watch" and some other buttons now trigger refresh

### Demo:
reload-demo


eyesplitter (approved):
LG

maximt (approved):

PR #332, opened 2023-05-30
Add updating post to fetch new comments when attempting to scroll...

Aivean:
…to a comment that is not fetched yet.

Demo:
notif-demo

Tested some common scrolling scenarios, including scrolling to not existing hash.

### DIFF
--- a/frontend/src/Pages/PostPage.tsx
+++ b/frontend/src/Pages/PostPage.tsx
@@ -20,6 +20,10 @@ export default function PostPage() {
     const unreadOnly = search.get('new') !== null;
     const {post, comments, anonymousUser, postComment, editComment, editPost, error, reload, updatePost} = usePost(site, postId, unreadOnly);
 
+    // indicates that scrollToComment wasn't able to find the comment, and we've tried fetching it
+    // needed to prevent infinite loop
+    const [fetchingComment, setFetchingComment] = useState<number | undefined>(undefined);
+
     useEffect(() => {
         let docTitle = `Пост #${postId}`;
         if (post) {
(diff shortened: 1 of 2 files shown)


frontend/src/Pages/PostPage.tsx
```diff
@@ -59,6 +63,11 @@ export default function PostPage() {
         if (location.hash) {
             commentId = parseInt(location.hash.substring(1));
             scrollToComment = document.querySelector<HTMLDivElement>(`[data-comment-id="${commentId}"]`);
+
+            if (!scrollToComment && fetchingComment !== commentId) {
+                reload(unreadOnly).then(() => setFetchingComment(commentId));
```
4vanger, line 68: I got your idea but this will reload the post by just clicking at the link which may be not something user wants - I often skip reading answers in big posts because I don't have enough time to read all of them right now. At least we should ask for user's confirmation
Second scenario - comment is not on the page because it is in read branch - in this case reloading won't show it.

Aivean: I think both cases can be fixed by relying on the last read timestamp/comment instead of the materialized "isNew" flag.

This allows performing the comment presence check easily, as well as loading the "new" comments without removing existing "new" comments from the page.

4vanger:
I'm almost done with this proposal ;)

Aivean:
@4vanger 
> I'm almost done with this proposal ;)

An idea for that:

Instead of having two statuses for comments:  "read" and "new", have three:
* "read"
* "previously new"
* "new"

When page is loaded originally, comments are either "read" or "new".

However, user has the ability to refresh the page or or follow notification link, which loads a batch of fresh comments. To distinguish between them and "new" comments that were already on the page, I suggest to introduce a new status with working name "previously new".

When fresh comments are added to the page, the currently "new" comments will transition to the status "previously new", and freshly added comments will become "new".

Obviously, both statuses are calculated based on the corresponding timestamp/last comment id.

PR #336, opened 2023-06-16
Restrict votes to 100 requests per day for each (voterId, userId) pair

Aivean:


4vanger (approved):
as for me `100` number is a bit high but that's something

PR #340, opened 2023-07-31
UserProfileName - keyboard support added

maximt:
Save the new name by hitting enter key.

### DIFF
--- a/frontend/src/Components/UserProfileName.tsx
+++ b/frontend/src/Components/UserProfileName.tsx
@@ -5,6 +5,7 @@ import {ReactComponent as EditIcon} from '../Assets/edit.svg';
 import {ReactComponent as SendIcon} from '../Assets/send.svg';
 import {useAPI} from '../AppState/AppState';
 import {toast} from 'react-toastify';
+import {useHotkeys} from 'react-hotkeys-hook';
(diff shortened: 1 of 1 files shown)


Aivean (approved):

frontend/src/Components/UserProfileName.tsx
```diff
@@ -15,6 +16,7 @@ export default function UserProfileName(props: UserProfileNameProps) {
   const [editing, setEditing] = useState(false);
   const [name, setName] = useState(props.name);
   const api = useAPI();
+  const refEditName = useHotkeys<HTMLInputElement>('enter', () => handleEditNameComplete(), {enableOnFormTags: ['INPUT'], preventDefault: true});
```
4vanger, line 19: we don't care much about it but I don't think we should skip on this - this is not accessible - instead we should have a wrapping `form` element and then clicking enter or using voice browser command will submit. In addition phone browser will recognize that they are submitting the form and will replace "enter" on virtual keyboard with "submit"

Aivean: Yeah, also had similar thoughts. Ideally we should have a reusable component that handles that.

PR #343, opened 2023-08-21
Reset to first page on change Feed Sorting 

maximt:
Small fix

### DIFF
--- a/frontend/src/Pages/FeedPage.tsx
+++ b/frontend/src/Pages/FeedPage.tsx
@@ -16,7 +16,7 @@ const FeedPage = observer(() => {
     const { site, siteInfo } = useAppState();
     const api = useAPI();
 
-    const [search] = useSearchParams();
+    const [search, setSearch] = useSearchParams();
 
     const matchRoutePosts = !!useMatch('/posts');
     const matchRouteAll = !!useMatch('/all');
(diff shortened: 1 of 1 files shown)


Aivean:
Not sure that this is a right way to achieve this.

Currently when sorting is changed, the `subscriptions` request (or other similar request that corresponds to the selected feed) is triggered once.

After this change, the state change causes chain reaction and the same request is triggered FOUR times.

To test:
1. go the second page of the current feed
2. change sorting

Can instead simply remove the "?page=..." parameter from the feed selection buttons?

maximt:
The problem is that any change of the page triggers useFeed, and changing of the page and sorting at the same time causes double call to useFeed.

Think we can encapsulate the separate fields feedType, sorting, page into object just like FeedFilterCriteria to be able to change many params at the same time.

Or have to ugly trick with one more flag just like: 
```
setBeginUpdate(true);
setPage(page);
setSorting(sorting);
setBeginUpdate(false);

...
useFeed(....) {
 if (beginUpdate)
  return;
 ...
}
```

Aivean:
I'll try to think about a more elegant way to fix this.

It also could be worth the effort to resurrect #53 and replace `useFeed` with mobX state, which simplifies these invariants significantly.

PR #345, opened 2023-09-07
Handle pasting image/video to comment box

maximt:
Opens MediaUploader and sets preview when user try to paste image\video to comment box, that is more intuitive behavior for user.
Peek 2023-09-07 17-43

### DIFF
--- a/frontend/src/Components/CreateCommentComponent.tsx
+++ b/frontend/src/Components/CreateCommentComponent.tsx
@@ -94,6 +94,7 @@ export default function CreateCommentComponent(props: CreateCommentProps) {
     const [isPosting, setPosting] = useState(false);
     const [previewing, setPreviewing] = useState<string | null>(null);
     const [mediaUploaderOpen, setMediaUploaderOpen] = useState(false);
+    const [mediaUploaderData, setMediaUploaderData] = useState<File | undefined>();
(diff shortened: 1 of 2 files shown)


frontend/src/Components/CreateCommentComponent.tsx
```diff
@@ -240,6 +241,27 @@ export default function CreateCommentComponent(props: CreateCommentProps) {
         suggestResults.style.setProperty('left', left.toString() + 'px');
     });
 
+    useEffect(() => {
+        document.addEventListener('paste', handlePaste);
```
Aivean: What happens if multiple comment fields are expanded?

Maybe we should just use `onPaste={handlePaste}` in textarea?

Aivean:
Awesome job overall, but please check that this works correctly when multiple `CreateCommentComponent`s are mounted on the page.

Aivean (approved):
LGTM.

During the testing, noticed some issues, but they are pre-existing and could be addressed later:
* Paste doesn't set focus to the "upload button". Ideally `Ctrl-V + Enter` and `Esc` should work.
* File paste doesn't seem to work (as well as multiple files paste). Looks like it sets the "upload url" field which in turn inserts that value instead of uploading the image(s).

PR #346, opened 2023-09-13
use client-side Google Translate

4vanger:
Replace translation implementation with client-side Google Translate - no backend, no quotas, no payments.
Tested locally but @ZMSkG would like to test it when it will be on staging

recording


frontend/src/Components/CommentComponent.tsx
```diff
@@ -77,17 +78,19 @@ export default function CommentComponent(props: CommentProps) {
...
         } else {
             setTranslation(false);
-            api.postAPI.translate(props.comment.id, 'comment')
-                .then(res => setTranslation(res.html))
-                .catch(() => {
-                    setTranslation(undefined);
-                    toast.error('Не удалось перевести');
-                });
+            try {
+                const html = await googleTranslate(props.comment.content);
```
Aivean, line 91: Let's memoize this. Some people tend to switch between translation and original several times.

Aivean (dismissed):

Aivean (approved):

PR #347, opened 2023-09-15
Fix --dim1 --dimX rgba value overflow

Aivean:
Fix incorrect color values leading to broken styles in the dark theme.
image


4vanger:
Have you verified in actual Chrome? Some other styles appeared broken as well

4vanger (approved):

Aivean:
> Have you verified in actual Chrome? Some other styles appeared broken as well

I haven't :)

PR #348, opened 2023-09-18
Fix watch feed notification about the posts from exiled users

Aivean:
#315 prevents post bumps from exiled users, but it also stops posts from being moved to the top of the "watch" feed, when such posts are watched.

This PR fixes that. Now comments from exiled users won't bump the main feed for everybody, but they will bump the posts in the "Watch"/"fireplace" feed.

### DIFF
--- a/backend/src/managers/FeedManager.ts
+++ b/backend/src/managers/FeedManager.ts
@@ -347,7 +347,8 @@ export default class FeedManager {
         return { main, bookmarks };
     }
 
-    async postFanOut(subsite_id: number, post_id: number, createdAt: Date | undefined, updatedAt: Date | undefined) {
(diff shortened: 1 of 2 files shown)


maximt (dismissed):

backend/src/managers/FeedManager.ts
```diff
@@ -347,19 +347,22 @@ export default class FeedManager {
         return { main, bookmarks };
     }
 
-    async postFanOut(subsite_id: number, post_id: number, createdAt: Date | undefined, updatedAt: Date | undefined) {
+    async postFanOut(subsite_id: number, post_id: number, createdAt: Date | undefined, updatedAt: Date | undefined,
+                     onlyDbUpdate= false) {
```
4vanger: nit
```suggestion
                     onlyDbUpdate = false) {
```

4vanger (dismissed):

maximt (approved):

PR #350, opened 2023-09-24
Support full screen image zoom via react-zoom-pan-pinch

Aivean:
Demo:

https://github.com/spaceshelter/orbitar/assets/2865203/703575e2-2524-4cad-8e8a-58bb2961a21a

The new zoom is enabled by default, but could be disabled in settings.

### DIFF
--- a/frontend/src/Components/ContentComponent.module.scss
+++ b/frontend/src/Components/ContentComponent.module.scss
@@ -93,6 +93,7 @@
   top: 35px;
   right: 30px;
   font-size: 40px;
+  cursor: pointer;
 }
 
 .cutCover {
--- a/frontend/src/Components/ContentComponent.module.scss
+++ b/frontend/src/Components/ContentComponent.module.scss
@@ -88,6 +88,13 @@
   }
 }
 
+.overlayCloseButton {
+  position: absolute;
+  top: 35px;
+  right: 30px;
+  font-size: 40px;
+}
+
 .cutCover {
   position: relative;
 }
--- a/frontend/src/Components/ContentComponent.tsx
+++ b/frontend/src/Components/ContentComponent.tsx
@@ -434,7 +434,7 @@ interface ZoomComponentProps {
 
 function ZoomComponent(props: ZoomComponentProps) {
     // need to account for retina displays
-    const minScale = Math.min(1, window.innerWidth / props.width / 2, window.innerHeight / props.height / 2);
+    const minScale = Math.min(1, window.innerWidth / props.width, window.innerHeight / props.height);
     const defaultScale = Math.min(window.innerWidth / props.width, window.innerHeight / props.height);
     const defaultTranslateX = (window.innerWidth - props.width * defaultScale) / 2;
     const defaultTranslateY = (window.innerHeight - props.height * defaultScale) / 2;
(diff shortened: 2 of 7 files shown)


frontend/src/Components/UserProfileSettings.tsx
```diff
@@ -23,6 +23,10 @@ export function getVideoAutopause(): boolean {
     return JSON.parse(localStorage.getItem('autoStopVideos') || 'false');
 }
 
+export function getLegacyZoom(): boolean {
+    return JSON.parse(localStorage.getItem('legacyZoom') || 'false');
```
4vanger:
```suggestion
    return localStorage.getItem('legacyZoom') === 'true';
```

frontend/src/Components/ContentComponent.module.scss
```diff
@@ -88,6 +88,13 @@
   }
 }
 
+.overlayCloseButton {
+  position: absolute;
+  top: 35px;
+  right: 30px;
+  font-size: 40px;
+}
```
4vanger, line 97:
```suggestion
  cursor: pointer;
}
```

frontend/src/Components/ContentComponent.tsx
```diff
@@ -400,6 +418,63 @@ export default function ContentComponent(props: ContentComponentProps) {
...
+            onClick={(e) => {
+                // check if click originated from this element
+                if ((e.target as HTMLElement).classList.contains('react-transform-wrapper')) {
+                    props.onExit();
+                }
+            }}
+        >
+            <TransformWrapper
+                initialScale={defaultScale}
+                // centerOnInit={true}
```
4vanger: remove comment

frontend/src/Components/ContentComponent.tsx
```diff
@@ -400,6 +418,63 @@ export default function ContentComponent(props: ContentComponentProps) {
...
+interface ZoomComponentProps {
+    src: string;
+    width: number;
+    height: number;
+    onExit: () => void;
+}
+
+function ZoomComponent(props: ZoomComponentProps) {
+    // need to account for retina displays
+    const minScale = Math.min(1, window.innerWidth / props.width / 2, window.innerHeight / props.height / 2);
```
4vanger: What sense to allow scaling to less than screen? 
```suggestion
    const minScale = Math.min(1, window.innerWidth / props.width, window.innerHeight / props.height);
```

4vanger (dismissed):

frontend/src/Components/ContentComponent.tsx
Aivean: I think it's reasonable to allow to downscale at least to the original size of the image, if it's less than full screen. However, it's true that I should use here the scale for the original size specifically. UPD: which is exactly what you suggested.

frontend/src/Components/ContentComponent.tsx
```diff
@@ -400,6 +418,63 @@ export default function ContentComponent(props: ContentComponentProps) {
...
+            onClick={(e) => {
+                // check if click originated from this element
+                if ((e.target as HTMLElement).classList.contains('react-transform-wrapper')) {
+                    props.onExit();
+                }
+            }}
+        >
+            <TransformWrapper
+                initialScale={defaultScale}
+                // centerOnInit={true}
```
Aivean:
```suggestion
```

4vanger (approved):
tested locally - works like a charm

frontend/src/Components/ContentComponent.module.scss
```diff
@@ -88,6 +88,14 @@
   }
 }
 
+.overlayCloseButton {
```
maximt, line 91: sorry for late comment, I think it would be nice to have a shadow for clear vision on a light background.
image
image

PR #354, opened 2023-11-03
Fix crash on translations error, remove prompt duplication for short content

Aivean:


backend/src/api/PostController.ts
```diff
@@ -499,8 +499,15 @@ export default class PostController {
                 msg = 'OpenAI error';
             }
             this.logger.error(err);
-            // TODO analyze OpenAI response here and show custom error message once we will know how "run out of money" response looks like
-            return response.error('error', msg, 500);
+            try {
+                response.error('error', msg, 500);
```
4vanger, line 503: does `response.error()` throws if there were chunks sent already?

backend/src/api/PostController.ts
```diff
@@ -499,8 +499,15 @@ export default class PostController {
...
+            } catch (err) {
+                // in case some chunks were already written
+                // should not happen now, just a precaution if the invariant in translateEntity is broken
+                this.logger.error('Error writing response');
+                this.logger.error(err);
+                response.end();
```
4vanger: We should still do `response.write('{"result":"error","code":"error"}')` - frontend looks for this message in every chunk and will popup an error in browser if it meets it
https://github.com/spaceshelter/orbitar/blob/ece8d877d0c216291dc1b262bc6f02d0d010f3fe/frontend/src/API/use/useInterpreter.ts#L71C39-L71C71

backend/src/managers/TranslationManager.ts
```diff
@@ -99,18 +99,29 @@ export default class TranslationManager {
...
-                        'Strive for the best quality. Answer in Russian if not specified otherwise.'
-                }
-            ],
+            );
+        }
+
+        // push all
+        messages.push(
+            {role: 'user', content: content},
+            {role: 'system', content: (`Текст выше. ${ short ? 'Инструкция:' : 'Напоминаю инструкцию:'}\n' + ${prompt}`)},
```
4vanger:
```suggestion
            {role: 'system', content: (`Текст выше. ${ short ? 'Инструкция:' : 'Напоминаю инструкцию:'}${prompt}`)},
```

backend/src/managers/TranslationManager.ts
```diff
@@ -156,6 +167,12 @@ export default class TranslationManager {
             return Promise.resolve();
         }
 
+        const readableGPTStream = await TranslationManager.interpreterString(prompt, content, temp);
+
+        // Important!
+        // Ensure that the first chunk is written after the stream is created,
+        // otherwise the error cannot be set in http response downstream
```
4vanger, line 174: It can - see comment above.

backend/src/api/PostController.ts
Aivean: Yes.

Aivean: There is no `response.error()`:
image

backend/src/api/PostController.ts
Aivean: I can add this, but it doesn't work as you expect:
image

Aivean: Ok, found how to make it work. Pushed.

backend/src/managers/TranslationManager.ts
Aivean: Preserving this invariant when we can works better.

backend/src/api/PostController.ts
4vanger: It is there - we are using it [everywhere](https://github.com/search?q=repo%3Aspaceshelter%2Forbitar%20response.error&type=code)

4vanger (approved):

PR #356, opened 2023-11-03
Better detect when to show annotation buttons

Aivean:
Parse html content on demand and strip the tags before checking the length. Display corresponding buttons only for appropriate content length.


frontend/src/Components/PostComponent.tsx
```diff
@@ -184,16 +185,18 @@ export default function PostComponent(props: PostComponentProps) {
 
                     <button onClick={toggleOptions} className={styles.options + ' ' + (showOptions ? styles.active : '')}><OptionsIcon /></button>
                     {showOptions &&
+                        <InterpreterWrapper content={props.post.content}>{(showAltTranslate, showAnnotate) =>
```
4vanger: Let's add `showAltTranslate` and `showAnnotate` into `useInterpreter` - we already have everything needed in there

Aivean: No, the whole purpose of this wrapper is to perform this logic on demand.  I'm actually concerned about the overhead of `useInterpreter`, but I don't have any easy way to optimize it atm.

Aivean:
@4vanger, refactored according to your suggestions.

frontend/package.json
```diff
@@ -45,7 +45,8 @@
     "use-debounce": "^8.0.4",
     "use-local-storage": "^3.0.0",
     "webfonts-loader": "^7.5.2",
-    "xss": "^1.0.14"
+    "xss": "^1.0.14",
+    "string-strip-html": "^13.4.3"
```
4vanger: How about using `xss` since we already have it in the bundle?  It can strip nodes completely with `stripIgnoreTagBody` option

frontend/src/API/use/useInterpreter.ts
```diff
@@ -14,15 +16,31 @@ export const ANNOTATE_LIMIT = 1024;
...
     const [cachedAnnotation, setCachedAnnotation] = useState<string | undefined>();
     const [streamingAltTranslation, setStreamingAltTranslation] = useState<string | undefined | null>();  // null is a special value indication that content is being fetched
     const [cachedAltTranslation, setCachedAltTranslation] = useState<string | undefined>();
     const [inProgress, setInProgress] = useState<boolean>(false);
 
+    const calcStrippedOriginalContentLength = useLazy(() =>
+        stripHtml(originalContent,
+            {
+                stripTogetherWithTheirContents: [
+                    'script', // default
```
4vanger: Also `img` and `video` to stop showing annotate for image-only posts

Aivean: `img` and `video` don't have content.

frontend/package.json
Aivean: Done.

However, note, that we used `stripHtml` before on backend and in other microservices, before `xss` was introduced. Ideally we should pick one and consolidate.

Aivean:
Addressed comments. Also found and fixed accidental mistake from the previous PR 🤦

4vanger (approved):

PR #357, opened 2023-11-03
Show both annotations and altTranslations at the bottom of the content

Aivean:
Also fix the way annotations are cached (original content is no longer cached).

### DIFF
--- a/backend/src/managers/TranslationManager.ts
+++ b/backend/src/managers/TranslationManager.ts
@@ -163,7 +163,7 @@ export default class TranslationManager {
 
         // if parsingResult without html tags and whitespace is too short, don't do anything
         if (stripHtml(parsingResult).result.trim().length < 6) {
-            write(parsingResult);
+            // return empty response
             return Promise.resolve();
         }
 
@@ -175,16 +175,8 @@ export default class TranslationManager {
 
         const fullResponse: string[] = [];
 
-        if (mode === 'annotate') {
-            fullResponse.push(parsingResult);
-            fullResponse.push('<br />');
-            fullResponse.push('<br />');
-        }
+        write(hint);
         fullResponse.push(hint);
-
-        for (const chunk of fullResponse) {
-            write(chunk);
-        }
 
         for await (const part of readableGPTStream) {
             const chunk = part.choices[0]?.delta?.content || '';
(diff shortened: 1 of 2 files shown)


backend/src/managers/TranslationManager.ts
```diff
@@ -163,7 +163,7 @@ export default class TranslationManager {
 
         // if parsingResult without html tags and whitespace is too short, don't do anything
         if (stripHtml(parsingResult).result.trim().length < 6) {
-            write(parsingResult);
```
4vanger, line 166: don't forget to remove debug

Aivean: ?

Aivean: It just returns empty response.

frontend/src/API/use/useInterpreter.ts
```diff
@@ -25,9 +25,17 @@ export function useInterpreter(originalContent: string, originalTitle: string |
 
     const altTitle = currentMode === 'translate' && cachedTitleTranslation ? xssFilter(cachedTitleTranslation) : undefined;
 
+    const mergeAnnotation = (cachedAnnotation?: string, streamingAnnotation?: string | null): string | undefined => {
```
4vanger:
```suggestion
    const mergeContent= (cachedContent?: string, streamingContent?: string | null): string | undefined => {
```

frontend/src/API/use/useInterpreter.ts
```diff
@@ -120,7 +128,7 @@ export function useInterpreter(originalContent: string, originalTitle: string |
             const rect = contentRef.current.getBoundingClientRect();
             const topbarHeight = document.getElementById('topbar')?.clientHeight;
 
-            if (currentMode === 'annotate') {
+            if (currentMode === 'altTranslate' || currentMode === 'annotate') {
```
4vanger, line 122: Actually we should avoid scrolling to the beginning of the comment if we are adding content right above the button use has just clicked

Aivean: When the content is added to the bottom, it's scrolling to the bottom. This is needed in case when autocat is expanded (although I'm not sure if it works well).

frontend/src/API/use/useInterpreter.ts
Aivean: Fixed

backend/src/managers/TranslationManager.ts
4vanger: Ah, sorry - for some reason thought you've commited local debug code

4vanger (approved):
LGTM but need to check autoscroll implementation for long posts/comments

PR #360, opened 2023-11-04
Watch and Unwatch button were swapped, this fixes it

Aivean:


4vanger (approved):
that's why it should be named `isWatched`!

PR #365, opened 2023-11-13
Fix long migration names leading to out-of-order application

Aivean:
Two db migrations mistakenly had the longer names, that were being trimmed internally during the migration process (using some weird logic), leading to them being in the end of the migration list.

This change renames the migration files, ensuring the backward-compatibility for the existing systems.

### DIFF
--- a/backend/migrations/20221021345323-fix-migrations.js
+++ b/backend/migrations/20221021345323-fix-migrations.js
@@ -1,0 +1,32 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+  * We receive the dbmigrate dependency from dbmigrate initially.
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function(db, callback) {
+    db.runSql("UPDATE migrations SET name = '/20221021345324-translation' WHERE name = '/22102134532452-translation'", [], callback);
+    db.runSql("UPDATE migrations SET name = '/20221026349872-fix-content-source-title' WHERE name = '/21026349872934-fix-content-source-title'", [], callback);
+
+  return null;
+};
+
+exports.down = function(db, callback) {
+    db.runSql("UPDATE migrations SET name = '/22102134532452-translation' WHERE name = '/20221021345324-translation'", [], callback);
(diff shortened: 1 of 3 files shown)


backend/migrations/20221021345323-fix-migrations.js
```diff
@@ -0,0 +1,32 @@
...
+  * This enables us to not have to rely on NODE_PATH.
+  */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
+exports.up = function(db, callback) {
+    db.runSql("UPDATE migrations SET name = '/20221021345324-translation' WHERE name = '/22102134532452-translation'", [], callback);
```
Lexy2, line 18: `db-migrate` seems to check the list of applied migrations before running the migrations per se, so if you apply this migration to the current database, the subsequent `translation` and `fix-content-source` migrations are applied **twice**.

Not a huge issue, but they will be listed in the `migrations` table twice, and I'm not sure how this will work with future migrations.

backend/migrations/20221021345324-translation.js
```diff
@@ -15,6 +15,17 @@ exports.setup = function (options, seedLink) {
...
 
 exports.up = async function (db) {
+    // skip if name '/20221021345324-translation' or '/22102134532452-translation'
+    // exists in migrations table
+    // temporary hack to fix migration order
+    const result = await db.connection.promise().query(`
+        select name from migrations where name = '/20221021345324-translation' or name = '/22102134532452-translation';
+    `);
+
+    if (result.length > 0) {
```
Lexy2: Always true, so this migration never executes `CREATE TABLE` in a blank DB.

backend/migrations/20221026349872-fix-content-source-title.js
```diff
@@ -19,6 +19,7 @@ exports.up = async function (db) {
     await db.runSql(`
         UPDATE content_source
         SET title = (SELECT title FROM posts WHERE posts.content_source_id = content_source.content_source_id)
+        WHERE title IS NULL or title = '';
```
Lexy2, line 22: 👍

Lexy2 (changes requested):

backend/migrations/20221021345323-fix-migrations.js
Aivean: Should not really matter, and I'm not sure whether there's a better way to fix the issue.

backend/migrations/20221021345324-translation.js
Aivean: Fixed, please review.

Lexy2 (approved):

PR #366, opened 2023-11-18
Fix rate limits for content votes, add hourly limit for minuses.

Aivean:
Fix: correct key calculation.

Current limits:
* 100 per day for all content votes.
* 30 per hour for minuses.

### DIFF
--- a/backend/src/api/VoteController.ts
+++ b/backend/src/api/VoteController.ts
@@ -8,6 +8,7 @@ import {VoteListRequest, VoteListResponse} from './types/requests/VoteList';
 import {VoteListItemEntity} from './types/entities/VoteEntity';
 import UserManager from '../managers/UserManager';
 import rateLimit from 'express-rate-limit';
+import {RateLimiterMemory} from 'rate-limiter-flexible';
 
 export default class VoteController {
     public router = Router();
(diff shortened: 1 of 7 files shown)


backend/src/api/VoteController.ts
```diff
@@ -55,7 +51,7 @@ export default class VoteController {
             id: Joi.number().required()
         });
 
-        this.router.post('/vote/set', this.voteRateLimiter, validate(voteSchema), this.voteDayRateLimiter, (req, res) => this.setVote(req, res));
+        this.router.post('/vote/set', this.voteRateLimiter, validate(voteSchema), (req, res) => this.setVote(req, res));
```
4vanger, line 54: nit
```suggestion
        this.router.post('/vote/set', this.voteRateLimiter, validate(voteSchema), this.setVote);
```

backend/src/api/VoteController.ts
```diff
@@ -76,6 +72,19 @@ export default class VoteController {
             const max = type === 'user' ? 2 : 1;
             const rangedVote = Math.max(Math.min(vote, max), -max);
 
+            const targetUserId = await this.voteManager.getUserIdByVote(id, type);
+            if (targetUserId && type !== 'user') {
+                const voteKey = `${userId}:${targetUserId}`;
+                try {
+                    await this.voteDailyRateLimiter.consume(voteKey);
+                    if (rangedVote < 0) {
```
4vanger, line 80: But why only for minuses? mass-plusing will have same effect on regulation

Aivean: > mass-plusing will have same effect on regulation

You mean zero effect?

backend/src/api/VoteController.ts
Aivean: Are you sure that `this` will remain properly bound? I'd prefer not to change this blindly without testing.

backend/src/api/VoteController.ts
4vanger: Yup - so why treat them differently

backend/src/api/VoteController.ts
4vanger: No, it wouldn't
```
this.setVote.bind(this)
```
but it is not that slick. Ignore.

4vanger (approved):

backend/src/api/VoteController.ts
Aivean: Solely because people are using them differently and complaining only about minuses.

PR #367, opened 2023-11-19
RSA-encrypted direct messages, attempt 2

Aivean:
Now the public key is stored on the server, created in settings:
image
image
image

When user has public key set, "secret message" button is available in the comment editor:
image

This button opens modal which allows to write a text message:
image

image

image

image

image


pazoozoo42:
I'm super excited about this feature, thank you! I fetched the branch and was able to make it work but didn't yet get to the code review itself. But I already have a couple of comments about the flow and UI: 
1) Need to make clear that password is separate from your account password and can be used to read private messages only (or call it PIN? or show INFO section right away). Initially I thought this is my main password input for inbox creation confirmation

2) need better explanation on how it works, it's not clear right away that you can send private messages to users who have created their private key **and only** as comments in **their** posts

3) need to show error on entering a wrong password, it's not clear what is going on when wrong password is entered

4) need to hide send icon for me in my own posts, this can lead to confusion

5) would be great to hide sender name of the private comment so that this could be used as a private inbox. Or having this as an option at least

Will be looking into code soon and will leave more feedback. 
Great improvement, thank you!

Aivean:
Thanks for the review and suggestions!

> 1. Need to make clear that password is separate from your account password and can be used to read private messages only (or call it PIN? or show INFO section right away). Initially I thought this is my main password input for inbox creation confirmation

Makes sense. 

> 2. need better explanation on how it works, it's not clear right away that you can send private messages to users who have created their private key **and only** as comments in **their** posts

Any ideas on how to achieve that? Need to brainstorm the better explanation. Also, not only in **their** posts, any direct reply should work.

> 3. need to show error on entering a wrong password, it's not clear what is going on when wrong password is entered

👍 

> 4. need to hide send icon for me in my own posts, this can lead to confusion

👍 

> 5. would be great to hide sender name of the private comment so that this could be used as a private inbox. Or having this as an option at least

This is not trivial. I guess barmalini can be used as a workaround. Please check the discord design discussion for the full fledged inboxes.

pazoozoo42:
> This is not trivial. I guess barmalini can be used as a workaround. Please check the discord design discussion for the full fledged inboxes.

Good point, I keep forgetting we have barmalini, this means there is already a way to send anonymously, nice

> Any ideas on how to achieve that? Need to brainstorm the better explanation. Also, not only in their posts, any direct reply should work.

I think just simple and clear message would work. Once we decide on the message, we also could test how clear it is by asking users unaware of the feature to let us know if the message is clear enough. ChatGPT suggestions: 

image

Aivean:
@pazoozoo42, addressed most review comments.

frontend/src/Components/CreateCommentComponent.tsx
```diff
@@ -96,11 +101,39 @@ export default function CreateCommentComponent(props: CreateCommentProps) {
...
+        api.postAPI.getPublicKeyByUsername(parentUserName)
+            .then(res => {
+                if (res.publicKey) {
+                    setParentPublicKey({
+                        publicKey: res.publicKey,
+                        username: parentUserName
+                    });
+                }
+            });
+    }, [props.parentAuthorUserName, props.comment]);
```
pazoozoo42, line 136: warning about missing hook dependencies: `api.postAPI` and `username`

backend/src/api/UserController.ts
```diff
@@ -14,7 +14,7 @@ import {
     UserSaveNameRequest,
     UserSaveNameResponse,
     UserSaveGenderRequest,
-    UserSaveGenderResponse
+    UserSaveGenderResponse, UserSavePublicKeyRequest, UserSavePublicKeyResponse
```
pazoozoo42: let's stick to one per line formatting

backend/src/parser/TheParser.ts
```diff
@@ -376,6 +380,57 @@ export default class TheParser {
...
+        this.removeInnerMailTagsRec(node);
+
+        const result = this.parseChildNodes(node.children);
+        // render node.children back to html
+        const rawNodes = node.children.map((n) => render(n, {
+            encodeEntities: false,
+        })).join('');
+
+        const text =  `<span class="i i-mailbox-secure secret-mailbox" data-secret="${secret}" `+
+            `data-raw-text="${toBase64(rawNodes)}">${result.text}</span>`;
```
pazoozoo42: is it possible to use `.toString("base64");` instead of a 3rd party library?

frontend/src/Components/ContentComponent.tsx
```diff
@@ -56,6 +78,123 @@ function updateContent(div: HTMLDivElement, setZoomedImg: (img: ZoomedImg | null
...
+    }
+
+    // just the text
+    const title = mail.innerText.trim();
+
+    const mailInnerHtml = mail.innerHTML;
+    let decoded = false;
+
+    mail.addEventListener('click', () => {
+        if (decoded || !cipher) {
```
pazoozoo42, line 175: I think this check can be done above with early return so that event handler would be just not set at all

frontend/src/Components/ContentComponent.tsx
```diff
@@ -56,6 +78,123 @@ function updateContent(div: HTMLDivElement, setZoomedImg: (img: ZoomedImg | null
...
+        } else if (j.to && j.to === currentUsername && j.toKey) {
+            encodedKey = j.toKey;
+            cipher = j.c;
+        } else if (j.from && j.from === currentUsername && j.fromKey) {
+            encodedKey = j.fromKey;
+            cipher = j.c;
+        } else if (!j.to) {
+            encodedKey = j.toKey;
+            cipher = j.c;
+        }
```
pazoozoo42, line 156: it is possible to modify this code on the client side and start guessing other's people's passwords. Ideally we need a check on the backend, that only sender or recipient are trying to decipher

Aivean: Doesn't really matter. All the encoded info is in the open — anybody can try to bruteforce. No protection against it, unless we move all encoding to the backend, but that's kinda against the point — to have encryption transparent and not send anything to the server.

frontend/src/Components/ContentComponent.tsx
Aivean: Can add an extra check, but variables (`let`) have to be double-checked in the closure.

backend/src/parser/TheParser.ts
Aivean: Good point, I'll try.

frontend/src/Components/CreateCommentComponent.tsx
Aivean: Both of these values shouldn't change. However, I wonder if that weirdness with two requests firing was due to them being present previously as dependencies. Try adding them back and see if the double firing reproduces.

frontend/src/Components/ContentComponent.tsx
Aivean: Nevermind that. If you look carefully, `cipher` is already checked before, and `decoded` can't be true before the callback.

pazoozoo42 (dismissed):

PR #368, opened 2023-11-23
Fix hidden menu tab order and search

pazoozoo42:
It is a bit annoying that when the menu is hidden and you are using browser page search then menu items are still counted by the browser as searchable elements and it shows that your search has results but you don't see them because they are in the menu that is hidden.

This also fixes the problem navigating with tab key, when the menu is hidden it is now excluded from tab order


Aivean (approved):
Whoa, haven't noticed it. Checked locally, works as expected.

PR #371, opened 2024-01-02
OAuth2 support

pazoozoo42:
Hello and sorry for the long PR. I will try to provide some details in the description to make it easier to review

## OAuth2 support:
- clients (apps) catalog in the user profile
- option to register your client
- option to mark you  client public to be available for other users
- `authorization_code` and `refresh_token` grants are supported
- there's also a simple sample app in PHP to test the flow

## Security: 
- access_token ttl is 7 days
- authorization code ttl is 5 minutes
- access token format is JWT, hash of it is saved into db, hash of refresh_token is saved in db
 - when a client is deleted by owner or it's authorization is removed by a user ("uninstall") then tokens are marked as revoked, revoke status is loaded to cache on backend start

## Running locally
Run migration and make sure to edit `.env` by adding these new variables (you need to generate the secret key or you this sample one): 
```
JWT_ISSUER=http://orbitar.local
JWT_SECRET_KEY=d39addb1c1fa559481285644ab5fd8d6b698f786d05eabf1a6e87d1ed769439723bcd442c3ae7b52a6e66de8509e1237e15a24f7503b99bfc86e65141bd053c0
```

Screenshots: 

**Empty catalog**
image

**Empty catalog on mobile**
image

**Register app on mobile**
image

**Register app on desktop**
image

**Newly created client app**
image

**Catalog of public or installed apps (if an app is private but you have it installed, you will see it with an option to uninstall)**
image

**Catalog on mobile**
image

**Consent screen**
image

**Consent screen on mobile**
image


backend/src/managers/OAuth2Manager.ts
```diff
@@ -0,0 +1,323 @@
...
+    } catch (error) {
+      this.logger.error('Error registering OAuth client', { error });
+      throw error;
+    }
+  }
+
+  async listClients(userId: number): Promise<OAuth2ClientEntity[]> {
+    try {
+      const clients = await this.oauthRepository.getClients(userId);
+      console.log()
```
Aivean: Leftover from debugging?

backend/src/oauth/TokenService.ts
```diff
@@ -0,0 +1,61 @@
+import jwt from 'jsonwebtoken';
+import crypto from 'crypto';
+
+const accessTokenExpiresInHours = 24;
+
+export default class TokenService {
+  static getSecretKey(): string {
+    return process.env.JWT_SECRET_KEY ||
+    'b32e958a7b81375fc041215956df893ee61a7bc6dcd937deaed2102ba9533858f6fb1cafb764afbaf096e4dfde5558fe310527fd1651ddec0df0a79884960dc6';
```
Aivean: Why is it hardcoded?

frontend/src/Components/OAuth2AppCardComponent.tsx
```diff
@@ -0,0 +1,219 @@
...
+    api.oauth2Api.updateClientLogo(client.id, url).then(() => {
+
+    }).catch(() => {
+      toast.error('Не удалось обновить логотип');
+    });
+  };
+
+  const { client, scope } = props;
+
+  console.log(client.isPublic);
```
Aivean: Please make sure to remove debug logging before merge.

frontend/src/Components/OAuth2ClientLogoComponent.module.scss
```diff
@@ -0,0 +1,19 @@
...
...7F+qQhrpyoylQZpUwuYJnjhSjfsJlCwaAvTL6MqTAvCjHZnnTVYKPSfe75gl2+dKd3oil55likwKQgG6CUoaVNOolFyDHEZCRSJGc5DK1Q9KGmBSJFBAkCvxIpDC2gJBRBVQZI7NDqA6pA81kQYzU+U9iqlNGkjzrF4N0LrGi4KrRhCknm+qCJs4yAFGSYkL92VWGfFU4HoAoKCl5yR+yAAE5TSuqsBpXFoRmqwoEgF+fClquCpaTgRzdB79GUc/uVbPjopoqW4AAADwQRx6T7F9La1qp1s1kL9AA/xb7GAtcgI+eHSEJTwvJAlS1BDR7lx7tNoBPqqAC/4CFbdCatEEZMEBinOKN9iZAT2f+dLm+FVIAqCqDCiCgAL8TXgBKNF8lFa8ADmsvZZlU12ne9Yl5XQ4GBjCBBTRAL6ybwAAwMNpG3pa/jV3IY10TVBxwAHKUc6oMMEdcFPV3nP9tjI4avAQBKBdPEVbIhBXYABIjIW+8JSVz1bsAFxdBAPuN24ybt17kJoFsD9ZxevnWSd0OgQGaDPGxUpyQFR8HbEIdggNW66UPM1TJmgnAAwpg3R8coAA2K5OVS4plh340v0GwS9k6HNghb03LCrCxdBRgszITkckIcTLgzFkADTzXBRXQQAH4yWbtdu1O5jLABTDQ5e5g4AIELP907fB8ED2jiFkJKIDONHyCaWVAAdfq353G7FU7/yxgFHBVARY8OckdgAEDLsAEGkCBDYT3UZQ2iKUtGN5JRUhBmNJU8AR1sB33KXjDJrZwcl2QXSsbX8wmiLOfrapoD2Ta1CYTqR1r6myj2NjeDrd4rC0QbIv7cFLq9rnLqCUk8XLd8K5OBBIrEh/FON5y4kyjOULHIOM72xHoQFhNUgEMpOzd/35UYhpg5I8woACo84vEJ07xilv84hjPuMY37pcFZGDf5gm5yEdO8pKb/OQoT7nKV87ylrv85TCPucxnTvOa2/zmOM+5znfO8577/OdAD7rQh070ohv96EhPutKFl870pjv96VCPutSnTvWqW/3qWM+61rfO9a57/etgD7vYx072spv97GhPu9rXzva2u/3tcI+73OdO97rb/e54z7ve9873vvv974APvOAHT/jCG/7wiE+84hfP+MY7/vGQj7zkJ0/5ylv+8pjPvOY3z/nOe/7zoA+96EdP+tKb/vSod0IIAAA7');
```
Aivean: Let's use Asset instead, unless you have a good reason to embed it.

oauth2-sample-client/README.md
Aivean, line 1: Let's move it into a separate repo under `spaceshelter` ?

backend/src/oauth/TokenService.ts
pazoozoo42: just as an example, I'll remove

oauth2-sample-client/README.md
pazoozoo42: I will just delete the folder before merging, this is for reviewers convenience only

Aivean: I think it's a good idea to have a working example somewhere, but a separate repo is (imho) a better place for that.

backend/.env
```diff
@@ -10,4 +10,6 @@ DEFAULT_TL_LANGUAGE=ru
 FEED_INDEX_HOST=feed
 FEED_INDEX_PORT=6767
 SEARCH_HOST=elasticsearch
-SEARCH_PORT=9200
\ No newline at end of file
+SEARCH_PORT=9200
+JWT_ISSUER=
```
Aivean: This is not the right place for these properties.

in `backend/.env` there are only fixed values specific to the backend. Values that are environment-dependent should be in the root `.env`. 

On the other hand, `backend/.env.development` contains all values, needed for the backed in local dev environment (as for local backend run the root .env is not used).

I know we should document this better.

--
TL;DR: move `JWT_ISSUER` and `JWT_ISSUER` from `backend/.env` into `.env.sample`,

but leave them in `backend/.env.development`.

docs/api/oauth2.md
Aivean, line 1: 👍

Aivean:
Bugs discovered so far:
1.  when app permissions are revoked on the client user's side, the app icon is "stuck" in the app catalogue, without ability to remove:
image

the "public" checkbox in the author's account seems to hide those app for the clients, but that's seems like a weird / unintended UX
image


Improvements (could be added later, but ideally before the release):
- [x]  restrict the ability to create / use apps for low karma users
- [x]  cleanup expired tokens from the DB

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,377 @@
...
+
+  constructor(db: DB) {
+    this.db = db;
+  }
+
+  /**
+   * node-oauth2-server model methods
+   * https://node-oauthoauth2-server.readthedocs.io/en/master/model/spec.html
+   * **/
+  async getClient(clientId: string, clientSecret: string) {
```
Aivean: Most of these methods are unused, should we remove them?

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,377 @@
...
+
+  async getClientById(id: number): Promise<OAuth2ClientRaw | undefined> {
+    return await this.db.fetchOne<OAuth2ClientRaw>('select * from oauth_clients where id=:id', { id });
+  }
+
+  async getClients(userId: number): Promise<OAuth2ClientRaw[]> {
+    return await this.db.fetchAll<OAuth2ClientRaw>(
+      `select 
+          oauth_clients.*,
+          oauth_consents.user_id as is_authorized,
```
Aivean: what is the purpose of this field?

backend/src/api/OAuth2Controller.ts
```diff
@@ -0,0 +1,340 @@
...
+   * Regenerate client secret in case it was compromised or lost
+   */
+  async regenerateClientSecret(request: APIRequest<OAuth2ClientManageRequest>, response: APIResponse<OAuth2ClientRegenerateSecretResponse>) {
+    if (!request.session.data.userId) {
+      return response.authRequired();
+    }
+
+    try {
+      const { id } = request.body;
+      const newSecret = await this.oauth2Manager.regenerateClientSecret(id, request.session.data.userId);
```
Aivean: we probably also need to revoke all current tokens?

backend/src/api/types/requests/OAuth2.ts
```diff
@@ -0,0 +1,66 @@
...
+export type OAuth2RegisterResponse = {
+  client: OAuth2ClientEntity;
+};
+
+export type OAuth2ClientRequest = {
+  clientId: string;
+};
+
+export type OAuth2ClientManageRequest = {
+  id: number;
```
Aivean: `id` is too ambiguous. let's rename all fields to clientId where relevant

backend/src/api/types/requests/OAuth2.ts
```diff
@@ -0,0 +1,66 @@
...
+export type OAuth2TokenResponse = {
+  token: OAuth2Token;
+};
+
+export type OAuth2ClientRegenerateSecretResponse = {
+  newSecret: string;
+};
+
+export type OAuth2ClientUpdateLogoUrlRequest = {
+  id: number;
```
Aivean: let's rename to clientId

backend/src/api/types/entities/OAuth2ClientEntity.ts
```diff
@@ -0,0 +1,19 @@
+import { UserInfo } from '../../../managers/types/UserInfo';
+
+export type OAuth2ClientEntity = {
+  id: number;
```
Aivean: if clientId is unique, why do we have numeric `id` here?

backend/src/api/ApiMiddleware.ts
```diff
@@ -102,3 +132,18 @@ export const joiFormat = Joi.valid('html', 'source').default('html');
 
 export const joiSite = Joi.string().max(siteDomainMaxLengthChars).regex(/^[a-z\d-]*$/i);
 export const joiSiteName = Joi.string().min(siteNameMinLengthChars).max(siteNameMaxLengthChars);
+
+export const urisListValidator = Joi.string().custom((value, helpers) => {
```
Aivean, line 138: would be nice to have tests for this

backend/src/db/repositories/OAuth2Repository.ts
Aivean: ah, nvm, I see they are used by `OAuthServer` which, btw, is not obvious

backend/src/db/types/OAuth2.ts
```diff
@@ -0,0 +1,68 @@
...
+  client_id: number;
+  client_client_id: string,
+  user_id: number;
+  scope: string;
+  revoked: number;
+}
+
+export interface OAuth2AuthorizationCodeRaw {
+  client_id: number;
+  client_client_id: string;
```
Aivean: really?

backend/src/app.ts
```diff
@@ -107,6 +112,14 @@ const voteRepository = new VoteRepository(db);
 const userRepository = new UserRepository(db);
 const webPushRepository = new WebPushRepository(db);
 const translationRepository = new TranslationRepository(db);
+const oauthRepository = new OAuth2Repository(db);
+
+app.oauth = new OAuthServer({
+    model: oauthRepository,
```
Aivean: let's not do blind casts here, as it hides incompatibility errors

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,384 @@
...
+      scope: token.scope
+    });
+
+    return {
+      accessToken: token.accessToken,
+      accessTokenExpiresAt: token.accessTokenExpiresAt,
+      refreshToken: token.refreshToken,
+      client: {
+        id: clientId,
+        grants: [] //FIXME return grants
```
Aivean: interface expects grants here

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,384 @@
...
+    if (!tokenFromDb) {
+      return null;
+    }
+    return {
+      accessToken,
+      accessTokenExpiresAt: new Date(tokenFromDb.access_token_expires_at),
+      scope: tokenFromDb.scope?.split(','),
+      client: {
+        id: tokenFromDb.client_client_id,
+        grants: [] //FIXME return grants
```
Aivean: and here

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,384 @@
...
+
+    return {
+      authorizationCode: code,
+      expiresAt: codeFromDb.expires_at,
+      redirectUri: codeFromDb.redirect_uri,
+      scope: codeFromDb.scope,
+      client: {
+        id: codeFromDb.client_client_id,
+        numeric_id: codeFromDb.client_id,
+        grants: [] //FIXME return grants
```
Aivean: here as well

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,384 @@
...
+      await this.db.query('update oauth_codes set expires_at = now() - 1 where code_hash=:code_hash', {
+        code_hash: TokenService.hashString(code.authorizationCode)
+      });
+      return true;
+    } catch (e) {
+      return false;
+    }
+  }
+
+  async getRefreshToken(refreshToken) {
```
Aivean: unused?

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,384 @@
...
+      client: {
+        id: tokenFromDb.client_client_id
+      },
+      user: {
+        id: tokenFromDb.user_id
+      }
+    };
+  }
+
+  async revokeToken(token) {
```
Aivean: unused?

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,384 @@
...
+      client_secret_hash: clientSecretHash,
+      initial_authorization_url: initialAuthorizationUrl,
+      redirect_uris: redirectUris,
+      user_id: userId,
+      grants: 'authorization_code,refresh_token',
+      is_public: isPublic ? 1 : 0
+    } as OAuth2ClientRaw;
+  }
+
+  async createToken(clientId: number, userId: number, accessTokenHash: string, accessTokenExpiresAt: Date, refreshTokenHash: string, scope: string): Promise<void> {
```
Aivean: unused?

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,384 @@
...
+    });
+  }
+
+  async getTokenByAccessTokenHash(accessTokenHash: string): Promise<OAuth2TokenRaw | undefined> {
+    return await this.db.fetchOne<OAuth2TokenRaw>(`select ot.*, oc.client_id, oc.grants as client_client_id from oauth_tokens ot, oauth_clients oc where ot.access_token_hash = :access_token_hash and ot.revoked != 1 and ot.client_id = oc.id`, {
+      access_token_hash: accessTokenHash
+    });
+  }
+
+  async getConsent(clientId: number, userId: number): Promise<OAuth2ConsentRaw | undefined> {
```
Aivean: unused?

backend/src/api/OAuth2Controller.ts
```diff
@@ -0,0 +1,340 @@
...
+    if (!request.session.data.userId) {
+      return response.authRequired();
+    }
+
+    try {
+      const {name, description, logoUrl, initialAuthorizationUrl, redirectUris, isPublic} = request.body;
+      const userId = request.session.data.userId;
+      const author = await this.userManager.getById(userId);
+
+      const client: OAuth2ClientRaw = await this.oauth2Manager.registerClient(name, description, logoUrl, initialAuthorizationUrl, redirectUris, userId, isPublic);
```
Aivean: let's limit the total number of created clients per user (say, 32 is enough) to avoid abuse

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,387 @@
...
+      id: clientId,
+      redirectUris: clientFromDB.redirect_uris.split(',').map(uri => uri.trim()),
+      grants: clientFromDB.grants.split(',').map(grant => grant.trim()),
+
+    };
+  }
+
+  async generateAccessToken(client, user, scope) {
+    const nowTs = Math.floor(Date.now() / 1000);
+    const accessTokenExpiresAtTs = nowTs + (parseInt(process.env.ACCESS_TOKEN_TTL_SECONDS, 10) || 3600 * 24 * 7);
```
Aivean: we have config for that

backend/src/oauth/TokenService.ts
```diff
@@ -0,0 +1,58 @@
...
+const accessTokenExpiresInHours = 24;
+
+export default class TokenService {
+  static getSecretKey(): string {
+    return process.env.JWT_SECRET_KEY;
+  }
+
+  static generateAccessToken(sub: string, aud: string, scope: string, exp: number, iat: number, authTime: number): string {
+    const payload = {
+      iss: process.env.JWT_ISSUER || 'https://orbitar.space',
```
Aivean: we have `Config` for env variables

backend/src/db/types/OAuth2.ts
pazoozoo42: wym?

pazoozoo42:
@Aivean, I’ve noticed several questions about the distinction between `id` and `client_id`. The key issue is that "Client ID" is a well-established term in OAuth, referring to a unique string generated when a client is registered. Meanwhile, in our project, we follow an `entity_id` convention for numeric auto-incremented IDs, such as `post_id`, `user_id`, etc.

This explains why I chose `id` as the internal numeric auto-incremented ID for clients, and why names like `client_client_id` exist and I agree this naming can be confusing.

I’m happy to discuss potential renaming if we find a better solution that balances clarity and consistency. In the meantime, I’d love to get more constructive PR comments. For example, simply commenting "Really" doesn’t provide actionable feedback and can be unhelpful. If something seems unclear or suboptimal, suggesting an alternative or explaining what exactly if confusing would speed up the review process and reduce unnecessary back-and-forth.

Let’s work together to refine this. Your help and feedback are always much appreciated, thank you!

Aivean:
@pazoozoo42 

> @Aivean, I’ve noticed several questions about the distinction between `id` and `client_id`. The key issue is that "Client ID" is a well-established term in OAuth, referring to a unique string generated when a client is registered. Meanwhile, in our project, we follow an `entity_id` convention for numeric auto-incremented IDs, such as `post_id`, `user_id`, etc.

I can see two ways how to resolve this:
1. get rid of the numeric id, use `client_id` as a PK (since it's unique)
2. rename `oauth_clients` to `oauth_apps` and have `app_id` as a numeric PK (leaving `client_id` as is).

I'm personally leaning towards (1), it's a simpler natural approach and currently `client_id` is relatively short, so there should not be **too** much of a bloat in data size. 

However (2) could be easier to implement (straightforward renaming) and it's more efficient from the schema normalization perspective.


>  In the meantime, I’d love to get more constructive PR comments. For example, simply commenting "Really" doesn’t provide actionable feedback and can be unhelpful. If something seems unclear or suboptimal, suggesting an alternative or explaining what exactly if confusing would speed up the review process and reduce unnecessary back-and-forth.

Sorry, I thought the explanation would be [redundant](https://gist.github.com/Aivean/d97341756b54ce2641ec64fcda8c63f8) in this case.


> Let’s work together to refine this. Your help and feedback are always much appreciated, thank you!

Let's do this! Thank you again for creating this huge feature. It wasn't easy to review, but I think I mostly went through it, once we fix the major issues (like the schema), we can merge it and then iterate on the minor things later.

backend/src/db/types/OAuth2.ts
Aivean: very confusing and inconsistent naming. please see my comment at the bottom with some suggestions how to address.

Aivean:
There is also one thing that doesn't make sense to me. If we are using JWT, why are we storing access and refresh tokens in DB (unless they are revoked)?

Aivean:
> The value of the scope parameter is expressed as a list of space-
   delimited, case-sensitive strings.

https://datatracker.ietf.org/doc/html/rfc6749#section-3.3

current implementation expects comma-separated scopes, which is invalid

UPD: fixed

pazoozoo42:
Converting to draft per our discussion: 
 - fix management UI bugs
 - don't store tokens and codes, use only two tables (clients and consents)
 - no numeric id for clients table, use client ID for management API
 - allow more than one consent per app for end users, each has its own scope
 - limit number of clients per developer

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,393 @@
...
+  OAuth2ClientRaw,
+  OAuth2ConsentRaw,
+  OAuth2TokenRaw
+} from '../types/OAuth2';
+import {ResultSetHeader} from 'mysql2';
+import TokenService from '../../oauth/TokenService';
+import {config} from '../../config';
+import {AuthorizationCode, AuthorizationCodeModel, Falsey, Token} from 'oauth2-server';
+
+export default class OAuth2Repository implements AuthorizationCodeModel {
```
Aivean: just a thought, it would be cleaner to move `AuthorizationCodeModel` implementation into a dedicated class (it can have  `OAuth2Repository` or `OAuth2Manager` as a field) and keep `OAuth2Repository` responsibility limited to DB layer

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -38,219 +35,195 @@ export default class OAuth2Repository implements AuthorizationCodeModel {
...
+   * As tokens are not persisted, nothing needs to be revoked.
+   */
+  async revokeToken(token): Promise<boolean> {
+    return true;
+  }
+
+  /**
+   * Verifies token scope.
+   *
+   * With token persistence removed, this method can't perform any real verification.
```
Aivean: it should decode `accessToken` and read the scopes from it.

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -38,219 +35,195 @@ export default class OAuth2Repository implements AuthorizationCodeModel {
...
-      scope: tokenFromDb.scope?.split(' ')?.filter(scope => scope.trim() !== ''),
-      client: {
-        id: tokenFromDb.client_client_id,
-        grants: [] //FIXME return grants
-      },
-      user: {
-        id: tokenFromDb.user_id
-      }
-    };
+    return null;
```
Aivean: it should decode access token and return it

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -38,219 +35,195 @@ export default class OAuth2Repository implements AuthorizationCodeModel {
...
+    return await this.db.fetchOne<OAuth2ClientRaw>(
+      'select * from oauth_clients where client_id = :client_id',
+      { client_id: clientId }
+    );
   }
 
+  /**
+   * Retrieves a client by its client_id and client secret hash.
+   */
   async getClientByClientIdAndClientSecretHash(clientId: string, clientSecretHash: string): Promise<OAuth2ClientRaw | undefined> {
```
Aivean, line 214: Not sure if this method is needed.

frontend/.env.development
```diff
@@ -5,4 +5,5 @@ BROWSER=none
 REACT_APP_VAPID_PUBLIC_KEY=
 REACT_APP_STATUS_UPDATE_INTERVAL=10
 DEFAULT_TL_LANGUAGE=ru
-VAPID_PUBLIC_KEY=
\ No newline at end of file
+VAPID_PUBLIC_KEY=
+JWT_SECRET_KEY=d39addb1c1fa559481285644ab5fd8d6b698f786d05eabf1a6e87d1ed769439723bcd442c3ae7b52a6e66de8509e1237e15a24f7503b99bfc86e65141bd053c0
```
Aivean: let's not commit secrets, even if fake ones

backend/migrations/20231222000022-oauth-setup.js
```diff
@@ -0,0 +1,51 @@
...
+        redirect_uris text not null,
+        grants varchar(255),
+        user_id int not null,
+        foreign key (user_id) references users(user_id) on delete cascade
+    ) engine=InnoDB default charset=utf8mb4;
+
+    create table if not exists oauth_consents (
+      user_id int not null,
+      client_id varchar(255) not null,
+      scope varchar(255),
```
Aivean: can be long, need to change to text

frontend/src/Types/OAuth2.ts
```diff
@@ -0,0 +1,322 @@
...
+  'user:savebio': 'менять био пользователя',
+  'user:savegender': 'менять пол пользователя',
+  'user:savename': 'менять имя пользователя',
+  'user:suggest-username': 'искать юзернейм пользователя',
+  'vote': 'все действия с голосами',
+  'vote:list': 'получать список голосов',
+  'vote:set': 'голосовать'
+};
+
+export const OAuth2ScopeEndpointsMap = {
```
Aivean: unused?

backend/src/api/OAuth2Middleware.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Session Integration:
+ *   - When authentication succeeds, the middleware updates session data
+ *     via the Session class.
+ *
+ * @param app Express application instance that has been configured with OAuth2.
+ * @param db Database instance (used for sessions persistence).
+ * @param logger Logger instance for logging errors and warnings.
+ * @returns An Express middleware function that handles OAuth2 authentication.
+ */
+export default function OAuth2Authenticate(app: Application, db: DB, logger: Logger) {
```
Aivean: this needs proper types

frontend/src/Types/OAuth2.ts
pazoozoo42: used in controllers, when OAuthMiddleware is generated for a route
https://github.com/spaceshelter/orbitar/blob/3d39b5ccc329f0e9756efb03b1950b87a6a23660/backend/src/api/FeedController.ts#L56-L60

Aivean: but there are two (identical?) `OAuth2ScopeEndpointsMap` instances, one in the frontend, one in the backend, and the frontend one appears to be unused

pazoozoo42: yes, confused this with backend, will check

backend/src/oauth/AuthorizationCodeModelImpl.ts
```diff
@@ -0,0 +1,394 @@
...
+        scope,
+        type
+      } = jwt.verify(refreshToken, process.env.JWT_SECRET_KEY) as JwtPayload;
+
+      const clientId = aud.toString();
+      const userId = parseInt(sub, 10);
+
+      // Verify token validity (returns Error object on error and grants on success)
+      const result = await AuthorizationCodeModelImpl.verifyTokenValidity(
+        this.repo,
```
Aivean, line 227: nit: since JS/TS doesn't have named params, it's a good practice to pass wrapper objects into functions with 5+ params

backend/src/oauth/AuthorizationCodeModelImpl.ts
```diff
@@ -0,0 +1,394 @@
...
+  static generateClientSecret = (): string => {
+    return crypto.randomBytes(32).toString('hex');
+  };
+
+  static hashString(value: string): string {
+    return crypto.createHash('sha256').update(value).digest('hex');
+  }
+
+  static async verifyTokenValidity(
+    repo: OAuth2Repository,
```
Aivean, line 310: why not make it non-static to avoid passing repo?

pazoozoo42: easier to write tests for it, but I haven't yet checked if we have db layer in tests, if not, I will make it non-static

backend/src/oauth/AuthorizationCodeModelImpl.ts
```diff
@@ -0,0 +1,394 @@
...
+    // Convert the required scope(s) into an array.
+    const requiredScopes = typeof scope === 'string'
+      ? scope.split(' ').filter((s) => s.trim())
+      : scope;
+
+    const valid = requiredScopes.every(requiredScope => {
+      return tokenScopes.some(tokenScope => {
+        // Exact match check.
+        if (tokenScope === requiredScope) {
+          return true;
```
Aivean, line 152: IMPORTANT! callback must be called, if provided, otherwise the request will hang

frontend/src/Components/UserProfileClientAppsCreateForm.tsx
```diff
@@ -48,17 +48,17 @@ export default function UserProfileClientAppsCreateForm(props: UserProfileClient
   const validateUrls = (value: string) => {
     const urls = value.split(',').map(url => url.trim());
     return urls.every((url) => isURL(url, {
-      ...(process.env.NODE_ENV === 'development' && { host_whitelist: ['localhost'] }),
+      ...(process.env.NODE_ENV === 'development' && { host_whitelist: ['localhost', '1ffe-2600-1700-1b0-23b0-b123-655f-7110-c44a.ngrok-free.app'] }),
```
Aivean: please remove ngrok urls

backend/src/api/OAuth2Middleware.ts
pazoozoo42: I think all three are correct, which one is wrong specifically?

Aivean: this whole thing `export default function OAuth2Authenticate(app: Application, db: DB, logger: Logger)` needs an explicit type, as it's currently passed as an untyped parameter to controllers

Aivean: and its intermediate result needs an explicit type as well

backend/src/api/OAuth2Controller.ts
```diff
@@ -161,7 +161,9 @@ export default class OAuth2Controller {
       }
     })(req, res, () => {}));
     this.router.post('/oauth2/unauthorize', commonLimiter, validate(clientManageSchema), (req, res) => this.unAuthorizeClient(req, res));
-    this.router.post('/oauth2/token', commonLimiter, (req, res) => this.oauthExpressServer.token({})(req, res, () => {}));
+    this.router.post('/oauth2/token', commonLimiter, (req, res) => this.oauthExpressServer.token({})(req, res, () => {
+      console.log(1);
```
Aivean: accidental?

pazoozoo42: yeah, will remove, thanks!

backend/src/db/repositories/OAuth2Repository.ts
```diff
@@ -0,0 +1,213 @@
...
+export default class OAuth2Repository {
+  private db: DB;
+
+  // Define a base SQL query that both methods share.
+  private clientBaseQuery = `
+    select 
+      oauth_clients.*,
+      oauth_consents.scope as scopes,
+      oauth_consents.last_revoked_ts as last_revoked_ts,
+      (select count(oauth_consents.user_id) from oauth_consents where client_id = oauth_clients.client_id) as installations_count
```
Aivean: this could lead to performance issue if lots of apps are rendered at the same time

frontend/src/Components/OAuth2ScopesComponent.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+    </label>
+  );
+}
+
+export default function OAuth2ScopesComponent(props: { appRequests: string | null }) {
+  const requestedScopesNames = ['openid'].concat(props.appRequests?.split(' ').map(s => s.trim()) || []);
+  return (<div className={styles.container}>
+    {
+      Object.keys(OAuth2ScopesLabels)
```
Aivean: ok, this is super dangerous

this view shows only the subset of the scopes that are hardcoded on the client

the scopes that are actually requested, and will be saved to db are not validated

backend/src/db/repositories/OAuth2Repository.ts
Aivean: fixed by caching locally

frontend/src/Components/OAuth2ScopesComponent.tsx
Aivean: fixed, scopes are filtered on both the backend and the frontend

Aivean (dismissed):
@pazoozoo42, I think it's in a tolerable state to be merged and we should merge it to unblock other PRs. We can improve what's left in a separate follow-up PRs.

Please merge it at your discretion or lmk if you want more time to add more changes.

Aivean (approved):

PR #374, opened 2024-02-19
Inline expand for internal links (posts and comments)

Aivean:
Adds the ability to expand the content of internal links inline.

### Demo:
https://github.com/spaceshelter/orbitar/assets/2865203/4bba1603-4636-405b-ae1a-150a986a3f04

### DIFF
--- a/backend/src/parser/TheParser.ts
+++ b/backend/src/parser/TheParser.ts
@@ -7,12 +7,18 @@ import qs from 'qs';
 import {mentionsRegex, urlRegex, urlRegexExact} from './regexprs';
 import {MediaHostingConfig} from '../config';
 import render from 'dom-serializer';
+import {joiSite} from '../api/ApiMiddleware';
 
 export type ParseResult = {
     text: string;
     mentions: string[];
     urls: string[];
     images: string[];
+};
+
+export type ParserConfig = {
+    mediaHosting: MediaHostingConfig;
+    siteDomain: string;
 };
 
 class ParserExtended extends Parser {
@@ -32,10 +38,12 @@ export default class TheParser {
     static readonly VERSION = 2;
 
     private readonly mediaHostingConfig: MediaHostingConfig;
+    private readonly parserConfig: ParserConfig;
     private readonly mediaHostingUrlOrigin: string;
 
-    constructor(mediaHosting: MediaHostingConfig) {
-        this.mediaHostingConfig = mediaHosting;
+    constructor(parserConfig: ParserConfig) {
+        this.parserConfig = parserConfig;
+        this.mediaHostingConfig = parserConfig.mediaHosting;
         // add origin. subdomain to the url
         this.mediaHostingUrlOrigin =
             this.mediaHostingConfig.url.replace(/^https?:\/\//, 'https://origin.');
@@ -203,7 +211,8 @@ export default class TheParser {
             this.processVimeo(pUrl) ||
             this.processImage(pUrl) ||
             this.processCoub(pUrl) ||
-            this.processVideo(pUrl);
+            this.processVideo(pUrl) ||
+            this.processInternalUrl(url);
         if (res !== false) {
             return res;
         }
@@ -226,6 +235,26 @@ export default class TheParser {
 
         return false;
     }
+
+    processInternalUrl(url: string, text?: string) {
+        const internalUrlPattern = new RegExp(`^(https?://${this.parserConfig.siteDomain})/(?:s/([^/]+)/)?p(\\d+)(?:#(\\d+))?$`, 'i');
+
+        const match = url.match(internalUrlPattern);
+        const site = match && match[2] || null;
+        if (match && (!site || !joiSite.validate(site).error)) {
+            const host = match[1];
+            const postId = match[3];
+            const commentId = match[4] || null;
+
+            const origUrl = `${host}/${site ? 's/' + site + '/' : ''}p${postId}${commentId ? '#' + commentId : ''}`;
+            const expandButton = `<span role="button" class="expand-button i i-expand" data-post-id="${postId}"${commentId ? ' data-comment-id="' + commentId + '"' : ''}></span>`;
+
+            return `${expandButton}<a href="${encodeURI(origUrl)}" target="_blank">${text || url}</a>`;
+        }
+
+        return false;
+    }
+
 
     processCoub(url: Url<string>) {
         const coubIdPattern = /^\/view\/(\w+)$/;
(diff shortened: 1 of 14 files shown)


backend/src/parser/TheParser.ts
```diff
@@ -227,6 +236,26 @@ export default class TheParser {
         return false;
     }
 
+    processInternalUrl(url: string, text?: string) {
+        const internalUrlPattern = new RegExp(`^(https?://${this.parserConfig.siteDomain})/(?:s/([^/]+)/)?p(\\d+)(?:#(\\d+))?$`, 'i');
```
4vanger:
```suggestion
        const escapeRegexp = (str) => str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
        const internalUrlPattern = new RegExp(`^(https?://${escapeRegexp (this.parserConfig.siteDomain)})/(?:s/([^/]+)/)?p(\\d+)(?:#(\\d+))?$`, 'i');
```

backend/src/parser/TheParser.ts
```diff
@@ -227,6 +236,26 @@ export default class TheParser {
...
+
+        const match = url.match(internalUrlPattern);
+        const site = match && match[2] || null;
+        if (match && (!site || !joiSite.validate(site).error)) {
+            const host = match[1];
+            const postId = match[3];
+            const commentId = match[4] || null;
+
+            const origUrl = `${host}/${site ? 's/' + site + '/' : ''}p${postId}${commentId ? '#' + commentId : ''}`;
+            const expandButton = `<span role="button" class="expand-button i i-expand" data-post-id="${postId}"${commentId ? ' data-comment-id="' + commentId + '"' : ''}></span>`;
```
4vanger, line 250: I think we shouldn't be using `expand` icon on different purpose. May be hyperlink icon?
add tooltip for better UX
```suggestion
            const expandButton = `<span role="button" class="expand-button i i-expand" title="А что там?" data-post-id="${postId}"${commentId ? ' data-comment-id="' + commentId + '"' : ''}></span>`;
```

backend/test/parser/TheParser.test.ts
```diff
@@ -2,8 +2,11 @@
 import TheParser from "../../src/parser/TheParser";
 
 const p = new TheParser({
+    mediaHosting: {
     url: 'https://orbitar.media',
-    dimsAesKey: ''
+        dimsAesKey: ''
```
4vanger, line 7: we need prettier

frontend/src/Components/InternalLinkExpandComponent.tsx
```diff
@@ -0,0 +1,58 @@
...
+        } else {
+            api.post.get(postId).then((post) => {
+                setData({type: 'post', post: post.post});
+            });
+        }
+    }, [postId, commentId]);
+
+    const render = () => {
+        if (!data) {
+            return <div>Загружаем...</div>;
```
4vanger, line 47: animated loading indicator will be better

frontend/src/index.tsx
```diff
@@ -38,16 +38,28 @@ const MobXAwareRouter: FunctionComponent = (props) => {
...
-            <ThemeProvider themeCollection={ getThemes() }>
-                <ReloadOnUpdate>
-                    <App />
-                </ReloadOnUpdate>
-            </ThemeProvider>
-        </MobXAwareRouter>
-    </AppStateProvider>
+    <FakeRoot>
+        <ReloadOnUpdate>
+        <App/>
```
4vanger: nit: indentation

frontend/src/Components/ContentComponent.tsx
```diff
@@ -197,6 +208,56 @@ function updateMail(mail: HTMLSpanElement,
     });
 }
 
+function updateInternalExpandButton(expandButton: HTMLElement, appState: AppState) {
```
4vanger, line 211: I don't like this approach at all - it works but stinks horribly.

frontend/src/index.tsx
```diff
@@ -38,16 +38,28 @@ const MobXAwareRouter: FunctionComponent = (props) => {
     );
 };
 
+export const FakeRoot: React.FunctionComponent<{
+    children: React.ReactNode;
+    appState?: AppState;
+}> = ({appState, children}) => {
+    const inner = () => <MobXAwareRouter>
```
4vanger: No need for a function
```suggestion
    const inner = <MobXAwareRouter>
```

frontend/src/Components/PostComponent.module.scss
```diff
@@ -6,6 +6,10 @@
   margin-bottom: 32px;
 }
 
+:global(.internalLinkExpand) .post {
+  margin-bottom: 12px;
```
4vanger, line 10: Let's add a border to separate expand content from the rest of the post/comment - currently it is hard to tell which is which

4vanger (dismissed):
Approving but it stinks a bit - don't like how we are mixing react and DOM manipulations.
Also it is obvious that we need an app level - will try adding that in upcoming weeks

Aivean:
> Approving but it stinks a bit - don't like how we are mixing react and DOM manipulations. Also it is obvious that we need an app level - will try adding that in upcoming weeks

Wdym by app level?

backend/src/parser/TheParser.ts
Aivean: I kinda agree that slightly different icon would be better, but it shouldn't be completely different, as both actions share the similar "expand" part.

@ArtCraft, do you think you can come up with a better icon, or is the current one fine?
image

frontend/src/Components/InternalLinkExpandComponent.tsx
Aivean: It would, but not essential for the first version.

frontend/src/Components/ContentComponent.tsx
Aivean: Are there any concerns with it beside aesthetics? 

Reactifying all content (like here: https://github.com/spaceshelter/orbitar/pull/77) could be a performance problem, but I guess we could do this selectively.

frontend/src/Components/PostComponent.module.scss
Aivean: But there is a border already (on the left side).

Aivean:
Thanks for the review!

frontend/src/Components/PostComponent.module.scss
4vanger: Imho it is very subtle and hardly visible - it mixes up with comment layers.

4vanger:
> > Approving but it stinks a bit - don't like how we are mixing react and DOM manipulations. Also it is obvious that we need an app level - will try adding that in upcoming weeks
> 
> Wdym by app level?



> > Approving but it stinks a bit - don't like how we are mixing react and DOM manipulations. Also it is obvious that we need an app level - will try adding that in upcoming weeks
> 
> Wdym by app level?

Sorry, I meant `app level store`

Aivean:
> > Wdym by app level?
> 
> Sorry, I meant `app level store`

We already have it:
https://github.com/spaceshelter/orbitar/blob/3e815a8baa6f71dd44039c4096b28cea0a2068a5/frontend/src/AppState/AppState.tsx#L43

frontend/src/Components/PostComponent.module.scss
Aivean: Made the border brighter:
image

Aivean:
@4vanger, can you stamp again please?

4vanger (approved):

PR #375, opened 2024-02-23
Changeable indent for comments

sly2m:
Add configuration for changeable comments branches indent. 
or1

### DIFF
--- a/frontend/src/Components/CommentComponent.module.scss
+++ b/frontend/src/Components/CommentComponent.module.scss
@@ -123,8 +123,9 @@
     }
 }
 
-$bigShift: 12px;
-$smallShift: 8px;
+$indentScale: var(--indent-scale);
+//$bigShift: calc(6px * $indentScale);
+$smallShift: calc(4px * $indentScale);
 
 .answers {
     margin-top: 8px;
@@ -134,7 +135,7 @@ $smallShift: 8px;
 
     padding-left: $smallShift;
     @media only screen and (min-width: 990px) {
-        padding-left: $bigShift;
+        //padding-left: $bigShift;
         margin-left: 8px; // half of [-] icon width
     }
 }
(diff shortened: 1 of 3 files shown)


frontend/src/Components/UserProfileSettings.tsx
```diff
@@ -67,6 +75,7 @@ export default function UserProfileSettings(props: UserProfileSettingsProps) {
 const [autoStop, setAutoStop] = React.useState<boolean>(getVideoAutopause());
 const [legacyZoom, setLegacyZoom] = React.useState<boolean>(getLegacyZoom());
 const [preferredLang, setPreferredLang] = React.useState<string>(getPreferredLang());
+const [indentScale, setIndentScale] = React.useState<string>(getIndentScale());
```
Aivean, line 78: Could be replaced with `useLocalStorage` hook. See example here:
https://github.com/spaceshelter/orbitar/blob/02f60bfdea89b6d5aa19db6886bcf2569f1fa7b9/frontend/src/Components/Topbar.tsx#L74

frontend/src/Components/CommentComponent.module.scss
```diff
@@ -127,8 +127,9 @@
     }
 }
 
-$bigShift: 12px;
-$smallShift: 8px;
+$indentScale: var(--indent-scale);
+//$bigShift: calc(6px * $indentScale);
```
Aivean: Please don't leave dead code.

Aivean (commented):
If you're adamant on having it as a setting, please add an "Auto" option, and make it a default.

However, I personally don't think that having it as a setting is warranted. Are your personal settings that different from reasonable default?

frontend/src/Components/UserProfileSettings.tsx
sly2m: useState, localStore and useEffect hook was implemented inline with other functions of the component - getPreferredLang(), getLegacyZoom() etc. to maintain code uniformity throughout the page.

frontend/src/Components/CommentComponent.module.scss
sly2m: Noted.

sly2m:
Added Auto option for preference settings.

Aivean:
Sorry for the delay.

I don't like the change in the default behavior:
image

The default value is different, and doesn't shrink automatically on mobile.

PR #377, opened 2024-04-10
Introduce a tooltip to the activity icon providing details regarding the user's most recent interaction. 

MrKalach:
Introduce a tooltip to the activity icon providing details regarding the user's most recent interaction. Time resolution is set to days

Couple screenshots:
image
image

### DIFF
--- a/backend/src/managers/UserManager.ts
+++ b/backend/src/managers/UserManager.ts
@@ -18,6 +18,7 @@ import {FeedSorting} from '../api/types/entities/common';
 import TheParser from '../parser/TheParser';
 import {UserCache} from './UserCache';
 import {aesDecryptFromBase64, aesEncryptToBase64} from '../parser/CryptoUtils';
+import moment from 'moment';
 
 const USER_RESTRICTIONS = {
     MIN_KARMA: -1000,
@@ -696,4 +697,14 @@ export default class UserManager {
     async getPublicKey(userId: number) {
         return this.userCache.getPublicKey(userId);
     }
+
+    async getUserVisitedDaysAgo(userId: number): Promise<number> {
+        const lastVisited = await this.userRepository.getUserLastVisited(userId);
+
+        const daysAgo =  lastVisited ?
+            moment().diff(lastVisited, "days") : 0;
+
+        return daysAgo;
+    }
+
 }
--- a/backend/src/managers/UserManager.ts
+++ b/backend/src/managers/UserManager.ts
@@ -18,7 +18,6 @@ import {FeedSorting} from '../api/types/entities/common';
 import TheParser from '../parser/TheParser';
 import {UserCache} from './UserCache';
 import {aesDecryptFromBase64, aesEncryptToBase64} from '../parser/CryptoUtils';
-import moment from 'moment';
 
 const USER_RESTRICTIONS = {
     MIN_KARMA: -1000,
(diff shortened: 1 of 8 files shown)


backend/src/managers/UserManager.ts
```diff
@@ -696,4 +697,14 @@ export default class UserManager {
     async getPublicKey(userId: number) {
         return this.userCache.getPublicKey(userId);
     }
+
+    async getUserVisitedDaysAgo(userId: number): Promise<number> {
+        const lastVisited = await this.userRepository.getUserLastVisited(userId);
+
+        const daysAgo =  lastVisited ?
+            moment().diff(lastVisited, "days") : 0;
```
Aivean: lint

frontend/src/Pages/UserPage.module.scss
```diff
@@ -28,8 +28,9 @@
     }
 }
 .active {
+  position: relative;
   padding-right: 20px;
-  pointer-events: none;
+  //pointer-events: none;
```
Aivean: Please remove dead code.

backend/src/managers/UserManager.ts
```diff
@@ -18,6 +18,7 @@ import {FeedSorting} from '../api/types/entities/common';
 import TheParser from '../parser/TheParser';
 import {UserCache} from './UserCache';
 import {aesDecryptFromBase64, aesEncryptToBase64} from '../parser/CryptoUtils';
+import moment from 'moment';
```
Aivean: Moment.js is a transitive dependency. Either add it explicitly as a dependency into `package.json`, or, better yet, just implement it as a formula, like `Math.abs(date2 - date1) / (3600 * 24 * 1000)`.

frontend/src/Pages/UserPage.module.scss
```diff
@@ -84,3 +85,42 @@
         margin-right: 4px;
     }
 }
+
+.active .tooltipText {
```
Aivean, line 88: Tooltip looks nice, but please consolidate the color and border thickness with other menus, like:
image

https://github.com/spaceshelter/orbitar/blob/3c8d12aae420a65adde04dd0f2c22732439e5ebf/frontend/src/Components/PostComponent.module.scss#L116

frontend/src/Pages/UserPage.tsx
```diff
@@ -88,6 +88,13 @@ export const UserPage = observer(() => {
             refreshProfile();
         };
 
+        const daysAgoToStr = (daysAgo: number) => {
+            const titles = ['день', 'дня', 'дней'];
+            const cases = [2, 0, 1, 1, 1, 2];
+            return daysAgo === 0 ? 'сегодня'
+                : `${daysAgo} ${titles[daysAgo % 100 > 4 && daysAgo % 100 < 20 ? 2 : cases[daysAgo % 10 < 5 ? daysAgo % 10 : 5]]} назад`;
```
Aivean: image

Two notes:
1. Multiline text breaks the layout of the tooltip
2. Precise number of days makes little sense for anything over ≈5. Consider using 
`moment.duration.humanize`:

```
`${moment.duration(daysAgo, 'days').humanize(true)} назад`
```
with different threshold for days, weeks, months, years

backend/src/managers/UserManager.ts
Aivean: returning `0` for users who never visited the site could be confusing. Maybe worth the effort to return `undefined`/`null` in this case and handle it in UI.

Aivean (changes requested):
Overall looks good, but there are some minor notes.

frontend/src/Pages/UserPage.module.scss
MrKalach: Fixed 
image

backend/src/managers/UserManager.ts
MrKalach: done

frontend/src/Pages/UserPage.module.scss
MrKalach: done

backend/src/managers/UserManager.ts
MrKalach: Fixed, calculating days directly, without the lib

frontend/src/Pages/UserPage.tsx
MrKalach: both 1 & 2 fixed
image

backend/src/managers/UserManager.ts
MrKalach: Fixed, now we do not show any tooltip, if there are no user records in activity db.

Aivean (approved):
LGTM

PR #378, opened 2024-04-27
fix unclickable links in toolbar

maximt:
Quick fix for Vivaldi browser.


Aivean (approved):
Tested locally, it works, thanks!

PR #382, opened 2024-08-13
usage updates

4vanger:
update model to a cheaper but higher-quality chatgpt-4o-mini
significantly increase limits


Aivean (approved):

Aivean:
I wouldn't call gpt-4o-mini a "higher quality" model, but probably good enough for those usecases.

4vanger:
wouldn't argue with you on that ;)

PR #383, opened 2024-08-14
Option to edit site info for site owners

pazoozoo42:
rejected

### DIFF
--- a/backend/migrations/20240811000000-site-info.js
+++ b/backend/migrations/20240811000000-site-info.js
@@ -1,0 +1,29 @@
+'use strict';
+
+var dbm;
+var type;
+var seed;
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function(options, seedLink) {
+  dbm = options.dbmigrate;
+  type = dbm.dataType;
+  seed = seedLink;
+};
+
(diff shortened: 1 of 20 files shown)


Aivean:
wdym "rejected"?

pazoozoo42:
> wdym "rejected"?

Following our discussion on Discord, I’ve realized that the proposed solution may not align with the vision of most of the other developers. As a result, I’ve decided not to invest time and effort in exploring alternative solutions

Aivean:
> > wdym "rejected"?
> 
> Following our discussion on Discord, I’ve realized that the proposed solution may not align with the vision of most of the other developers. As a result, I’ve decided not to invest time and effort in exploring alternative solutions

I proposed a relatively minor change that brings it back into alignment: instead of creating a single "owner" of the subsite, give the edit ability to the top N current contributors to the subsite content. Do you see any problems with this suggestion? Do you want me (or somebody else) try to implement it?

pazoozoo42:
I’m personally not a big fan of the idea of having more than one owner and 'dynamic' ownership in general. This approach could lead to unpredictable consequences, edit wars similar to Barmalini posts and may potentially result in even more sraches than we already experience, and we already have quite a few

Aivean:
> edit wars similar to Barmalini posts and may potentially result in even more sraches than we already experience, and we already have quite a few

sounds beneficial, no?

PR #384, opened 2024-08-15
find-replace docker-compose to docker compose

Lexy2:
Update workflows in response to github agent changes


Lexy2:
Ready for CR

Aivean (approved):
Looks good, thanks! 

we also need to do the same thing for media

PR #388, opened 2024-08-23
Fix a bug when expanding Expand will trigger auto cut

4vanger:
Long standing bug - expanding an <expand> section will trigger autocut

Before:
cut-before

After:

https://github.com/user-attachments/assets/7958c255-6607-449a-b736-ca4ce40a7ad0


Aivean (approved):

Aivean:
thanks!

ideally this needs to be applied to link expansion as well

PR #390, opened 2024-08-23
display translate button outside of the menu if detected content lang…

4vanger:
…uage is different from user's preferred lang
per [discussion](https://orbitar.space/p24564#911588)
preferred lang: `ua`

preferred lang: `ru`


Aivean (dismissed):
LGTM code-wise, but there might be an issue with preferred languages other than the most frequent one — essentially all comments will have the button displayed inline, which arguably is not the best UX. Maybe add a setting how to display the button?

Aivean (approved):

PR #394, opened 2024-08-25
revert "do not render controls for no inview comments"

4vanger:
Too many unexpected bugs and apparently noone else was having same problem


Aivean:
Perhaps we should fix scrolling once and for all instead?

I think selective rendering is a useful feature and could evolve in the future.

Aivean (approved):

PR #395, opened 2024-08-29
Split compose & caddy files

Lexy2:
## Problem statement

We've got some configuration drift by virtue of having almost identical configurations in different `docker-compose` files and different Caddy configurations.

## Solution

This PR attempts to mitigate those issues by making hierarchical structure of the Docker & Caddy configuration files used.

### Development

The main configuration file is now `compose-dev.yml`. Local development requires use of `docker compose -f ./compose-dev.yml ...`

### Deployment

Main deployment requires plugging in 1 compose file: `compose.yml`

### Configuration

There is now one main `Caddyfile` that serves it all.

Its configurability is managed by environment variables ~~and imported snippet files.~~
~~- `*.tls.Caddyfile`~~

~~Default `off.tls.Caddyfile` can be copied as is.~~
~~Override file needs to have its dash replaced with a dot (to `on.tls.Caddyfile` to provide updated behaviour. They are also mutually exclusive with the default files; only one of each should be copied.~~

The statement above is no longer actual.

### Subdomains

We don't use or support subdomains in our prod, so I dropped its support from caddy.
`www.orbitar.space` / `local` is now redirected to `orbitar.space` / `local`.

### CORS

~~I finally configured CORS the way it *should* work.~~

~~Only backend requests are responded with CORS policy allowing cross-domain requests from the frontend and frontend only. Even in development. You want access to backend directly - you always have your `*:5001` server running.~~

The statement above is no longer actual. For external OAuth apps, `Access-Control-Allow-Origin` header should stay `*`.

The backend CORS-related headers are removed by caddy - this backend feature caught me by surprise and I puzzled for a little bit _(I configured it right, why is it not working?)_

## Environment

Most of the changes in the environment variables were for Caddy. 

- `API_DOMAIN`. This variable was always defined by never used. Instead, we used `api.$SERVER_DOMAIN`. I don't know why. Now API / backend host is defined exclusively by this variable.
- `BACKEND_DOMAIN`. Set to either `host.docker.internal` or `backend` respectively in the docker compose files.
- `TLS_CERT_FILE` & `TLS_KEY_FILE`. Default to `orbitar.crt` & `orbitar.key` respectively. Overridden by compose files.
- `SERVER_HOST`. Set to `http://$SERVER_DOMAIN` or `https://$SERVER_DOMAIN`, depending on the configuration.
- `API_HOST`. Set to `http://$API_DOMAIN` or `https://$API_DOMAIN`, depending on the configuration.
- `DEV`. If set to `true`, Caddy will reverse-proxy request to the Docker host (for local backend & frontend development), otherwise - to the relevant containers.

## Other

Updated `uses` actions to their latest versions.


backend/package.json
```diff
@@ -24,9 +24,7 @@
     "@types/express": "^4.17.13",
     "@types/html-escaper": "^3.0.0",
     "@types/jest": "^27.5.1",
-    "@types/joi": "^17.2.3",
```
Aivean: are these changes relevant / intentional? 

let's move them to a separate PR

README.md
```diff
@@ -7,33 +7,34 @@
...
 ```
-docker compose -p orbitar -f docker-compose.local.yml build --no-cache frontend backend
-```
+
 Запуск:
-```
-docker compose -p orbitar -f docker-compose.local.yml up
+
+```sh
+docker compose -p orbitar -f docker-compose.yml -f docker-compose.deploy.yml -f docker-compose.deploy.nonssl.yml up
```
Aivean: can we use `include` to avoid specifying multiple files here?

https://docs.docker.com/compose/how-tos/multiple-compose-files/include/

caddy/Caddyfile.dev
```diff
@@ -1,32 +0,0 @@
-http://{$SERVER_DOMAIN} {
```
Aivean, line 1: It may be possible to use Caddy snippets to further deduplicate Caddy files:
https://gist.github.com/Aivean/05e94e4d5b6bb910f86f76bae354445f

README.md
Lexy2: I'll take a look. IIRC, includes can't conflict.

caddy/Caddyfile.dev
Lexy2: I'll take a look. IIRC, snippets are static, can't be parameterized, and that was an issue for further deduplication.

Aivean: they can be conditionally enabled based on env variable:
```
    # Optional TLS configuration based on environment
    @has_ssl {
        environment SSL true
    }
    tls @has_ssl {$SSL_CERT_FILE} {$SSL_KEY_FILE}
```

Lexy2: Is that a special plugin?

`caddy validate ./Caddyfile`:

`Error: adapting config using caddyfile: getting matcher module 'environment': module not registered: http.matchers.environment`

Lexy2: Plus, `tls` directive does not accept a matcher, so it's unconditional.

Lexy2: I also don't see too much value in creating snippets for parts that are used in one place only, like uploading.

Lexy2: Updated the dev configuration for frontend / backend, but don't see a way of optimizing TLS configuration this way. Let me know if you have any other ideas.

README.md
Lexy2: Renamed base files and made real ones consist of includes

caddy/Caddyfile
```diff
@@ -0,0 +1,72 @@
+import *.tls.Caddyfile
```
Aivean: can you please try substituting filename based on env variable?
https://chatgpt.com/share/679f24a6-2f7c-800e-84f4-e59f07ec1e84

https://caddyserver.com/docs/caddyfile/concepts?utm_source=chatgpt.com#environment-variables
> Environment variables in this form are substituted before Caddyfile parsing begins, so they can expand to empty values (i.e. ""), partial tokens, complete tokens, or even multiple tokens and lines.

Aivean: or, I guess, it doesn't have to be filename, it can be two TLS snippets

Lexy2: Yup, this should be possible. Will do.

Aivean:
I feel like an ideal (idiomatic) state would be:
* TLS enabled and configured in .env
* a single caddyfile
* a single docker-compose.yml file
* a docker compose profile to switch between dev/prod

not sure if last two items are technically feasible, despite the fact that most of the current differences between configs come from services

maybe a compromise would be to have just two docker-compose files, for dev and prod

Lexy2:
> I feel like an ideal (idiomatic) state would be:
> 
>     * TLS enabled and configured in .env
> 
>     * a single caddyfile
> 
>     * a single docker-compose.yml file
> 
>     * a docker compose profile to switch between dev/prod
> 
> 
> not sure if last two items are technically feasible, despite the fact that most of the current differences between configs come from services
> 
> maybe a compromise would be to have just two docker-compose files, for dev and prod

I think this is feasible. Would you like me to try it within this PR? I mean, `.env` TLS?

Aivean:
> I think this is feasible. Would you like me to try it within this PR? I mean, .env TLS?

I'd appreciate that, if you have time.

I also have some questions regarding CORS:

> Only backend requests are responded with CORS policy allowing cross-domain requests from the frontend and frontend only. Even in development. You want access to backend directly - you always have your *:5001 server running.

1. we're not breaking general thirdparty access to api on `api.orbitar.space:80`, are we? 

2. would CORS in this case only block requests that are made from the browser?  

3. we probably need to have a separate CORS policy for the [new oauth API](https://github.com/spaceshelter/orbitar/pull/371) that allows cross-domain access for requests that have `authorization` header.

_compose.production.tls_on.yml
```diff
@@ -0,0 +1,13 @@
...
+      - "443:443"
+      - "443:443/udp"
+    environment:
+      SERVER_HOST: https://${SERVER_DOMAIN}
+      API_HOST: https://${API_DOMAIN}
+      TLS_CERT_PATH: /certs/${TLS_CERT_FILE}
+      TLS_KEY_PATH: /certs/${TLS_KEY_FILE}
+      TLS_MODE: tls_on
+    volumes:
+      - $PWD/caddy/certs:/certs:ro
```
Aivean: I think `tls_on` and `tls_off` docker compose configs could be unified.

something like this:
```
services:
  caddy:
    ports:
      # Standard HTTP port
      - "80:80"

      # Additional HTTPS ports when TLS is enabled
      - "${TLS_ENABLED:+443}:${TLS_ENABLED:+443}"
      - "${TLS_ENABLED:+443}:${TLS_ENABLED:+443}/udp"

    environment:
      SERVER_HOST: http${TLS_ENABLED:+s}://${SERVER_DOMAIN}
      API_HOST: http${TLS_ENABLED:+s}://${API_DOMAIN}
      TLS_CERT_PATH: ${TLS_ENABLED:+/certs/${TLS_CERT_FILE}}
      TLS_KEY_PATH: ${TLS_ENABLED:+/certs/${TLS_KEY_FILE}}
      TLS_MODE: ${TLS_ENABLED:+tls_on}

    volumes:
      - ${TLS_ENABLED:+$PWD/caddy/certs:/certs:ro}
```

relies on the interpolation syntax:
https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#interpolation-syntax

just need to make sure compose handles empty values for volumes and ports gracefully (it should)

Aivean: if this works, let's just have TWO files with dev and prod configs, e.g. `compose.yml` and `compose-dev.yml` (without base files and includes), I believe having just two files is the sweet spot between readability and maintainability, while current setup requires significant cognitive effort to parse and understand where to make changes.

Lexy2:
> 
>     1. we're not breaking general thirdparty access to api on `api.orbitar.space:80`, are we?
> 
>     2. would CORS in this case only block requests that are made from the browser?
> 
CORS is used by browsers that would block 3rd-party HTTP / API calls based on CORS headers. If you're using an API client that does not respect CORS, it will work. However, if any API clients are browser-based (such as PowerShell's `Invoke-RestMethod`), they could be blocked.

>     3. we probably need to have a separate CORS policy for the [new oauth API](https://github.com/spaceshelter/orbitar/pull/371) that allows cross-domain access for requests that have `authorization` header.
I'll take a look

_compose.production.tls_on.yml
Lexy2: > I believe having just two files is the sweet spot between readability and maintainability, while current setup requires significant cognitive effort to parse and understand where to make changes.

Well, readability is at question with this approach :-)

How about 3 files? Base, dev & prod?

Lexy2:
>     3. we probably need to have a separate CORS policy for the [new oauth API](https://github.com/spaceshelter/orbitar/pull/371) that allows cross-domain access for requests that have `authorization` header.

I'll have to test this PR separately to see whether we can limit CORS extension for OAuth apps.

Aivean:
> > ```
> > 3. we probably need to have a separate CORS policy for the [new oauth API](https://github.com/spaceshelter/orbitar/pull/371) that allows cross-domain access for requests that have `authorization` header.
> > ```
> 
> I'll have to test this PR separately to see whether we can limit CORS extension for OAuth apps.

Basically, once we have oauth, I'd like to preserve the ability of static (html+js) third-party apps to at least login with orbitar's oauth.

_compose.production.tls_on.yml
Aivean: my point is that when you have all relevant configuration in the plain view (in a single file) it's more readable than when you have to switch between two files and merge it in your head. also, consider that for some changes, you'd need to modify all three files, and do additional thinking on how to split the change across the files.

anyway, I'd defer to your judgement, just keep in mind that deduplication comes at a cost

Aivean: and either approach is significant improvement over what we have currently

Lexy2:
> Basically, once we have oauth, I'd like to preserve the ability of static (html+js) third-party apps to at least login with orbitar's oauth.

Of course, but in this case we could relax CORS for specific endpoints.

Lexy2:
# TODO:

## CORS
### Problem statement
1. CORS helps if a 3rd-party site is making requests to backend. If the user is authenticated, browser will add cookie to the request, and this site can scrape the information from orbitar silently.
2. CORS may block OAuth apps because it will now allow access to backend only from frontend, and OAuth app will need to call the backend API with the bearer token with relevant permissions.

### Next actions:
* we still need CORS
* in order not to block OAuth apps, we can enforce CORS requirement only if a session cookie is present in the request.

## Default docker-compose configuration
### Problem statement
You should be able to spin up the project with minimal effort.
1. Minimal effort spin-up should deploy production to try orbitar.
2. Minimal effort spin-up should deploy development to facilitate regular developers.
3. Deployment by hand in production environment should be safe to execute.
4. Deployment by hand in production environment is never safe and should require special setup.

### Next actions:
* Ask contributors how they use orbitar in their development environment
* Based on their preference, choose the minimal effort configuration:
   * `docker compose up` without `-f` for dev (if possible, also see below)
   * `docker compose up` without `-f` for prod (if possible, also see below)

## Best practice in dockerfile development / duplication
### Problem statement:
1. We have different services spun up for development environment and production environment. 
2. We also have different configuration for services spun up in both environments.
3. Based on this, we can't just get away with compose profiles. And it also doesn't make too much sense to use both profiles and something else to reconfigure services.
4. However, now we have many files copying the same configuration.

### Next actions:
* Investigate if it's possible to create fewer files with overrides.
* We can either (accounting for dev preferences above):
   * Set up 3 docker files: base / dev override / prod override.
   * Set up 2 docker files: base aka prod / dev override
* Override files should try to minimize duplication of the config, and not be copies of the base config.
* TLS configuration should stay configurable via `.env` files and not require additional override files.

Aivean:
By default cross-origin requests don't include cookies, to allow that  `Access-Control-Allow-Credentials: true` must be set.

Aivean:
trying to implement what we discussed:
https://gist.github.com/Aivean/333114cf4f25367489f26a19bb7decc2

also noticed, that websocket connection to frontend is broken in dev mode

Lexy2:
> trying to implement what we discussed: https://gist.github.com/Aivean/333114cf4f25367489f26a19bb7decc2
> 
> also noticed, that websocket connection to frontend is broken in dev mode

At this stage I think it's better to **not** set `Access-Control-Allow-Credentials` and keep `Access-Control-Allow-Origin: *`

Lexy2:
> also noticed, that websocket connection to frontend is broken in dev mode

Always has been

Set default webpack dev server port to 80, should work at least in non-TLS mode now.

Lexy2:
Also sneaked in another prettier change. Turned out, removal of `endOfLine: "lf"` didn't impact anything as this is the default behaviour. So it has to be set to `"auto"` to stop bothering me on Windows.

Aivean:
> Also sneaked in another prettier change. Turned out, removal of `endOfLine: "lf"` didn't impact anything as this is the default behaviour. So it has to be set to `"auto"` to stop bothering me on Windows.

a better practice is to submit unrelated changes as a separate PR, and merge /rebase them in, once they are approved and merged into the main branch

Aivean:
I don't want to merge this PR before the planned release this weekend, need to test it thoroughly.

Lexy2:
> > Also sneaked in another prettier change. Turned out, removal of `endOfLine: "lf"` didn't impact anything as this is the default behaviour. So it has to be set to `"auto"` to stop bothering me on Windows.
> 
> a better practice is to submit unrelated changes as a separate PR, and merge /rebase them in, once they are approved and merged into the main branch

No worries :-)

Lexy2:
@Aivean waiting on your review here.

Aivean:
> @Aivean waiting on your review here.

Sorry. I'll try to find some time to finally get this in, but now sure how soon :(

PR #396, opened 2024-10-07
Parse mobile Youtube URls when pasting video links

cajax:
Handle `m.youtube.com/watch...` URLs in post parsing.


pazoozoo42 (approved):
@cajax thank you for the improvement, very nice! Sorry for the late review, to speed things up you can add reviewers explicitly and they will get notifications

PR #397, opened 2025-01-31
Update artifact action

Lexy2:


Aivean (approved):
awesome, thanks!

let's test in prod

PR #398, opened 2025-02-08
Route media through bunny cdn

Aivean:
Rewrites media urls in the content to point to b.orbitar.media — new bunny cdn setup:
* images
* videos and previews
* vimeo and coub previews

---

This is the first stage of the migration to bunny cdn:

1. Switch to temporary bunny domain (b.orbitar.media) for the new media
2. Monitor
3. Switch main domain (orbitar.media) to bunny
4. Reprocess all content (b.orbitar.media -> orbitar.media)

---

### Testing:

1. Tested locally, re-parsing all local content.
2. Verified that upload (/api/upload.php) also works through cdn (would be important during the final switch).

### DIFF
--- a/backend/src/parser/TheParser.ts
+++ b/backend/src/parser/TheParser.ts
@@ -40,6 +40,7 @@ export default class TheParser {
     private readonly mediaHostingConfig: MediaHostingConfig;
     private readonly parserConfig: ParserConfig;
     private readonly mediaHostingUrlOrigin: string;
+    private readonly mediaHostingUrlBunny: string;
 
     constructor(parserConfig: ParserConfig) {
         this.parserConfig = parserConfig;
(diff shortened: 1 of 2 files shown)


pazoozoo42 (approved):
code looks ok, didn't test

minor, I would only suggest to use more generic names rather than `this.mediaHostingUrlBunny` so that if some other cdn to be used, you won't need to rename the var in all methods, so something like

```
mediaHostingUrlBunny = this.mediaHostingConfig.url.replace(/^https?:\/\//, 'https://b.');
this.mediaHostingUrlToUse = mediaHostingUrlBunny;

// .. use this.mediaHostingUrlToUse instead
```
with this approach if something changes only constructor will need to be changed. 
This is minor, so up to you

Aivean:
> code looks ok, didn't test
> 
> minor, I would only suggest to use more generic names rather than `this.mediaHostingUrlBunny` so that if some other cdn to be used, you won't need to rename the var in all methods, so something like
> 
> ```
> mediaHostingUrlBunny = this.mediaHostingConfig.url.replace(/^https?:\/\//, 'https://b.');
> this.mediaHostingUrlToUse = mediaHostingUrlBunny;
> 
> // .. use this.mediaHostingUrlToUse instead
> ```
> 
> with this approach if something changes only constructor will need to be changed. This is minor, so up to you

this code is not meant to be permanent, it's needed only for the duration of migration

PR #401, opened 2025-02-13
Edit post from profile

LazyKarlson:
Edit button added to posts in username profile


Aivean (approved):
lgtm, works locally

we'll merge it after oauth2

PR #392, opened 2024-08-23
Adopt prettier

4vanger:
It is time to adopt prettier and comb the codebase - it is way too messy atm - inconsistent indents, inconsistent style, need to insert semicolons manually.
Prettier is a solution.
Plan:
- let's discuss the config on Discord
- I've applied prettier to 3 files to show how result will look like
- after we agree on config - I'll add a single commit to apply prettier to all the codebase

Here is documentation for Prettier config - https://prettier.io/docs/en/options.html

PR #403, opened 2025-02-19
Adopt Prettier - Minimal version

Aivean:
This is the stripped-down version of #392
(with code changes removed)

### Plan

- [x] Rebase this minimal version onto the `dev` once all giant prs, like oauth2 are merged
- [x] Create a [workflow](https://github.com/spaceshelter/orbitar/actions/workflows/prettier-format.yml) that takes a branch, reformats it and pushes it as a new branch ([prettier-format](https://github.com/spaceshelter/orbitar/tree/prettier-format) )
- [x] Run reformatting workflow on the branch of this PR (to generate reformatting commit)
- [x] then merge this reformatting commit back to this PR
- [x] Manually fix the rest of the lint issues
- [x] Merge this PR into `dev`

### Testing

workflow works remotely:
https://github.com/spaceshelter/orbitar/actions/workflows/prettier-format.yml

it creates a reformatting commit in the [prettier-format](https://github.com/spaceshelter/orbitar/tree/prettier-format) branch, which could be trusted (autogenerated, hash verifiable by logs).

### How to verify the codebase reformatting

Since it's virtually impossible to visually inspect reformatting of the whole codebase to ensure that nothing was sneaked in, here're the instructions how to make it formally:

1. Commits of this PR:
https://github.com/spaceshelter/orbitar/pull/403/commits
image

2. [Reformatting action run](https://github.com/spaceshelter/orbitar/actions/runs/13418664975/job/37485745968)

Checked out revision: 976570a0ddf1dc4c13bc2e338b2fc50cb1c153ae  (corresponds to the second commit above)
image

Pushed commit after automatic reformatting:
image

The same commit was merged back into this PR:
image

-----

> let's discuss the config on Discord
> I've applied prettier to 3 files to show how result will look like
after we agree on config - I'll add a single commit to apply prettier to all the codebase
> Here is documentation for Prettier config - https://prettier.io/docs/en/options.html


.github/workflows/prettier-format.yml
```diff
@@ -1,9 +1,54 @@
...
+
 on:
-  workflow_dispatch:
+  workflow_dispatch:  # Allows you to trigger this workflow manually
+permissions:
+  contents: write
+
 jobs:
-  build:
+  reformat:
```
4vanger, line 9: this is excessive - eslint check will fail if there are changes coming from prettier. 
Also it is kinda scary to commit automatically generated changes without review - in theory it is safe but sometimes may be undesirable - in rare cases you might want to disable prettier for a code block to have better formatting through indentation by example

Aivean: this is not the intention. I should clarify the plan.

the sole purpose of this job is to provide a verifiable commit that reformats the whole codebase. It's easy to verify by reviewers using e.g. `git patch-id` .

the idea is to use merge this commit into this (or a separate PR), and do the rest of minor fixed manually on top

4vanger: after it is merged `eslint` target will be failing for every commit having unformatted code. It may be easier to wait until there is no active work and launch prettier and commit result once for whole codebase

4vanger: we might disable eslint rule for now

Aivean: >  It may be easier to wait until there is no active work and launch prettier and commit result once for whole codebase

Already done, this branch is rebased on the latest dev.

4vanger (approved):
lgtm

PR #406, opened 2025-02-20
bookmarks implementation

LazyKarlson:
Bookmarks.
But problem with routing. user/bookmarks open user/posts

### DIFF
--- a/frontend/src/API/UserAPIHelper.ts
+++ b/frontend/src/API/UserAPIHelper.ts
@@ -12,6 +12,7 @@ export type UserProfileResult = {
   trialApprovers?: VoteListItemEntity[]
   numberOfPosts: number
   numberOfComments: number
+  numberOfBookmarks: string
   numberOfInvitesAvailable?: number
   isBarmalini?: boolean
   publicKey: string
--- a/frontend/src/Components/UserProfileBookmarks.tsx
+++ b/frontend/src/Components/UserProfileBookmarks.tsx
@@ -1,0 +1,116 @@
+import React, { useEffect, useRef, useState } from 'react'
+import { useLocation, useSearchParams } from 'react-router-dom'
+import { useDebouncedCallback } from 'use-debounce'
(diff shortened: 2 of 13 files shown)


frontend/src/Components/UserProfileBookmarks.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const newFilterValue = newSearchParams.get('filter') as string
+    setFilter(newFilterValue)
+    if (filterInputRef.current) {
+      filterInputRef.current.value = newFilterValue
+    }
+  }, [search])
+
+  const { posts, loading, pages, error, updatePost } = useFeed(
+    props.username,
+    'user-profile',
```
Aivean: needs correct feed here (and implemented inside `useFeed`)

Aivean:
please run prettier

Aivean:
https://github.com/user-attachments/assets/c9e02086-42ac-4325-9d05-034cc09b6113

on bookmarks page the posts are not updated when "bookmark" toggle is clicked

frontend/src/API/UserAPIHelper.ts
```diff
@@ -12,6 +12,7 @@ export type UserProfileResult = {
   trialApprovers?: VoteListItemEntity[]
   numberOfPosts: number
   numberOfComments: number
+  numberOfBookmarks: string
```
Aivean, line 15: why on earth is this a string?

Aivean:
@LazyKarlson, regarding your question on how to refresh the number of bookmarks, I think it actually would be easier to just call `refreshProfile` from here:

https://github.com/spaceshelter/orbitar/blob/cc77210ffeab78d71b5daaf4ef551280b3a4e728/frontend/src/Pages/UserPage.tsx#L33

pass it down to `UserProfileBookmarks` as an optional property

Aivean:
let's pause this PR for now, I think the new token mechanics that we are discussing might effectively provide a better bookmarks

PR #408, opened 2025-02-21
Telegram post embed

LazyKarlson:
With expand


backend/src/parser/TheParser.ts
```diff
@@ -223,6 +224,24 @@ export default class TheParser {
     return `<a href="${encodeURI(decodeURI(url))}" target="_blank">${htmlEscape(decodeURI(url))}</a>`
   }
 
+  processTelegram(url: Url<string>) {
+    if (url.host !== 't.me' && url.host !== 'telegram.me') {
+      return false
+    }
+
+    const match = url.pathname.match(/^\/([^/]+)\/(\d+)/)
+    if (!match) {
```
4vanger: nit: It is always better to be more explicit
```suggestion
    if (match === null) {
```

backend/src/parser/TheParser.ts
```diff
@@ -223,6 +224,24 @@ export default class TheParser {
...
+      return false
+    }
+
+    const [, channelName, postId] = match
+    const telegramUrl = `https://t.me/${channelName}/${postId}`
```
4vanger, line 282: Why are we reconstructing URL here? Few potential problems:
1) we are omitting GET parameters that may be important
2) depending on host embed may behave differently (unless you can prove that it is behaving the same)
Imho we should just use `url.href` here

frontend/src/Components/ContentComponent.tsx
```diff
@@ -104,6 +106,66 @@ function updateContent(
...
+  // Create expand button
+  const expandButton = document.createElement('span')
+  expandButton.className = 'expand-button i i-expand'
+  expandButton.setAttribute('role', 'button')
+  telegramDiv.insertBefore(expandButton, link)
+
+  // Add click event listener to both expand button and link
+  const listener = (e: Event) => {
+    // Allow Ctrl+click to follow the link
+    if ((e as MouseEvent).ctrlKey) {
```
4vanger: should also support `metaKey` for apple users

frontend/src/Components/ContentComponent.tsx
```diff
@@ -104,6 +106,66 @@ function updateContent(
   div.querySelectorAll('div.oauth-app').forEach((appEl) => {
     updateOauthAppEmbed(appEl as HTMLDivElement, appState)
   })
+  div.querySelectorAll('div[data-telegram-url]').forEach((telegramDiv) => {
+    updateTelegramEmbed(telegramDiv as HTMLElement, appState)
+  })
+}
+
+function updateTelegramEmbed(telegramDiv: HTMLElement, appState: AppState) {
```
4vanger: It is very much existing `updateInternalExpandButton`. Could you abstract shared functionality into separate function for more clear code please

4vanger (dismissed):

backend/src/parser/TheParser.ts
Aivean: 1. we're matching tg urls that we know we can embed. those come without GET params
2. seems like a stretched concern. I think having a dedicated format for the parameters that our embed  component expects on the fronted is a better practice

backend/src/parser/TheParser.ts
```diff
@@ -223,6 +224,24 @@ export default class TheParser {
...
+
+    return `<div data-telegram-url="${encodeURI(telegramUrl)}">
```
Aivean: let's change `div` to inline span, consistent with the expand for internal links

frontend/src/Components/ContentComponent.tsx
```diff
@@ -314,7 +319,15 @@ function updateInternalExpandButton(expandButton: HTMLElement, appState: AppStat
       // Add the rect after the link
       link.parentNode?.insertBefore(newRect, link.nextSibling)
 
-      const content = (
+      const content = src ? (
```
Aivean: this is not good design. Andrew asked to:
>  **abstract shared functionality** into separate function for more clear code please

but instead you just added more logic to the existing method.

The correct implementation should look like this:
1. shared method, like `processExpandLInk` that takes an html element and function that returns react component
2. two separate methods for processing internal links and tg embeds, that use the shared method from (1)

LazyKarlson: Done

frontend/src/Components/ContentComponent.tsx
```diff
@@ -337,6 +331,66 @@ function updateInternalExpandButton(expandButton: HTMLElement, appState: AppStat
   }
 }
 
+function processTelegramEmbed(expandButton: HTMLElement, appState: AppState, cleanupRegistry: CleanupRegistry): void {
+  const src = expandButton.getAttribute('data-telegram-url')
+  if (!src) return
+
+  const nextLink = expandButton.nextElementSibling
```
Aivean: `nextLink` can be calculated inside the `processExpandLink`

Aivean (approved):

PR #402, opened 2025-02-13
Preformatted text in comments

LazyKarlson:
2025-02-14 01 19 44 orbitar local c9d6bac2164e

PR #411, opened 2025-02-22
General mechanism that prohibits nesting of certain tags

Aivean:
Previously there was an ad-hoc functionality that removed nesting of mailboxes (in `removeInnerMailTagsRec`).

This PR adds a more general mechanism for prohibiting nesting of certain tags, and adds more rules, like `app` inside of `a`, or `pre` inside of `a`.

### Testing

manual + unit tests

### DIFF
--- a/backend/src/parser/TheParser.ts
+++ b/backend/src/parser/TheParser.ts
@@ -33,7 +33,7 @@ class ParserExtended extends Parser {
 
 export default class TheParser {
   private readonly allowedTags: Record<string, ((node: Element) => ParseResult) | boolean>
-  private disallowedTagNesting: Record<string, string | string[]>
+  private readonly disallowedTagNesting: Record<string, string | string[]>
 
   // Bump this version when introducing breaking changes to the parser.
   // Content will be re-parsed and saved on access when this version changes.
@@ -43,7 +43,7 @@ export default class TheParser {
   private readonly parserConfig: ParserConfig
   private readonly mediaHostingUrlOrigin: string
   private readonly mediaHostingUrlBunny: string
-  private blockTags: string[]
+  private readonly blockTags: string[]
 
   // Stack of tags that are currently being parsed, for internal use only
   private parseChildNodesStack: string[]
--- a/backend/src/parser/TheParser.ts
+++ b/backend/src/parser/TheParser.ts
@@ -33,6 +33,7 @@ class ParserExtended extends Parser {
 
 export default class TheParser {
   private readonly allowedTags: Record<string, ((node: Element) => ParseResult) | boolean>
+  private disallowedTagNesting: Record<string, string | string[]>
 
   // Bump this version when introducing breaking changes to the parser.
   // Content will be re-parsed and saved on access when this version changes.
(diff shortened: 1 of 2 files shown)


backend/src/parser/TheParser.ts
```diff
@@ -86,15 +106,29 @@ export default class TheParser {
...
+      return false
+    }
+    if (Array.isArray(disallowed)) {
+      for (const parent of disallowed) {
+        if (this.parseChildNodesStack.includes(parent)) return true
+      }
+      return false
+    }
+    return this.parseChildNodesStack.includes(disallowed)
+  }
```
4vanger, line 127: Maybe convert it to Sets and use [Set#intersection](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/intersection) ?

backend/src/parser/TheParser.ts
```diff
@@ -33,6 +33,7 @@ class ParserExtended extends Parser {
 
 export default class TheParser {
   private readonly allowedTags: Record<string, ((node: Element) => ParseResult) | boolean>
+  private disallowedTagNesting: Record<string, string | string[]>
```
4vanger: Let's make it 
```suggestion
  static #disallowedTagNesting = {
      pre: new Set(['a']), /* e.g. `pre` cannot be inside `a`*/,
      a: new Set(['a']),
      mailbox: new Set(['a', 'mailbox', 'mail']),
      mail: new Set(['a', 'mailbox', 'mail']),
      app: new Set(['a', 'mailbox', 'mail']),
      expand: new Set(['a', 'mailbox', 'mail']),
    }
```

backend/src/parser/TheParser.ts
```diff
@@ -42,6 +43,10 @@ export default class TheParser {
   private readonly parserConfig: ParserConfig
   private readonly mediaHostingUrlOrigin: string
   private readonly mediaHostingUrlBunny: string
+  private blockTags: string[]
```
4vanger: In fact we should do that for all properties that currently assigned static values in constructor
```suggestion
  static #blockTags = new Set(['blockquote', 'expand', 'pre'])
```

backend/src/parser/TheParser.ts
Aivean: I considered that, but there are two reasons why I chose this approach:
1. maintaining a set instead of the stack is not trivial, will require additional data structure with associated overhead
2. both arrays here are tiny (max depth of tag nesting in the typical content is rarely > 3), my intuition says the lower constant factor of the simple loops will win

backend/src/parser/TheParser.ts
Aivean: let's not 
https://www.measurethat.net/Benchmarks/Show/26731/0/sethas-vs-arrayincludes

backend/src/parser/TheParser.ts
Aivean: agreed on static, disagreed on Set

4vanger: Oki, let's convert to static and then good to go

Aivean: > Oki, let's convert to static and then good to go

I thought a bit more about it, and I don't think converting to static makes much sense:

1. `Parser` is already singleton in prod, so what are we trying to win?
2. I deliberately put initialization of all tag-related "config-like" properties in a single place:
  `this.disallowedTagNesting`, `this.blockTags`, `this.allowedTags`, so that contributors have a better chance to discover them and change, when needed. E.g. #402 missed updating `blockTags`.
3. extensibility: in the future we might want to modify behavior of a concrete parser instance based on constructor arguments

4vanger (approved):
:shipit:

PR #412, opened 2025-02-22
Add cache clearing for oauth apps cards

Aivean:
fixes minor bugs:
1. when logging out / in without page refresh the embedded app cards showed installation status for the previous user
2. make sure that embedded app cards reflect the latest changes made in settings

additionally, install / uninstall now updates the app state, and even the embedded app card:

https://github.com/user-attachments/assets/224cf0f1-bcd4-443d-a5b1-8125f61ccfda

### DIFF
--- a/frontend/src/Components/OAuth2AppCardModalComponent.tsx
+++ b/frontend/src/Components/OAuth2AppCardModalComponent.tsx
@@ -160,13 +160,12 @@ export function OAuth2AppCardModalComponent(props: OAuth2AppCardModalProps) {
       const authWindow = window.open(client.initialAuthorizationUrl, '_blank')
       if (authWindow) {
         const handleFocus = () => {
-          window.removeEventListener('focus', handleFocus)
           api.oauth2Api.clearClientCache()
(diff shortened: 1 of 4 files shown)


frontend/src/Components/OAuth2AppCardModalComponent.tsx
```diff
@@ -146,7 +157,17 @@ export function OAuth2AppCardModalComponent(props: OAuth2AppCardModalProps) {
...
+      const authWindow = window.open(client.initialAuthorizationUrl, '_blank')
+      if (authWindow) {
+        const handleFocus = () => {
+          window.removeEventListener('focus', handleFocus)
+          api.oauth2Api.clearClientCache()
+          api.oauth2Api.getClientCached(client.clientId).then((updatedClient) => {
+            props.onClientUpdate?.(updatedClient)
+          })
+        }
+        window.addEventListener('focus', handleFocus)
```
pazoozoo42: I think it can be something like this
```js
window.addEventListener('focus', () => {
}, { once: true })
```
and then you won't need to extract the callback and call remove listener

pazoozoo42 (dismissed):

PR #418, opened 2025-02-23
Split lintstaged into eslint and prettier

Aivean:
Noticed that `eslint --fix` fails for `scss` files.

Thus splitting the lint-staged config into:
* "*.{js,jsx,ts,tsx}" (runs eslint that includes prettier)
* "*.{json,scss,css}" only runs prettier

### Test

tested locally by modifying all relevant files

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -79,6 +79,10 @@
     "winston": "^3.7.2"
   },
   "lint-staged": {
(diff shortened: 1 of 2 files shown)


4vanger (approved):
Looks ok but maybe should investigate why `eslint --fix` fails?

Aivean:
> Looks ok but maybe should investigate why `eslint --fix` fails?

css is not supported by eslint by default: https://github.com/eslint/css

PR #430, opened 2025-03-02
Replace react-confirm-alert with our own component

Aivean:
Reimplements `react-confirm-alert` with a simple component, better integrated into our app structure. The result is cleaner and smaller.

**Main purpose**: fixes the scroll bug (described in discord).

Demo:

https://github.com/user-attachments/assets/1181535a-13c0-421e-80c5-00ac0363bcc0

---

Demo of the bug (previous behavior, notice the scroll to the top after click on the button):

https://github.com/user-attachments/assets/a906f503-ec1a-46a5-8403-e58a314bfca5

---

### Testing
---

Manually tested all usages of the `confirmAlert`: oauth apps, invites, logout

### DIFF
--- a/frontend/src/Components/ConfirmDialog.module.scss
+++ b/frontend/src/Components/ConfirmDialog.module.scss
@@ -77,7 +77,7 @@
 .primaryButton {
   border: none;
   background: var(--primary);
-  color: var(--onAccent, #fff);
+  color: var(--onAccent);
 
   &:hover {
     background: var(--primaryHover);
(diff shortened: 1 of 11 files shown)


frontend/src/Components/ConfirmDialog.module.scss
```diff
@@ -0,0 +1,115 @@
...
+    cursor: pointer;
+    transition: background-color 0.2s ease;
+  }
+}
+
+/* Button types */
+.primaryButton {
+  border: none;
+  background: var(--primary);
+  color: var(--onAccent, #fff);
```
pazoozoo42: this probably can fallback to white which exists in theme vars, something like this 
```css
color: var(--onAccent, var(--fg));
```

pazoozoo42 (dismissed):

frontend/src/Components/ConfirmDialog.module.scss
Aivean: I didn't notice fallbacks, not sure if they make sense at all, removed.

PR #431, opened 2025-03-02
Fix Topbar shifting when modal/noscroll is shown

Aivean:
when `useNoScroll` is invoked (usually when modals are shown), the Topbar currently shifts to the right, which is annoying. This fixes it.

before / after demo:

https://github.com/user-attachments/assets/c7f309a4-3552-402d-8ab1-46f1255431e8


pazoozoo42 (approved):
I think this may act differently on non macos, but the change shouldn't make it worse

PR #441, opened 2025-03-31
Comments and posts sorting in profile

LazyKarlson:
Comments and posts sorting by rating or by date in user profile

### DIFF
--- a/backend/src/db/repositories/CommentRepository.ts
+++ b/backend/src/db/repositories/CommentRepository.ts
@@ -59,11 +59,11 @@ export default class CommentRepository {
     filter: string,
     page: number,
     perPage: number,
-    sort: string = 'date',
+    sort = 'date',
   ): Promise<CommentRawWithUserData[]> {
     const limitFrom = (page - 1) * perPage
     const orderBy = sort === 'rating' ? 'c.rating desc, c.created_at desc' : 'c.created_at desc'
-    
+
     return await this.db.query(
       `
         select c.*, v.vote
--- a/backend/src/db/repositories/CommentRepository.ts
+++ b/backend/src/db/repositories/CommentRepository.ts
@@ -59,19 +59,22 @@ export default class CommentRepository {
     filter: string,
     page: number,
     perPage: number,
+    sort: string = 'date',
   ): Promise<CommentRawWithUserData[]> {
     const limitFrom = (page - 1) * perPage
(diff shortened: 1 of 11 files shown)


frontend/src/Components/UserProfilePosts.tsx
```diff
@@ -99,6 +115,20 @@ export default function UserProfilePosts(props: UserProfilePostsProps) {
           defaultValue={defaultFilter}
         />
       </div>
+      <div className={styles.feedControlsWrapper}>
```
Aivean, line 118: the filter (including sorting) should be extracted as a separate component, see e.g. https://github.com/spaceshelter/orbitar/blob/47781e6aa4667d50fada8e7b780a454f88d7bea3/frontend/src/Components/FeedFilterComponent.tsx

Aivean:
I'm hesitant to merge this, as it conflicts with Markers (#435), where I extracted feed filter into a separate component and added filters by marker types. I'll think about the best way to approach this conflict.

backend/src/db/repositories/CommentRepository.ts
```diff
@@ -59,19 +59,22 @@ export default class CommentRepository {
...
-            order by c.created_at desc
-            limit :limit_from, :limit_count
-        `,
+        select c.*, v.vote
+        from comments c
+                 left join comment_votes v on (v.comment_id = c.comment_id and v.voter_id = :for_user_id)
+        where
+            c.author_id = :user_id
+          and c.deleted = 0 ${filter ? ' and c.source like :filter ' : ''}
+        order by ${orderBy}
```
Aivean, line 75: need to add index by rating, otherwise the query is too slow

frontend/src/Components/UserProfileComments.tsx
```diff
@@ -134,6 +145,20 @@ export default function UserProfileComments(props: UserProfileCommentsProps) {
...
+        <div className={styles.feedControls}>
+          <a href='#' className={classNames({ [styles.active]: sort === 'date' })} onClick={handleSortChange('date')}>
+            <i className='i i-new'></i>ПО ДАТЕ
+          </a>
+          <a
+            href='#'
+            className={classNames({ [styles.active]: sort === 'rating' })}
+            onClick={handleSortChange('rating')}
+          >
+            <i className='i i-live'></i>ПО РЕЙТИНГУ
```
Aivean, line 158: better to use plus icon here

PR #442, opened 2025-04-04
Cypress

eyesplitter:
[Cypres](https://www.cypress.io/) e2e test framework integration

Run with:
```
cd e2e
npm run cy:open 
```
or
```
cd e2e
npm run test:e2e
```

https://github.com/user-attachments/assets/4caab2f2-3b05-4a5d-bc13-278481b690d3


e2e/cypress.config.ts
```diff
@@ -11,8 +8,8 @@ export default defineConfig({
     viewportWidth: 1280,
     viewportHeight: 720,
     env: {
-      username: process.env.CYPRESS_USERNAME,
-      password: process.env.CYPRESS_PASSWORD,
+      username: 'testuser',
```
Aivean: `process.env` + separate .env file is better: obvious where to change configs, easier to override in CI

e2e/.env
```diff
@@ -0,0 +1,9 @@
+MYSQL_ROOT_PASSWORD=testpass
```
Aivean, line 1: both .env and .env.sample seem redundant

e2e/docker-compose.test.yml
```diff
@@ -0,0 +1,109 @@
+version: '3.9'
```
Aivean, line 1: can't we reuse the existing config?

I think cypress should the service is already running and is set up, setting it up should be responsibility of the user (or CI pipeline).

mysql/test-users.sql
```diff
@@ -0,0 +1,31 @@
+USE orbitar_db;
+
+DELETE FROM user_invites WHERE child_id IN (SELECT user_id FROM users WHERE username LIKE 'testuser%');
+DELETE FROM invites WHERE code LIKE 'invite-testuser%';
+DELETE FROM users WHERE username LIKE 'testuser%';
+
+INSERT INTO users (username, password, email, gender, karma, name, registered_at)
+VALUES 
+  ('testuser1', '$2a$10$otySLU3y9XONpFnPqX.OfeNZrCeq4Ld.zC1RCXPCS4YAEGAlYEstO', 'test1@example.com', 0, 0, 'Test User 1', CURRENT_TIMESTAMP),
```
Aivean, line 9: I think there could be a small javascript generator in cypress that produces this sql, using configured passwords, and pipes it into mysql client, similar to this:

https://github.com/spaceshelter/orbitar/blob/9b5d39b375ba3810d2aed1b1275141f71e97679f/.github/workflows/restore-sanitized-db.yml#L92

frontend/package.json
```diff
@@ -60,7 +60,7 @@
     "sass": "^1.55.0"
   },
   "scripts": {
-    "start": "BABEL_DISABLE_CACHE=1 react-app-rewired start",
+    "start": "npx cross-env BABEL_DISABLE_CACHE=1 react-app-rewired start",
```
Aivean, line 63: is this in scope?

caddy/Caddyfile.test
```diff
@@ -0,0 +1,22 @@
+{
```
Aivean, line 1: why do we need a separate caddy file?

backend/package.json
```diff
@@ -8,8 +8,10 @@
     "watch:dev": "nodemon --watch src --ext ts --exec \"npm run start:dev\"",
     "build": "npx tsc",
     "start:production": "NODE_ENV=production node build/app.js",
+    "start:test": "NODE_ENV=test node build/app.js",
```
Aivean, line 11: do we really need another env here?

backend/src/managers/RateLimitManager.ts
```diff
@@ -0,0 +1,23 @@
+import rateLimit, { Options } from 'express-rate-limit'
+
+export class RateLimitManager {
+  private static isTestEnv = process.env.NODE_ENV === 'test'
+
+  /**
+   * Creates a rate limiter for routes.
+   * If `NODE_ENV=test`, simply skips the limiter.
```
Aivean, line 8: how about we introduce an env flag that disables limits instead of the whole env?

eyesplitter: 100% agree

backend/package.json
eyesplitter: Yes we do. For bypassing rateLimiter limits

caddy/Caddyfile.test
eyesplitter: I'm not sure

PR #443, opened 2025-04-10
Add entities doc

Aivean:


docs/entities.md
```diff
@@ -0,0 +1,267 @@
...
+  title?: string
+  content?: string
+  rating: number
+  comments: number
+  newComments: number
+  canEdit?: boolean
+  bookmark?: boolean
+  watch?: boolean
+  vote?: number
+  lastReadCommentId?: number
```
eyesplitter, line 104: it seems we don't use "id" postfix for ID entities

eyesplitter: take a look on `author` field

docs/entities.md
```diff
@@ -0,0 +1,267 @@
...
+  grants: string
+  userId: number
+  author: UserBaseEntity
+  // Other fields...
+}
+```
+
+### Raw Referencing Entity Types (Skip Direction)
+
+Some newer features like Polls have Raw types that reference Entity types, inverting the usual pattern:
```
eyesplitter, line 214: we uses only enums located in Entity

Raw
```
export interface PollSettingsRaw {
  allowMultipleChoice: boolean
  resultVisibility: ResultVisibility
  allowVoteRescinding: boolean
  voteAccess: VoteAccess
}
```
ResultVisibility
```
export enum ResultVisibility {
  ALWAYS = 'always',
  AFTER_VOTE = 'after_vote',
  AFTER_VOTE_END = 'after_vote_end',
}

```

eyesplitter (changes requested):

PR #444, opened 2025-04-13
Consolidate poll's canShowResults checks on backend

Aivean:
A single place (on the backend) to calculate whether to return poll results when:
* viewing the poll
* viewing the poll voters

### DIFF
--- a/backend/src/managers/PollManager.ts
+++ b/backend/src/managers/PollManager.ts
@@ -48,7 +48,7 @@ export default class PollManager {
     return pollId
   }
 
-  enrichPoll(poll: PollRaw, includeVotes = true, userVotes: number[] = []): PollInfo {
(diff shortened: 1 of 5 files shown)


backend/src/managers/PollManager.ts
```diff
@@ -132,22 +134,18 @@ export default class PollManager {
...
       throw new PollError('invalid-option', 'Invalid option ID', 400)
     }
 
-    const [poll] = await this.getPollsByIds([pollId])
+    const [poll] = await this.getPollsByIds([pollId], userId, IncludePollVotes.AUTO)
     if (!poll) {
       throw new PollError('not-found', 'Poll not found', 404)
     }
 
-    if (
```
Aivean, line 145: also fixes one missing case here

eyesplitter (approved):

PR #446, opened 2025-04-15
Updated scripts to support Windows

Lexy2:
NPM scripts are updated to support running Orbitar in Windows with less hassle.

Additionally, ran `npm audit fix` to update vulnerable packages to their non-vulnerable non-breaking versions.


Aivean (commented):
would appreciate more context on what exactly this fixes and how

eyesplitter (dismissed):

Aivean (changes requested):
sorry, I don't understand the change, why are we changing the pinned package versions here?

Lexy2:
Do you mean npm in the backend?

Aivean:
I don't understand why you're changing unrelated package versions in this PR (package-lock.json in both front and back).

Here I ran `npm install` on the changed `package.json`'s from this PR, look at the diff in `package-lock.json` specifically:
https://github.com/spaceshelter/orbitar/compare/windows-debugging-package-json?expand=1

Lexy2:
> I don't understand why you're changing unrelated package versions in this PR (package-lock.json in both front and back).
> 
> Here I ran `npm install` on the changed `package.json`'s from this PR, look at the diff in `package-lock.json` specifically: https://github.com/spaceshelter/orbitar/compare/windows-debugging-package-json?expand=1

Just fixing vulnerabilities on-the-spot. If that's a crime, no problem, I can remove it.

Aivean:
>  If that's a crime, no problem, I can remove it.

It's not a crime, but it's something that is not apparent from the PR description. When PR says it does A but it does B it raises questions. I'm ok with merging it as is, but let's wait until after the release, I'd prefer not to risk introducing new bugs by blindly updating deps.

In general, it's a good practice to keep the scope of the changes as small as reasonable and reduce the back-and-forth by providing a brief description of the changes that are not obvious.

Lexy2:
Let me update the description then

PR #450, opened 2025-04-18
Fix infinite refreshing poll

eyesplitter:
Replaced the poll dependency in `useEffect `with a derived `expiresTs `value (timestamp of `poll.expires`) to avoid unnecessary re-renders and infinite calls to `fetchPoll`.
This ensures the refresh logic only triggers when the expiration time actually changes.


eyesplitter:
the fix is #451

PR #453, opened 2025-04-24
Add a tooltip for long poll options

Aivean:
poll-tooltip


eyesplitter (commented):
Overall looks great!
There is on small thing I've noticed:
When the option content width is a little bit smaller then input width, the `...` button is not rendered after submitting the vote
not_enough

eyesplitter (dismissed):

Aivean:
> Overall looks great! There is on small thing I've noticed: When the option content width is a little bit smaller then input width, the `...` button is not rendered after submitting the vote

good find. fixed

PR #465, opened 2025-06-14
Add diff toggle to history viewer

pazoozoo42:
## Summary
- add react-diff-viewer-continued dependency
- implement diff toggle in HistoryComponent
- style history diff toggle button

## Testing
- `npm run lint` *(fails: ESLint config missing)*
- `npm test --silent`

------
https://chatgpt.com/codex/tasks/task_e_684d66757f80832c82b11713c7939f99


Aivean:
`frontend/package-lock.json` modifications should not be this extensive after a single dep addition

Aivean (commented):
Couple of notes:
1. the diff button position is not obvious. Maybe we should pin it to the top/bottom of the content view?
2. there is a concrete UX issue when the history is long and the button gets scrolled from the view, and if you want to see diffs for each item, you have to scroll back-and-forth to click the diff button (as currently the diff view resets every time)
3. dark theme colors are way off, looks jarring. light theme is better, but not ideal
4. "Expand 61 lines ..." — can we translate it?

otherwise, very useful change, I like it!

pazoozoo42:
> the diff button position is not obvious. Maybe we should pin it to the top/bottom of the content view?

I made the versions list header to be sticky, I think it looks good and header with buttons always visible now
image


> dark theme colors are way off, looks jarring. light theme is better, but not ideal

made the colors more subtle

> "Expand 61 lines ..." — can we translate it?

done

Aivean (approved):

PR #466, opened 2025-06-14
helper doc for Codex agent

pazoozoo42:


.codex.yml
```diff
@@ -0,0 +1,148 @@
...
+  root:
+    frontend: "React + TypeScript SPA"
+    backend: "Node.js + TypeScript (Express)"
+    feed: "Rust microservice for feed generation"
+    search: "Elasticsearch indexing utilities"
+    docs: "Architecture and API docs"
+    docker: "Dev/prod docker-compose files"
+
+  frontend:
+    port: 5000
```
Aivean: do we actually want codex to know that?  does it run the service successfully? if not, this information is distracting

.codex.yml
```diff
@@ -0,0 +1,148 @@
...
+    components: functional-only
+    hooks: strict
+    props_interface_threshold: 3
+  comments: "block only (/* … */ for >2 lines)"
+  scss:
+    indent: 2
+    max_nesting: 3
+    no_hard_coded_values: true
+
+workflow:
```
Aivean: same issue here. not sure what codex will do with this information.

Aivean:
also, is `.codex.yml` just a free-from document?

can we make `AGENTS.md`, that codex seem to look for:

```
I’ll check the AGENTS.md file at the root of the repo.

The AGENTS.md file isn't in the first two directories—let's look further down.

No AGENTS.md present, so I'll proceed without any instructions. Let's investigate the repo.
```

.codex.yml
pazoozoo42: changed to `AGENTS.md` and applied improvements suggested by Codex itself. In general I think this is a good starting point and we can keep improving it in the future

Aivean (approved):

PR #470, opened 2025-07-03
Implement exponential edit delay for posts and comments

Aivean:
image

Implements an exponential delay restriction for post/comment editing when users are downvoted below the first negative effective karma threshold (-10). The delay starts at 5 minutes and doubles with each edit of the same content.

### DIFF
--- a/frontend/src/API/APIBase.ts
+++ b/frontend/src/API/APIBase.ts
@@ -4,7 +4,7 @@ type APIResponseError = {
   result: 'error'
   code: string
   message: string
-  meta?: any
+  meta?: Record<string, unknown>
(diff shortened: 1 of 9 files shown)


frontend/src/API/APIBase.ts
```diff
@@ -4,6 +4,7 @@ type APIResponseError = {
   result: 'error'
   code: string
   message: string
+  meta?: any
```
pazoozoo42:
```suggestion
  meta?: Record<string, unknown>
```

pazoozoo42 (dismissed):
makes sense, didn't test

PR #477, opened 2025-08-23
Add images gallery

meandrewdev:
I put the icon for the editor from the free set, change it if it doesn't suit you


meandrewdev:
Tag Usage:
```
<gallery show-arrows="false" show-indicators="false" show-thumbnails="false" auto-play-interval="5">
<img src="src" alt="Caption text" />
...
</gallery>
```

`show-arrows`: Whether to show arrows on the sides that switch the image
`show-indicators`: Whether to show dots at the bottom to go to a specific image
`show-thumbnails`: Whether to show previews at the bottom
`auto-play-interval`: Auto-switch images in seconds

`alt` in ``: Caption for each image at the bottom

backend/src/parser/TheParser.ts
```diff
@@ -87,6 +88,7 @@ export default class TheParser {
       app: ['a', 'mailbox', 'mail'],
       expand: ['a', 'mailbox', 'mail'],
       poll: ['a', 'mailbox', 'mail', 'b', 'i', 'u', 'strike', 'irony', 'spoiler'],
+      gallery: ['a', 'blockquote', 'pre', 'mailbox', 'mail', 'b', 'i', 'u', 'strike', 'irony'],
```
Aivean, line 91: what about `<video`?

have you tested magic text conversions (youtube, vimeo, coub, mentions, links)?

depending on what you expect inside the `<gallery`, calling `parseChildNodes` recursively might be too permissive

backend/src/parser/TheParser.ts
```diff
@@ -683,6 +687,17 @@ export default class TheParser {
     return { ...result, text }
   }
 
+  parseGallery(node: Element): ParseResult {
+    const showArrows = node.attribs['show-arrows'] !== 'false'
```
Aivean: let's prefix those custom attributes with `data-` to make them standard-compliant

backend/src/parser/TheParser.ts
```diff
@@ -543,7 +545,9 @@ export default class TheParser {
       )
     }
 
-    return { text: `<img src="${encodeURI(url)}" alt=""/>`, mentions: [], urls: [], images: [url] }
+    const alt = node.attribs['alt'] ? htmlEscape(node.attribs['alt']) : ''
```
Aivean, line 548: this looks like a double escape. please cover with tests

backend/test/parser/TheParser.test.ts
```diff
@@ -270,6 +270,46 @@ test('parse expand tag', () => {
   )
 })
 
+test('parse gallery tag', () => {
+  // Basic gallery
+  expect(p.parse('<gallery><img src="https://orbitar.media/image1.jpg" /></gallery>').text).toEqual(
+    '<div class="gallery" show-arrows=true show-indicators="true" show-thumbnails="true" auto-play-interval="0"><img src="https://b.orbitar.media/image1.jpg" alt=""/></div>',
```
Aivean: > `show-arrows=true`

by convention, in html the presence of the attribute usually indicates `true`. meaning, either consider setting the defaults for the keys to be `false`, or inverting the names (e.g. `no-indicators`, `no-thumbnails`).

even better design, imo, is to remove all these options for the sake of simplicity and consistency of the UI (picking the best defaults or inferring them)

Aivean (changes requested):
Overall looks nice! but there are couple of things (some of them important) that need to be addressed.

Bugs:
1. when there are multiple galleries on the page, they all simultaneously react to keyboard arrows.
2. when images in the gallery are of non uniform height, the layout of the page shifts when navigating.
3. zoom works — good, but:
  - only on the images that are big enough (e.g. if first image is not zoomable, you have to navigate to the second one to open zoom). would make more sense make all images zoomable since user intentionally put them in the gallery
  - keyboard navigation of the gallery still works in the zoomed view, but the image doesn't change. a better UX would be to change the image without exiting zoom. and even better UX would make zoomed view display gallery thumbnails and arrows (for mobile)
4. white dots (and navigation arrows) over thumbnails are not visible if image has white background
5. editor preview closes when interacted with gallery
  
---

Also consider the following idea for better gallery creation UX:

1. add a checkbox "gallery" to the media upload form:
image

2. [optionally, but extremely useful] allow uploading multiple files in the media upload form, automatically set the "gallery" checkbox to true when multiple files are uploaded.

This has multiple advantages:
1. avoids having an extra icon in the edit bar (it's size is a premium on mobile, especially in the long threads). also gallery icon is similar to the existing image icon — could be confusing
2. users don't have to deal with manually editing the tag (making sure that uploaded images stay within the tag) — just drop multiple files and done

---

Alternatively, maybe we can just turn any 2+ consecutive `<img` tags into a gallery (in the parser)? Could be a simpler UX that requires no additional effort from the users.

tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
```
Aivean: accidental commit?

frontend/src/Components/CreateCommentComponent.tsx
```diff
@@ -25,6 +25,7 @@ import ThemeToggleComponent from './ThemeToggleComponent'
 
 import { ReactComponent as CodeIcon } from '../Assets/code-slash.svg'
 import { ReactComponent as ExpandIcon } from '../Assets/expand.svg'
+import { ReactComponent as Gallerycon } from '../Assets/gallery.svg'
```
Aivean: Gallery**I**con?

backend/src/parser/TheParser.ts
meandrewdev: > what about <video?

You mean the videos in the gallery itself? I've only done images so far, videos maybe later

meandrewdev:
1. disabled the keyboard outside the zoom
2. fixed
3. done, except for the preview
4. added background
5. fixed

added checkboxes when uploading and removed the icon from the editor. but I didn't write uploading several files at the same time, it's necessary to seriously redesign the component, I won't do it now, maybe later
added video display, now you can add links with video and tag video
and other small problems also fixed

backend/src/parser/TheParser.ts
```diff
@@ -545,7 +545,7 @@ export default class TheParser {
       )
     }
 
-    const alt = node.attribs['alt'] ? htmlEscape(node.attribs['alt']) : ''
+    const alt = node.attribs['alt'] ?? ''
```
Aivean: no, this is vulnerable:
```
<img src="https://example.com/pic.jpg" alt='"><b>INJECTED</b>'>
```

the previous code was probably correct, could you just cover it with tests that ensures that quotes are escaped or stripped correctly?

backend/src/parser/TheParser.ts
Aivean: > videos maybe later

wow, videos "just work" after the latest update.


what I mean, currently `gallery` still uses recursive logic to parse children and returns the results "as is", post-filtering of the elements done only on the client.

meaning, it's possible to add all kind of stuff inside the gallery:
* expanding links
* another gallery
* blocks
and it will be still parsed and returned to the client.

It would be more consistent with existing code to do filtering inside the Parser. Could be done by calling `TheParser.parseNode` for the child nodes inside the gallery and retaining only `<img` tags returned (would also cover videos).

Some of the corner cases that current logic allows, that warrant more restrictive behavior:
* hidden user mentions inside the gallery
* rich components (such as Poll) would be processed and rendered on the client inside the gallery div, before overridden by the gallery

Aivean:
> 1. disabled the keyboard outside the zoom
> 2. fixed
> 3. done, except for the preview
> 4. added background
> 5. fixed
> 
> added checkboxes when uploading and removed the icon from the editor. but I didn't write uploading several files at the same time, it's necessary to seriously redesign the component, I won't do it now, maybe later added video display, now you can add links with video and tag video and other small problems also fixed


Looks much better now, thanks!


Is there a reason to have four options in the gallery? How would you envision people using them (i.e. what are they solving)?

What do you think about an alternative UX — no gallery creation controls, just detect consecutive "gallery-compatible" elements in TheParser and wrap them into the gallery automatically?

backend/src/parser/TheParser.ts
meandrewdev: Got it. Yes, it would be better to filter in TheParser, you are right
Ok, I will fix the code soon.

meandrewdev:
The full interface is a bit cluttered, maybe someone will want to remove the preview at the bottom and have enough dots and arrows. Also, perhaps some people will not be able to see the dots well or they may want to keep the previews. Maybe someone will want to make a slideshow without controls at all. I think it's worth giving people the opportunity to customize, and they will figure out how to use it :)

backend/src/parser/TheParser.ts
Aivean: >  Could be done by calling TheParser.parseNode for the child nodes inside the gallery and retaining only <img tags returned (would also cover videos).

~~i thought a bit more about it and I believe the cleanest way to do that would be to extend `ParseResult` with the field that retains meta-info about the top-level parsed node. Let me propose a PR for that.~~

nvm, it's a bit more complicated than that: `parseText` returns an unstructured mix of everything. Need to think more about that.

backend/src/parser/TheParser.ts
meandrewdev: Oh, sorry, misunderstood. Added this option to the test cases
For some reason I thought this field was already being processed somewhere else 😅

Aivean:
> The full interface is a bit cluttered, maybe someone will want to remove the preview at the bottom and have enough dots and arrows. Also, perhaps some people will not be able to see the dots well or they may want to keep the previews. Maybe someone will want to make a slideshow without controls at all. I think it's worth giving people the opportunity to customize, and they will figure out how to use it :)

if you think the interface is cluttered, let's make it clean for everybody and make sure that everybody has the best UX.

currently you're not giving **consumers** the options to optimize interface for themselves, you're giving **authors** the knobs to configure for everybody else.  Maybe doing the former instead of the latter is a better way to achieve the goal (i.e. move gallery config into settings, and allow everybody to configure it for themselves). 

But I'd argue that having less options is better in general -> less things to maintain, better predictability of the interface behavior, less permutations to test, less knobs for users to think about.

So I'd suggest to simplify the whole thing and pick a reasonable set of defaults (if they become a contention, we can change or make them configurable later).

Aivean:
Found a bug in the latest version: arrow keys and buttons in zoomed view are stuttering.

meandrewdev:
Removed image switching in zoom due to the fact that videos are now supported and it became inconvenient. Completely removed keyboard control due to support for multiple galleries on a page

backend/src/parser/TheParser.ts
meandrewdev: Done

frontend/src/Components/MediaUploader.tsx
```diff
@@ -169,86 +247,153 @@ export default function MediaUploader(props: MediaUploaderProps) {
...
+        throw new Error('UPLOAD FAILED: no link')
+      }
+    } finally {
+      setUploading(false)
+    }
+  }
+
+  useEffect(() => {
+    if (props.mediaData) {
+      readFile(props.mediaData).then((data) => setUploadArray([...uploadArray, data]))
```
Aivean: why this is done here? `handlePaste` does the same thing?

frontend/src/Components/GalleryComponent.tsx
```diff
@@ -0,0 +1,490 @@
...
+
+  const goToNext = (e: React.MouseEvent | undefined = undefined): void => {
+    if (isTransitioning) return
+    e?.stopPropagation()
+
+    goToSlide(currentIndex + 1)
+  }
+
+  const goToSlide = (index: number): void => {
+    if (isTransitioning || index === currentIndex || index < 0 || currentIndex >= elements.length) return
```
Aivean: `index >= elements.length` ?

frontend/src/Components/MediaUploader.tsx
```diff
@@ -169,86 +247,153 @@ export default function MediaUploader(props: MediaUploaderProps) {
...
+    }
+
+    document.addEventListener('paste', handlePaste)
+    return () => {
+      document.removeEventListener('paste', handlePaste)
     }
+  }, [])
```
Aivean, line 424: probably need to add `handlePaste` as a dependency here.

also note, that without specifying dependencies the captured `uploadArray` would be always the same here.

Aivean (approved):

PR #478, opened 2025-08-24
Fix push notifications

meandrewdev:


Aivean:
please rebase only the necessary commit on `dev` (currently based on the image gallery)

frontend/src/Services/PushService.ts
```diff
@@ -11,7 +11,7 @@ export class PushService {
       return
     }
 
-    const registration = await navigator.serviceWorker.getRegistration('/service.js')
+    const registration = await navigator.serviceWorker.getRegistration('/serviceWorker.ts')
```
Aivean: this is not correct. see `build` folder

meandrewdev: Oh, I see. It was working in debug mode, I'll fix it now

PR #479, opened 2025-08-24
Fix push notifications

meandrewdev:


backend/src/managers/NotificationManager.ts
```diff
@@ -193,7 +195,9 @@ export default class NotificationManager {
     this.userCache.deleteUserStatsCache(forUserId)
 
     // send push in background
-    this.sendWebPush(forUserId, notification).then().catch()
+    this.sendWebPush(forUserId, notification)
+      .then()
+      .catch((err) => this.logger.error(err))
```
Aivean: maybe add a bit more info?

```suggestion
      .catch((err) => this.logger.error('Failed to send web push', { forUserId, error })
```

Aivean (dismissed):
lgtm overall, see the nit

backend/src/managers/NotificationManager.ts
Aivean: so, what's the resolution?

Aivean (approved):

PR #480, opened 2025-12-28
Revalidate html cache

Aivean:
This should fix the "white page" issue we have after deployments.

  Currently all our static pages (including SPA html) are cached indefinitely, meaning that after deployment users' browsers serve stale `index.html` from disk cache. The old HTML references JS bundles (e.g., main.ede71b45.js) that no longer exist on the server, resulting in a white page with Unexpected token '<' errors.

  Solution: Add explicit Cache-Control headers in Caddy:
  - /static/* → max-age=31536000, immutable (hashed assets, safe to cache forever)
  - Everything else → no-cache, no-store, must-revalidate (always revalidate HTML/SPA routes)

  This ensures browsers always fetch fresh index.html while still caching hashed JS/CSS bundles efficiently.

### DIFF
--- a/caddy/Caddyfile.local
+++ b/caddy/Caddyfile.local
@@ -20,8 +20,21 @@ http://{$SERVER_DOMAIN} {
     }
   }
 
+  # Hashed static assets (JS/CSS) - cache for 1 year (immutable)
+  @static {
+    path /static/*
+  }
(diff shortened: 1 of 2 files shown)


Lexy2 (approved):
LGWW

PR #481, opened 2026-01-11
Fix favicon not updating after closing notification popup

Aivean:
### Problem

  When marking notifications as read and closing the popup, the favicon would revert to showing the "unread" badge even though there were no unread notifications. This happened because Chrome caches favicons per browser history entry, and the notification popup uses history.pushState/history.back() for back button support.

### Solution

  - Add cache-busting fragment (#timestamp) to favicon URLs to prevent Chrome from using stale cached versions
  - Listen for popstate events and re-apply the correct favicon after history navigation


pazoozoo42 (approved):

4vanger (approved):
:shipit:

PR #483, opened 2026-01-16
Gallery bug fixes

meandrewdev:
Changes:
1. Hidden indicators in normal view as unnecessary.
2. Fixed video height so that controls are visible.
3. Removed legacy zoom.
4. Fixed zooming so that the entire image fits on the screen without cropping at the edges.
5. Fixed the indicators — with a large number of images, it didn't look very nice with line breaks. Now the indicators are displayed in a single line, and those that don't fit are hidden, and when switching images, the list moves forward.

Several dots are hidden in front here:
image

After turning the image, several dots are now hidden behind. If you turn back, it will be like in the previous screenshot, the list will shift:
image

6. Fixed the preview list - now when switching images, the scroll will move so that the active image is visible and there are 2 more before or after it. This is much more convenient and attractive.
7. Made submit in the form of uploading by pressing Enter.
8. Fixed incorrect behavior when switching images with the back arrow on the second image. Also fixed switching forward on the next to last image image. When double-clicking, the image either did not switch or zoomed in; this has now been fixed
9. Minor preview design change
image


frontend/src/Components/GalleryComponent.tsx
```diff
@@ -100,31 +96,21 @@ const GalleryComponent: React.FC<GalleryComponentProps> = ({
     }
   }, [isControlled, onCollapseProp])
 
-  const maxHeight = 450
+  const maxHeight = 500
   const fallbackElementHeight = 400
 
-  // Disable drag in expanded mode to allow zoom-pan-pinch to work (except for videos)
```
Aivean, line 106: why are you removing the comments here?

Aivean:
Please write the description of the PR, listing all the changes and rationale behind them. Screenshots would be appreciated to let the reviewers know what to look at.

frontend/src/Components/GalleryComponent.tsx
```diff
@@ -484,6 +467,107 @@ const GalleryComponent: React.FC<GalleryComponentProps> = ({
   )
 }
 
+interface SlidingIndicatorsProps {
```
Aivean, line 491: maybe let's move this component into a separate file?

frontend/src/Components/GalleryComponent.tsx
meandrewdev: I used a neural network to solve the scaling problem, instructed it to add a minimum of comments, and apparently it decided to clean those up as well. Now I'll return everything

frontend/src/Components/MediaUploader.tsx
```diff
@@ -72,6 +73,23 @@ export default function MediaUploader(props: MediaUploaderProps) {
     }
   }, [uploadArray.length])
 
+  useEffect(() => {
+    const handler = (e: KeyboardEvent) => {
+      if (e.key === 'Enter' && uploadArray.length > 0) {
```
Aivean: this looks like a bug, `uploadArray.length` should be in dependencies, otherwise only the first value gets captured

frontend/src/Components/GalleryComponent.tsx
```diff
@@ -699,6 +763,13 @@ function GalleryElementComponent({
               alt={image.alt}
               className={classNames(styles.image, styles.noDragging, styles.imageExpanded)}
               onClick={handleImageClick}
+              style={{
```
Aivean: why inline styles?

meandrewdev:
> Please write the description of the PR, listing all the changes and rationale behind them. Screenshots would be appreciated to let the reviewers know what to look at.

Okay, I'll do it now. Regarding the rationale, most of it is bug fixes that people have mentioned. I'll explain the rest, there isn't much.

frontend/src/Components/GalleryComponent.tsx
```diff
@@ -484,6 +467,107 @@ const GalleryComponent: React.FC<GalleryComponentProps> = ({
...
+      const availableWidth = parent.offsetWidth * 0.8 - CONTAINER_PADDING * 2
+      const dotWidth = INDICATOR_SIZE + INDICATOR_GAP
+      const max = Math.floor((availableWidth + INDICATOR_GAP) / dotWidth)
+      setMaxVisible(Math.max(3, max))
+    }
+
+    calculateMaxVisible()
+
+    const resizeObserver = new ResizeObserver(calculateMaxVisible)
+    resizeObserver.observe(container.parentElement!)
```
Aivean: let's do a proper non-null check here

frontend/src/Components/MediaUploader.tsx
```diff
@@ -260,39 +271,6 @@ export default function MediaUploader(props: MediaUploaderProps) {
     // If no valid URLs, let paste happen normally (user can edit and press Enter)
   }
 
-  const addUrlsFromInput = () => {
```
Aivean, line 263: why was this logic removed? is it intentional?

Aivean (changes requested):

frontend/src/Components/GalleryComponent.tsx
meandrewdev: It is not used anywhere else, so I think it is unnecessary to put it in a separate file. I think it would be better to put it separately if it is going to be used somewhere else

frontend/src/Components/MediaUploader.tsx
meandrewdev: No, this is normal behavior. The check is necessary so that pressing Enter does not submit without files

frontend/src/Components/MediaUploader.tsx
meandrewdev: Yes, because files are added automatically, and this behavior was activated by pressing Enter. I didn't see any use for this feature, and pressing Enter now performs a submit

frontend/src/Components/MediaUploader.tsx
Aivean: yes, but this effect gets executed only on component mount and the handler captures the first value of the `uploadArray` it sees.  when `uploadArray` is updated (`setUploadArray`), the handler still has the old value of the array captured. that's why you need to add `uploadArray.length` to the dependencies of the effect — handler needs to be recreated every time the array is updated.

frontend/src/Components/MediaUploader.tsx
Aivean: what about urls?  multiple urls? editing urls?  could you please verify that the UX for that is not broken?

meandrewdev: I checked that URL insertion works automatically, multiple URLs work too, and multiple URLs with junk strings also work as expected. But I just checked URL editing - it doesn't insert automatically here. But pressing Enter now submits it. Which button should I bind it to, or what's the best way to do it?

Aivean:
the very long alt text is now broken in zoomed-in mode:
image


previously it was collapsed by default


---

upd: actually, nvm, it got broken before release :(

frontend/src/Components/GalleryComponent.tsx
meandrewdev: Fixed

frontend/src/Components/GalleryComponent.tsx
meandrewdev: Fixed

frontend/src/Components/MediaUploader.tsx
meandrewdev: Fixed

meandrewdev:
> the very long alt text is now broken in zoomed-in mode: image
> 
> previously it was collapsed by default
> 
> upd: actually, nvm, it got broken before release :(

Fixed

frontend/src/Components/MediaUploader.tsx
meandrewdev: Fixed

frontend/src/Components/GalleryComponent.tsx
meandrewdev: Fixed

frontend/src/Components/MediaUploader.tsx
Aivean: I think the ideal UX would allow to separately confirm the URL that is being typed/edited (separate button and Enter), and would also allow to edit existing urls.

Enter could be double-function: when url is edited, first enter press confirms it. Otherwise enter submits the form.

But, it think what's implemented currently is good enough, I don't want to blow up the scope of this PR.

Aivean:
the default behavior for small images changes (in both zoomed-in and gallery views):

before:
image


now:
image


basically, small images are no longer expanded to fit.

meandrewdev:
> the default behavior for small images changes (in both zoomed-in and gallery views):
> 
> before: image
> 
> now: image
> 
> basically, small images are no longer expanded to fit.

Can we leave it as it is? Do we have to stretch them?
If the image is very small, it doesn't look very good when stretched

meandrewdev:
Added one more fix

Aivean:
> Can we leave it as it is? Do we have to stretch them? If the image is very small, it doesn't look very good when stretched

I don't like it, but ok



one more thing, there is that "border" that eats space on mobile:
image


but we can fix that in a followup PRs

Aivean (changes requested):
found another blocking issue: 

post autocut is now broken when gallery is present. see staging

----

or maybe not, now it seems to work, weird. let me test some more

frontend/src/Components/MediaUploader.tsx
```diff
@@ -469,7 +464,7 @@ export default function MediaUploader(props: MediaUploaderProps) {
         onDragLeave={handleDragLeave}
         onDrop={handleDrop}
       >
-        <form className={styles.controls} onSubmit={handleUpload}>
+        <form ref={formRef} className={styles.controls} onSubmit={handleUpload} onKeyDown={handlePressEnter}>
```
Aivean: why are you registering `handlePressEnter` two times?

frontend/src/Components/MediaUploader.tsx
```diff
@@ -72,6 +73,23 @@ export default function MediaUploader(props: MediaUploaderProps) {
     }
   }, [uploadArray.length])
 
+  const handlePressEnter = (e: React.KeyboardEvent<HTMLFormElement> | KeyboardEvent) => {
```
Aivean: functionally this is correct, but you could wrap this in callback to not register event listener every render:


```
 const handlePressEnter = useCallback((e: ...) => {
    // ...
    if (uploadArray.length > 0) formRef.current?.requestSubmit()
  }, [uploadArray.length])
  ```

meandrewdev:
> one more thing, there is that "border" that eats space on mobile: <img alt="image" width="996" height="462"

Removed the padding in zoom completely, I think it looks good on all devices

meandrewdev:
> or maybe not, now it seems to work, weird. let me test some more

I also see on stage that caption cropping isn't working, which is strange. It works locally for me, so I don't understand why

Aivean:
> > or maybe not, now it seems to work, weird. let me test some more
> 
> I also see on stage that caption cropping isn't working, which is strange. It works locally for me, so I don't understand why

staging has the same version as the prod (before the fixes)


I'm talking about post autocut:
image

it seems to be working intermittently for the galleries, not sure why (and it might not be caused by this PR)

frontend/src/Components/MediaUploader.tsx
meandrewdev: I was solving a problem when I pressed Enter in the input field, and for some reason it registered in the post and the window just closed. The solution turned out to be something else, but I decided not to remove it

frontend/src/Components/MediaUploader.tsx
meandrewdev: Done

frontend/src/Components/MediaUploader.tsx
Aivean: could you please consolidate? registering event listener in two places is confusing

meandrewdev:
> > > or maybe not, now it seems to work, weird. let me test some more
> > 
> > 
> > I also see on stage that caption cropping isn't working, which is strange. It works locally for me, so I don't understand why
> 
> staging has the same version as the prod (before the fixes)
> 
> I'm talking about post autocut: image
> 
> it seems to be working intermittently for the galleries, not sure why (and it might not be caused by this PR)

This is how it is for me
image

frontend/src/Components/MediaUploader.tsx
meandrewdev: Removed

Aivean:
> This is how it is for me 

that is correct behavior, but again, the staging has version of the code BEFORE this PR.  I'm seeing some glitching locally (when I create a new post with gallery and long text after it, autocut seems intermittent)

Aivean (approved):

Aivean:
ok, lgtm

thanks for the fast fixes!

PR #484, opened 2026-01-17
Fixes

meandrewdev:
Fix autocut height


Aivean:
@meandrewdev there are extra commits again.

PR #503, opened 2026-01-23
fixed route save

### LINKED ISSUE #487: Save the selected feed in localStorage so that it does not constantly reset to "Все"

meandrewdev:
When navigating to / in PWA, the saved path was not used and the transition always went to subscriptions. This behavior has now been fixed

### DIFF
--- a/frontend/src/Pages/FeedPage.tsx
+++ b/frontend/src/Pages/FeedPage.tsx
@@ -1,8 +1,9 @@
 import React, { useEffect, useState } from 'react'
-import { Link, useMatch, useSearchParams } from 'react-router-dom'
+import { Link, useLocation, useMatch, useSearchParams } from 'react-router-dom'
 
 import classNames from 'classnames'
 import { observer } from 'mobx-react-lite'
+import useLocalStorage from 'use-local-storage'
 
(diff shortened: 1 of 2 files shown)


frontend/src/Pages/FeedPage.tsx
```diff
@@ -20,22 +21,32 @@ const FeedPage = observer(() => {
...
+
+      case '/':
+        if (!savedRoute || savedRoute === '/') {
+          baseRoute = '/subscriptions'
+          feedType = 'subscriptions'
+          break
+        }
+        break
+
+      case '/subscriptions':
```
Aivean, line 47: missing explicit `/posts` case

Aivean (commented):
with this change the `/` route shows the saved choice, while `/posts`, `/subscriptions`, etc are still used for no reason. this is confusing when `/` and (e.g.) `/posts` correspond to the same feed.

the more consistent approach would be to save the feed in db, and have a single route `/`  (similar to feed sorting)

as an intermediate solution we can save the default feed in local storage

the legacy routes (`/posts`, `/subscriptions`, ) should still work but redirect to `/` (and change the saved feed setting)

PR #504, opened 2026-01-23
New comments navigation & scroll to top\bottom

### LINKED ISSUE #495: Implement navigation through unread comments and a scroll up button

meandrewdev:
I added 2 buttons for quick scrolling to the top and bottom, it can be quite useful when you want to go back to the post and read it again, or when you want to write a root comment, but have to scroll to the bottom for a long time to do it

Also, added buttons for cyclic navigation through new comments - this is a very useful feature that allows you to focus on new content at once, without having to scroll through large branches for a long time. The active comment is circled by a dotted line - it's thin and unobtrusive, but you can change it if you need to
Navigation by comments can be disabled in the account settings, if this feature will bother someone.

image

https://github.com/user-attachments/assets/4e6f8719-d2d5-4bf2-86f6-2181811a938c

### DIFF
--- a/frontend/src/Assets/chevron-down.svg
+++ b/frontend/src/Assets/chevron-down.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="lucide lucide-chevron-down-icon lucide-chevron-down">
+    <path d="m6 9 6 6 6-6"/>
+</svg>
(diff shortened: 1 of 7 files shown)


Aivean (commented):
1. navigation implementation is overcomplicated: why not to just rely on existing scrollToComment behavior and styles?

2. navigation is buggy: try pressing on the button in rapid succession

3. the navigation buttons are obtrusive on mobile. while it's possible to disable comment navigation, the new home/end buttons at the bottom are always there, eating space. Can we hide them behind a smaller slider? or maybe there is a fundamentally different better UI component for this?

Aivean:
4. navigation over the new comments doesn't take the current scroll position into account at all. this is extremely unintuitive

PR #505, opened 2026-02-09
Add FeedController unit tests

Aivean2:
## Summary
- Added 22 unit tests for `FeedController` — the first API controller test file in the project
- Covers all 5 feed endpoints: `feedSubscriptions`, `feedAll`, `feedPosts`, `feedWatch`, `saveFeedSorting`
- Tests auth checks, success paths, restricted posts handling, and error paths
- All existing 81 tests continue to pass (103 total now)

## Test plan
- [x] All 22 new tests pass
- [x] Full test suite (103 tests, 7 suites) passes
- [x] No changes to production code

### DIFF
--- a/backend/test/api/FeedController.test.ts
+++ b/backend/test/api/FeedController.test.ts
@@ -1,0 +1,484 @@
+import FeedController from '../../src/api/FeedController'
+import { FeedSorting } from '../../src/api/types/entities/common'
+
+// --- Mock factories ---
+
+function createMockEnricher() {
+  return {
+    enrichRawPosts: jest.fn().mockResolvedValue({
+      posts: [{ id: 1, title: 'Test Post' }],
+      users: { 10: { id: 10, username: 'testuser' } },
+      sites: { main: { site: 'main', name: 'Main' } },
+    }),
+    siteInfoToEntity: jest.fn().mockReturnValue({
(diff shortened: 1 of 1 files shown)


Aivean (approved):

backend/test/api/FeedController.test.ts
```diff
@@ -0,0 +1,484 @@
+import FeedController from '../../src/api/FeedController'
+import { FeedSorting } from '../../src/api/types/entities/common'
+
+// --- Mock factories ---
+
+function createMockEnricher() {
```
Aivean, line 6: I wonder if mock methods should be extracted into a shared place (if we have more backend tests in the future).

(also, this is a test for notifications, Aivean2, can you respond?)

Aivean: @Aivean2, can you see that comment?

Aivean2: Yes, notifications are working! 🎉

Re: shared mocks — agreed, once we add more controller tests (PostController, UserController, etc.), the mock factories (createMockEnricher, createMockDb, createMockUserStorage) should move to a shared test helper like `test/helpers/mocks.ts`. For now with a single test file it would be premature, but I will keep this in mind for the next PR.

PR #512, opened 2026-03-08
Encrypted Mailboxes v2

Aivean:
## What changed vs previous implementation

Previous version derived a stable RSA-512 keypair from the mailbox password on the client and embedded encrypted payload directly into post/comment HTML.

This PR replaces that flow with a clean v2:
- resets legacy mailbox public keys and adds server-side mail storage in a new [`mails` table](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/migrations/20260307000000-mails-v2.js#L17)
- stores only `<mail id="...">` in content; ciphertext is no longer embedded in rendered HTML ([parser](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/src/parser/TheParser.ts#L626), [binding on save](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/src/managers/PostManager.ts#L113))
- adds batched mail fetch: everyone can get metadata, but encrypted payload is returned only to sender/recipient ([API](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/src/api/MailController.ts#L95), [auth gating](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/src/managers/MailManager.ts#L49), [frontend batch helper](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/API/MailAPIHelper.ts#L6), [rendered component](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Components/Mail/MailComponent.tsx#L44))
- stores both recipient and sender copies, so sent mail remains readable by the sender as well ([compose flow](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Components/SecretMailbox.tsx#L70), [storage](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/src/db/repositories/MailRepository.ts#L11))
- restores the mailbox UI in profile settings and comment compose flow ([settings](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Components/UserProfileSettings.tsx#L360), [editor button](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Components/CreateCommentComponent.tsx#L778), [content integration](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Components/ContentComponent.tsx#L136))

## Crypto review notes

Goal: keep the old UX property "password -> stable keypair", but switch to modern primitives and let the backend gate access to encrypted payloads.

Flow:
1. Mailbox creation: password -> deterministic long-term X25519 keypair via [`scrypt-js` + `tweetnacl`](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Utils/mailCrypto.ts#L171); client saves the public key with `keyAlg = x25519-scrypt-v1` ([settings flow](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Components/UserProfileSettings.tsx#L399)); backend validates the format on save ([validation](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/src/api/UserController.ts#L563)).
2. Encryption: sender fetches the recipient public key ([lookup endpoint](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/src/api/PostController.ts#L673)), generates an ephemeral X25519 keypair and shared secret, derives an AES-256-GCM key with browser Web Crypto HKDF-SHA-256, and encrypts locally ([implementation](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Utils/mailCrypto.ts#L184), [compose path](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Components/SecretMailbox.tsx#L70)).
3. Storage: backend stores JSON envelopes (`to_payload` / `from_payload`) in `mails` and never receives plaintext ([migration](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/migrations/20260307000000-mails-v2.js#L29), [repository](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/src/db/repositories/MailRepository.ts#L11), [create API](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/src/api/MailController.ts#L71)).
4. Decryption: batch API returns payload only to an authorized sender/recipient ([payload selection](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/backend/src/managers/MailManager.ts#L49)); client re-derives the private key from password and decrypts locally, with in-memory keypair caching for the session ([crypto helper](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Utils/mailCrypto.ts#L209), [mail UI](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Components/Mail/MailComponent.tsx#L44), [password flow](https://github.com/spaceshelter/orbitar/blob/be47931adf894fc1830d725cb916139a42c21ca6/frontend/src/Components/SecretMailbox.tsx#L163)).

Libraries/primitives used:
- `scrypt-js` for password KDF
- `tweetnacl` for X25519 key agreement
- browser `Web Crypto API` for `HKDF-SHA-256` and `AES-256-GCM`


Aivean:
@Aivean2 please review

Aivean2 (commented):
## Review: Encrypted Mailboxes v2

Clean v1→v2 migration. The crypto flow is solid — X25519 + HKDF-SHA256 + AES-256-GCM is exactly the right primitive stack, and the architecture change (ciphertext in DB, not in HTML) solves the key rotation/size problems of v1 in one move.

### Crypto ✅

- `scrypt(password, "orbitar-mailbox-v2:{userId}")` → X25519 seed — good. User-specific salt prevents rainbow tables and means same password on two accounts produces different keys.
- Ephemeral X25519 keypair per message → forward secrecy for the encryption key. If the recipient's long-term key leaks, an attacker still needs the ephemeral private key (which only lived in memory during encryption).
- HKDF-SHA256 with domain separator `"orbitar-shifrovki-v2"` → AES-256-GCM with random 96-bit IV — textbook ECIES-style construction.
- `tweetnacl` 1.0.3 (audited) replaces `@daotl/cryptico` — major improvement.
- Backend never sees plaintext — only stores opaque JSON envelopes.

### Backend ✅

- `mails` table with separate `to_payload` / `from_payload` columns — clean dual-envelope storage.
- `syncPostBindings`/`syncCommentBindings` with `FOR UPDATE` and atomicity check (claimable.length !== mailIds.length → throw) — prevents orphaning and double-binding.
- Auth gating in `MailManager.mapMail`: recipient sees `to_payload`, sender sees `from_payload`, everyone else sees metadata only — correct.
- Server-side validation of public key format in `UserController` (`X25519_PUBLIC_KEY_REGEX`, `MAILBOX_KEY_ALG` check) — prevents garbage key registration.
- Rate limiting: 30/min for create, 120/min for batch get — reasonable.
- Migration resets all existing RSA keys (`UPDATE users SET public_key = ''`) — clean break as designed.

### Frontend ✅

- `BatchedCache` with 200ms debounce and 128 batch size for mail fetches — nice for pages with many `<mail>` tags.
- In-memory keypair caching per userId — good UX (enter password once, decrypt all mails in session).
- `MailComponent` as a proper React component replacing the old DOM-manipulation approach — much cleaner lifecycle.
- `SecretMailEncoderForm` now creates the mail server-side and returns `<mail id="...">` tag — no more base64 blob in HTML.

### Issues / Questions

1. **`from_payload` column is nullable (`TEXT DEFAULT NULL`) but `to_payload` is `TEXT NOT NULL`.** This is intentional (sender copy is optional if sender has no mailbox), but consider: could an attacker determine whether the sender has a mailbox by checking if `from_payload` is null? The batch API already exposes `canDecrypt` for the sender role which leaks this same info, so probably fine — just noting the information flow.

2. **`getUserIdByPublicKey` — no index on `(public_key, public_key_alg)`.** This is used during `createMail` when the sender provides a `toPublicKey` instead of `toUserId`. If many users have mailboxes, this becomes a full table scan. Consider adding a composite index, or require the frontend to always resolve `toUserId` before creating mail (which it currently does via `getPublicKeyByUsername`).

3. **Scrypt params: `N=16384, r=8, p=1`.** This is the same as v1 and provides ~100ms key derivation on modern hardware. Adequate for now but on the lower end — OWASP recommends N=32768+ for 2025+. Not blocking.

4. **`TheParser.VERSION` bump from 2 to 3** — this triggers re-parsing of all existing content on access. Is that the desired behavior? Old v1 `<mail secret="...">` tags will now fail the `parseMail` check (no `id` attribute) and get escaped as `&lt;mail ...&gt;`. This is the intended clean break, but worth being aware that old encrypted messages in existing posts become visible as raw HTML tags until the post is edited. Users will see the raw `<mail secret="...">` markup.

5. **`MailRepository.getMailsByIds` filters with `AND (m.post_id IS NOT NULL OR m.comment_id IS NOT NULL)`.** This means unbound mails (created but not yet attached to a post/comment) are invisible to the batch API. There's a race window: if `createMail` succeeds but the post/comment create fails, the mail exists but is unreachable. The frontend calls `createMail` in the encoder form before the parent post is submitted, so this window is real. The mail IDs would be "leaked" — created in DB but never fetchable. Consider either: (a) a cleanup job for unbound mails older than N hours, or (b) documenting this as acceptable.

6. **Minor: duplicate type definitions.** `MailInfo` (manager type) and `MailBatchEntity` (API entity) are identical. Could re-export one as the other.

### Verdict

The crypto is correct, the architecture is a significant improvement over v1, and the code is well-structured. The issues above are non-blocking — the PR is ready to ship as-is, with the index (#2) and orphan cleanup (#5) as follow-up items.

Aivean2:
Reviewed the two new commits (`10447db`, `d7b28f3`). Good changes:

**Addressed from previous review:**
- ✅ #3 — scrypt N raised to 32768, now matches OWASP 2025+ recommendation
- ✅ #4 — parser version reverted to 2, old `<mail secret="...">` tags won't break
- ✅ #5 — binding system (`syncPostBindings`/`syncCommentBindings`) removed entirely, so orphan mail race is gone

**New: public mail mode**
- `to_user_id` now nullable — enables broadcast mails with no specific recipient
- `mapMail` correctly handles public vs addressed: public mails return `to_payload` to everyone, `from_payload` to sender only
- `getUserIdByPublicKey` removed — cleaner; frontend now passes `toUserId` directly when known
- `.or('toUserId', 'toPublicKey')` → `.xor(...)` — correct, prevents sending both
- Clipboard copy + toast on encode — nice UX touch

**One observation on public mail semantics:**

When `to_user_id IS NULL` (public mail), `mapMail` returns `to_payload` to *any* viewer. This means `to_payload` for public mails acts as the "publicly decryptable" envelope — the sender encrypts to a public key that everyone can derive. Worth a brief comment in code or PR description clarifying the expected UX: who provides the decrypt password for public mails? If it's broadcast-encrypted to a shared/published key, that's fine. If it's meant to be plaintext-visible, the naming `to_payload` is slightly misleading (it's really `public_payload`).

Otherwise, clean follow-up. 👍

Aivean:
I decided to reimplement it with a slightly different UX.

Aivean2:
Got it. Looking forward to the new version.

Aivean:
superseded by https://github.com/spaceshelter/orbitar/pull/513

PR #513, opened 2026-03-09
Whole-body encrypted content v2.1

Aivean:
## Summary
- adds mailbox v2 foundation: `public_key_alg`, deterministic mailbox keypair derivation, global `CryptoSessionStore`, shared unlock UX
- adds generic encrypted payload storage (`encrypted_payloads`, `encrypted_payload_keys`) bound to `content_source` and cached on `posts` / `comments`
- makes whole-body encrypted comments first-class; encrypted post plumbing is included in the same model
- adds client-side preview, decrypt, edit, and history support for encrypted content
- encrypted replies still notify the parent author with a generic `🔒 Шифровка` preview and do not bump public feeds

## Different From The Original Implementation
- replaces the old RSA-512 / inline-mail approach with first-class whole-body encrypted content
- stores one ciphertext plus per-user key wraps instead of duplicating ciphertext per addressee
- content rows now keep `encryptedPayloadId`; ciphertext is stored separately and fetched through a dedicated batch API
- mailbox unlock is session-wide and reused across page render, history, and edit flows instead of being a one-off per content item
- encrypted content is now attached to `content_source`, so edit history and future private-thread work can build on the same model

## Inboxes Follow-Up
This PR is intentionally the direct encrypted-content foundation, not the full private-inbox implementation.

Reusable pieces already in place:
- global `CryptoSessionStore`
- generic `encrypted_payloads` / `encrypted_payload_keys`
- payload ids on `content_source`, `posts`, and `comments`
- batched payload fetch and locked renderer

Planned direction for private/group inboxes:
- introduce a thread key per inbox
- wrap that thread key once per participant
- encrypt each post/comment with the thread key instead of creating per-message wraps for every participant
- reuse the same unlock/session/rendering path from this PR

## Crypto Approach (For Audit)
High level:
- password deterministically derives the user's long-term mailbox keypair
- each encrypted body gets one random content key and one ciphertext
- that content key is wrapped separately for sender and recipient
- backend stores only ciphertext + wraps; plaintext never leaves the client

Details:
- password -> mailbox keypair: `scrypt-js` with `N=32768`, `r=8`, `p=1`, `dkLen=32`; output is used as the Curve25519 secret via `tweetnacl.box.keyPair.fromSecretKey`
- only `public_key` + `public_key_alg = x25519-scrypt-v1` are stored server-side
- plaintext is encrypted once with Web Crypto `AES-256-GCM` using a random 32-byte content key and random 12-byte IV
- for each allowed decryptor, the client generates an ephemeral Curve25519 keypair, computes ECDH with `tweetnacl.scalarMult`, derives a wrapping key via Web Crypto `HKDF-SHA-256`, and wraps the content key with `AES-GCM`
- `/encrypted-payload/get` returns the payload body only when the current user has a matching wrap
- decryption is client-side only: derive mailbox key from password, unwrap content key, decrypt ciphertext, render with the restricted `lite-v1` client parser
- encrypted content does not support implicit parser-based mentions; encrypted replies use generic notification/push previews (`🔒 Шифровка`)
- encrypted comments do not bump public feeds

Key code paths:
- mailbox KDF + per-content encrypt/decrypt: [frontend/src/Utils/mailCrypto.ts](https://github.com/spaceshelter/orbitar/blob/9da38b354dba856618d5878862d5e34b56349dfa/frontend/src/Utils/mailCrypto.ts#L58)
- session unlock cache: [frontend/src/AppState/CryptoSessionStore.ts](https://github.com/spaceshelter/orbitar/blob/9da38b354dba856618d5878862d5e34b56349dfa/frontend/src/AppState/CryptoSessionStore.ts#L5)
- editor encryption flow: [frontend/src/Components/CreateCommentComponent.tsx](https://github.com/spaceshelter/orbitar/blob/9da38b354dba856618d5878862d5e34b56349dfa/frontend/src/Components/CreateCommentComponent.tsx#L188)
- encrypted payload validation and comment create path: [backend/src/api/PostController.ts](https://github.com/spaceshelter/orbitar/blob/9da38b354dba856618d5878862d5e34b56349dfa/backend/src/api/PostController.ts#L45)
- encrypted payload storage: [backend/src/db/repositories/EncryptedPayloadRepository.ts](https://github.com/spaceshelter/orbitar/blob/9da38b354dba856618d5878862d5e34b56349dfa/backend/src/db/repositories/EncryptedPayloadRepository.ts#L12)
- encrypted payload batch fetch: [backend/src/api/EncryptedPayloadController.ts](https://github.com/spaceshelter/orbitar/blob/9da38b354dba856618d5878862d5e34b56349dfa/backend/src/api/EncryptedPayloadController.ts#L19)
- locked render / decrypt / auto-unlock reuse: [frontend/src/Components/EncryptedContentComponent.tsx](https://github.com/spaceshelter/orbitar/blob/9da38b354dba856618d5878862d5e34b56349dfa/frontend/src/Components/EncryptedContentComponent.tsx#L22)

### DIFF
--- a/backend/migrations/20260308000000-public-key-alg.js
+++ b/backend/migrations/20260308000000-public-key-alg.js
@@ -1,0 +1,36 @@
+'use strict'
+
+var dbm
+var type
+var seed
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate
+  type = dbm.dataType
+  seed = seedLink
+}
+
+exports.up = async function (db) {
+  await db.runSql(`
+        ALTER TABLE users ADD COLUMN public_key_alg VARCHAR(32) NOT NULL DEFAULT ''
+    `)
+
+  await db.runSql(`
+        UPDATE users
+        SET public_key = '', public_key_alg = ''
+    `)
+}
+
+exports.down = async function (db) {
+  await db.runSql(`
+        ALTER TABLE users DROP COLUMN public_key_alg
+    `)
+}
+
+exports._meta = {
+  version: 1,
+}
--- a/backend/migrations/20260308000001-encrypted-payloads.js
+++ b/backend/migrations/20260308000001-encrypted-payloads.js
@@ -1,0 +1,100 @@
+'use strict'
+
+var dbm
+var type
+var seed
+
+/**
+ * We receive the dbmigrate dependency from dbmigrate initially.
+ * This enables us to not have to rely on NODE_PATH.
+ */
+exports.setup = function (options, seedLink) {
+  dbm = options.dbmigrate
+  type = dbm.dataType
+  seed = seedLink
+}
+
+exports.up = async function (db) {
+  await db.runSql(`
+        CREATE TABLE encrypted_payloads (
+            encrypted_payload_id INT NOT NULL AUTO_INCREMENT,
+            v TINYINT NOT NULL,
+            parser_profile VARCHAR(32) NOT NULL DEFAULT 'lite-v1',
+            ciphertext MEDIUMTEXT NOT NULL,
+            iv VARCHAR(32) NOT NULL,
+            created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            PRIMARY KEY (encrypted_payload_id)
+        )
+    `)
+
+  await db.runSql(`
+        CREATE TABLE encrypted_payload_keys (
+            encrypted_payload_key_id INT NOT NULL AUTO_INCREMENT,
+            encrypted_payload_id INT NOT NULL,
+            user_id INT NOT NULL,
+            role VARCHAR(16) NOT NULL,
+            public_key VARCHAR(128) NOT NULL,
+            public_key_alg VARCHAR(32) NOT NULL,
+            ephemeral_public_key VARCHAR(128) NOT NULL,
+            iv VARCHAR(32) NOT NULL,
+            encrypted_key VARCHAR(128) NOT NULL,
+            created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            PRIMARY KEY (encrypted_payload_key_id),
+            UNIQUE KEY uq_encrypted_payload_user_role (encrypted_payload_id, user_id, role),
+            KEY idx_encrypted_payload_id (encrypted_payload_id),
+            KEY idx_user_id (user_id),
+            CONSTRAINT fk_encrypted_payload_keys_payload FOREIGN KEY (encrypted_payload_id) REFERENCES encrypted_payloads (encrypted_payload_id) ON DELETE CASCADE ON UPDATE CASCADE,
+            CONSTRAINT fk_encrypted_payload_keys_user FOREIGN KEY (user_id) REFERENCES users (user_id) ON DELETE CASCADE ON UPDATE CASCADE
+        )
+    `)
+
+  await db.runSql(`
+        ALTER TABLE content_source
+            ADD COLUMN encrypted_payload_id INT NULL,
+            ADD CONSTRAINT fk_content_source_encrypted_payload FOREIGN KEY (encrypted_payload_id) REFERENCES encrypted_payloads (encrypted_payload_id) ON DELETE RESTRICT ON UPDATE RESTRICT
+    `)
+
+  await db.runSql(`
+        ALTER TABLE posts
+            ADD COLUMN encrypted_payload_id INT NULL,
+            ADD CONSTRAINT fk_posts_encrypted_payload FOREIGN KEY (encrypted_payload_id) REFERENCES encrypted_payloads (encrypted_payload_id) ON DELETE RESTRICT ON UPDATE RESTRICT
+    `)
+
+  await db.runSql(`
+        ALTER TABLE comments
+            ADD COLUMN encrypted_payload_id INT NULL,
+            ADD CONSTRAINT fk_comments_encrypted_payload FOREIGN KEY (encrypted_payload_id) REFERENCES encrypted_payloads (encrypted_payload_id) ON DELETE RESTRICT ON UPDATE RESTRICT
+    `)
+}
+
+exports.down = async function (db) {
+  await db.runSql(`
+        ALTER TABLE comments
+            DROP FOREIGN KEY fk_comments_encrypted_payload,
+            DROP COLUMN encrypted_payload_id
+    `)
+
+  await db.runSql(`
+        ALTER TABLE posts
+            DROP FOREIGN KEY fk_posts_encrypted_payload,
+            DROP COLUMN encrypted_payload_id
+    `)
+
+  await db.runSql(`
+        ALTER TABLE content_source
+            DROP FOREIGN KEY fk_content_source_encrypted_payload,
+            DROP COLUMN encrypted_payload_id
+    `)
+
+  await db.runSql(`
+        DROP TABLE encrypted_payload_keys
+    `)
+
+  await db.runSql(`
+        DROP TABLE encrypted_payloads
+    `)
+}
+
+exports._meta = {
+  version: 1,
+}
--- a/backend/src/api/AuthController.ts
+++ b/backend/src/api/AuthController.ts
@@ -145,6 +145,7 @@ export default class AuthController {
         return response.error('wrong-credentials', 'Wrong username or password')
       }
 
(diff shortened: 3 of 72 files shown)


Aivean2 (commented):
## Review: Whole-body encrypted content v2.1

Massive architectural upgrade from v2 (per-mail envelopes) to a general-purpose encrypted content model. The shift from `mails` table with dual payloads to `encrypted_payloads` + `encrypted_payload_keys` is a significantly better foundation — one ciphertext per content, per-user key wraps, and it naturally extends to thread keys / group inboxes later.

Reviewed all 72 files across 6 commits. Here's the breakdown:

### Crypto ✅

- `scrypt(password, MAILBOX_SALT, N=32768, r=8, p=1)` → 32-byte seed → `nacl.box.keyPair.fromSecretKey` — clean deterministic X25519 derivation.
- **Per-content encryption:** random 32-byte AES-256-GCM key + random 12-byte IV. One ciphertext. No duplicate per recipient.
- **Key wrapping:** ephemeral X25519 keypair per recipient → `nacl.scalarMult` (ECDH) → HKDF-SHA-256 → AES-GCM wrap of content key. Correct ECIES construction.
- **HKDF domain separator:** `orbitar-encrypted-content-v1` (changed from v2's `orbitar-shifrovki-v2`) — good, since this is a different protocol.
- **tweetnacl 1.0.3** properly resolved as root dependency, with legacy 0.14.5 scoped under bcrypt-pbkdf/sshpk.
- `isValidMailboxPublicKey` — base64url regex `^[A-Za-z0-9_-]{43}$` matches exactly 32 bytes (43 chars = ceil(32*4/3) without padding). Correct.
- **Backend never sees plaintext** — only stores opaque ciphertext + key wraps.

### Backend ✅

**Migrations:**
- `20260308000000-public-key-alg`: adds `public_key_alg` column, resets all existing keys. Clean break from RSA-512 era.
- `20260308000001-encrypted-payloads`: creates `encrypted_payloads` (ciphertext + IV + parser profile) and `encrypted_payload_keys` (per-user wraps). FK constraints with `ON DELETE CASCADE` for keys, `ON DELETE RESTRICT` for content references. Correct — you don't want orphaned ciphertext, but you also don't want content deletion to silently destroy payloads.

**Repository layer:**
- `EncryptedPayloadRepository.createEncryptedPayload` — creates payload + wraps within an existing transaction (receives `DBConnection`). This means payload creation is atomic with post/comment creation. Clean.
- `getEncryptedPayloadsByIds` — LEFT JOIN on `encrypted_payload_keys` filtered by `userId`. Correctly returns metadata for everyone but ciphertext only to authorized users (via `key_user_id` check in manager).

**Manager layer:**
- `EncryptedPayloadManager.getEncryptedPayloadsByIds` — maps raw DB rows to API entities. `canDecrypt: !!row.key_user_id` — correct: if the LEFT JOIN found no matching key row, user can't decrypt.
- `PostManager` — properly short-circuits parser, language detection, and content storage for encrypted payloads (passes empty strings). Feed bump suppressed for encrypted comments.
- `NotificationManager` — `'🔒 Шифровка'` placeholder for both in-app and push notifications. Good — no plaintext leakage through notification channels.

**Controller layer:**
- `requirePlaintextOrEncrypted` custom Joi validator — XOR between content and encryptedPayload. Prevents both or neither.
- Logging uses `'<encrypted>'` placeholder instead of ciphertext. Good OPSEC.
- `UserManager.savePublicKey` — validates key format server-side with `isValidMailboxPublicKey`. Error surface correctly changed to 400 from 500.

**Test:**
- `EncryptedPayloadController.test.ts` — covers auth requirement, ID dedup, error handling. Adequate for the controller layer.

### Frontend ✅

**`mailCrypto.ts`:**
- Clean separation of concerns: KDF, encrypt, decrypt, validation, base64url encoding.
- `encryptContentForUsers` — generates one content key, wraps it per-recipient. Each wrap gets its own ephemeral keypair. This is the correct per-recipient forward secrecy pattern.
- `decryptEncryptedPayload` — unwraps content key, then decrypts ciphertext. Error wrapping prevents leaking crypto internals.

**`CryptoSessionStore`:**
- MobX-observable store with session-wide key caching. `syncCurrentUser` clears on user switch. `clear` on logout. No persistence — keys only live in memory for the browser session.
- `unlockMailbox` — derives key, validates against public key, caches on success. Race condition handling via `runInAction` batching.

**`EncryptedContentComponent`:**
- Fetches payload via `BatchedCache` (200ms debounce, 128 batch size). Auto-decrypts when mailbox is already unlocked (`mailboxMatchesPayload` check).
- Click-to-unlock for "can decrypt but not yet unlocked" state. Clean locked/disabled/error visual states.
- `observer` wrapper enables reactive auto-decrypt when `CryptoSessionStore` gets the key pair.

**`encryptedContentParser.ts` (lite-v1 client parser):**
- Whitelist-based HTML sanitizer: `b, i, u, strike, br, a, img, video, blockquote, pre`. Everything else gets escaped.
- `sanitizeLink` / `sanitizeMediaUrl` — protocol whitelist (http/https/mailto for links, http/https for media). `javascript:` blocked.
- This is the right approach for decrypted content — the server parser doesn't see the plaintext, so you need a restricted client-side parser.

**`CreateCommentComponent` changes:**
- Encryption toggle button with key resolution. `resolveEncryptionKeys` fetches both sender and recipient public keys.
- Disabled formatting buttons (irony, expand, spoiler, poll) when encrypted — correct, since the lite-v1 parser doesn't support them.
- Gallery wrapping disabled for encrypted mode.
- Preview uses client-side `renderEncryptedContentHtml` instead of server-side preview.

**Edit flow:**
- Both `CommentComponent` and `PostComponent` handle encrypted edit: fetch payload, decrypt source, populate editor with `initialEncrypted` flag. Re-encrypts on save.
- History view properly shows `EncryptedContentComponent` for encrypted entries, disables diff view for encrypted↔encrypted comparisons.

**Old code removal:**
- `SecretMailEncoderForm`, `SecretMailDecoderForm`, `updateMailbox`, `updateMail`, RSA key cache, `@daotl/cryptico` usage — all cleanly removed from `ContentComponent.tsx` and `SecretMailbox.tsx`. The only remaining export from `SecretMailbox.tsx` is `SecretMailKeyGeneratorForm` (now uses `deriveMailboxKeyPair`) and the new `MailboxUnlockForm` / `requestMailboxUnlock`.

### Issues / Observations

1. **🟡 `UserManager.savePublicKey` missing `await`:** Both the validation path and the clear path call `this.userRepository.savePublicKey(...)` which returns a Promise, but `savePublicKey` in UserManager is not `async` and doesn't `await` the result. The cache is cleared synchronously after a potentially-unresolved promise. If `savePublicKey` fails, the cache is already cleared but the DB still has the old key. This was a pre-existing issue (same pattern in the old code), but worth fixing now that validation logic was added.

2. **🟡 Global salt for scrypt KDF:** `MAILBOX_SALT = 'f8psK3rQ58K#J#j95@UXFt94RTyH!q9R'` is a global constant, not per-user. v2 had per-user salt (`orbitar-mailbox-v2:{userId}`). With a global salt, two users with the same mailbox password will derive the same X25519 keypair and public key. This is detectable from the `users` table (identical `public_key` values). Whether this matters in practice depends on how many users share passwords, but it's a minor information leak. Not blocking.

3. **🟡 Encrypted payload rate limiting:** The batch fetch endpoint (`/encrypted-payload/get`) uses `oauth()` but doesn't have its own `rateLimit()`. The PR description from v2 mentioned 120/min for batch get — worth adding a rate limiter to `EncryptedPayloadController` to prevent enumeration.

4. **ℹ️ Empty HKDF salt:** `EMPTY_SALT = new Uint8Array([])` is valid per RFC 5869 (HKDF treats empty salt as `HashLen` zeros), but a non-empty salt (e.g., the payload ID) would strengthen key separation. Not a bug, just a note for the planned thread-key extension.

5. **ℹ️ `encryptedPayloadId` on `posts`/`comments` tables is denormalized** — it's also on `content_source`. This is intentional for read performance (avoids joining through `content_source` on feed queries), but worth a brief comment in the migration or schema docs to explain the triple-store pattern.

6. **ℹ️ Reply notifications for encrypted comments:** Commit 5 (`76fef4d1`) correctly sends reply notifications even for encrypted comments, while commit 6 (`9da38b35`) suppresses feed bumps. Good separation — the parent author should know they got a reply even if the content preview is just `🔒 Шифровка`.

### Verdict

The architecture is significantly improved from v2 — the generic payload+wraps model is cleaner, extensible (thread keys, group inboxes, encrypted posts), and removes the v1 legacy entirely. Crypto is correct. The code is well-structured with clean separation between crypto, state management, and UI. Ship as-is, with the missing `await` (#1) and rate limit (#3) as follow-up items.

PR #514, opened 2026-03-17
Online users count

LazyKarlson:
LazyKarlson and Claude)

Each active user hits /status every 30s — we use this as a heartbeat.
On each request, backend updates a Redis Sorted Set (ZADD online_users <timestamp> <userId>) and counts users seen in the last 90 seconds (ZCOUNT). The result is returned in the /status response and displayed in the topbar's right section with a green dot.

Changes:

StatusController — ZADD + ZCOUNT on every /status request
StatusResponse — new onlineCount field
AppState — observable onlineCount + setOnlineCount action
Topbar — OnlineCount component in .right

### DIFF
--- a/backend/src/api/StatusController.ts
+++ b/backend/src/api/StatusController.ts
@@ -1,5 +1,4 @@
 import { Router } from 'express'
-import { RedisClientType } from 'redis'
 import { Logger } from 'winston'
 
 import SiteManager from '../managers/SiteManager'
@@ -9,16 +8,12 @@ import { OAuth2MiddlewareGenerator } from './OAuth2Middleware'
 import { StatusRequest, StatusResponse } from './types/requests/Status'
 import { Enricher } from './utils/Enricher'
 
-const ONLINE_USERS_KEY = 'online_users'
-const ONLINE_WINDOW_SECONDS = 90 // 3 × 30s polling interval
-
 export default class StatusController {
   public readonly router = Router()
   private readonly siteManager: SiteManager
   private readonly userManager: UserManager
   private readonly logger: Logger
   private readonly enricher: Enricher
-  private readonly redis: RedisClientType
 
   constructor(
     enricher: Enricher,
@@ -26,13 +21,11 @@ export default class StatusController {
     userManager: UserManager,
     oauth: OAuth2MiddlewareGenerator,
     logger: Logger,
-    redis: RedisClientType,
   ) {
     this.siteManager = siteManager
     this.userManager = userManager
     this.enricher = enricher
     this.logger = logger
-    this.redis = redis
     this.router.post('/status', oauth('читать статус'), (req, res) => this.status(req, res))
   }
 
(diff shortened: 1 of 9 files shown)


Aivean (changes requested):
## Review

Good approach — Redis sorted sets are the right tool here, and piggy-backing on `/status` as heartbeat is smart. A few things to fix:

### Bugs / Resilience

1. **Sorted set grows unbounded** — entries are never pruned. Add `zRemRangeByScore(ONLINE_USERS_KEY, '-inf', now - ONLINE_WINDOW_SECONDS)` on each request (it's cheap).

2. **Redis failure breaks `/status`** — the ZADD/ZCOUNT are inside the same try/catch that returns 500. If Redis hiccups, the entire polling loop dies (no notifications, no unread counts). Wrap online tracking in its own try/catch and degrade gracefully (`onlineCount = 0`).

### Architecture

3. **Redis calls in the controller** — this breaks the `Controller → Manager → Repository` pattern used everywhere else. The online tracking logic should live in a manager (e.g. `UserManager` or a new `StatusManager`), not directly in the controller.

### Frontend

4. **SCSS formatting** — `.onlineCount` properties aren't indented inside the nesting (rest of file uses 4-space indent). The `&::before` closing brace is also misaligned. Trailing space on `align-items` line, trailing blank lines at EOF.

5. **Hardcoded `#4caf50`** — rest of the theme uses CSS variables. This won't adapt to theme changes.

6. **Shows "0" on load** — `OnlineCount` renders before first `/status` response. Hide when count is 0: `if (!onlineCount) return null`.

Aivean:
One more optimization: if you prune stale entries on each request (point 1), then every member in the set is within the window. You can use `ZCARD` (O(1)) instead of `ZCOUNT` (O(log N)):

```ts
await this.redis.zRemRangeByScore(ONLINE_USERS_KEY, '-inf', now - ONLINE_WINDOW_SECONDS)
await this.redis.zAdd(ONLINE_USERS_KEY, { score: now, value: String(userId) })
const onlineCount = await this.redis.zCard(ONLINE_USERS_KEY)
```

Note: prune before add so the count is accurate even if the current user's previous entry was stale.

LazyKarlson:
UserManager.trackOnline — replaced 3 sequential Redis calls with a single pipeline (multi().exec())

OnlineCount — hidden when onlineCount === 0 (if (!onlineCount) return null)

Topbar.module.scss — replaced CSS green dot (::before) with i-user font icon; icon color var(--primary), size 20px
Screenshot_2026-03-18_12-15-40

Aivean:
@Aivean2, please verify the fixes

@LazyKarlson, still need to work on the design

Aivean2:
Verified all 6 review points + ZCARD optimization:

| # | Issue | Status |
|---|-------|--------|
| 1 | Unbounded sorted set | ✅ `zRemRangeByScore` prunes on every call |
| 2 | Redis failure breaks `/status` | ✅ Separate try/catch, degrades to `onlineCount = 0` |
| 3 | Redis calls in controller | ✅ Moved to `UserManager.trackOnline()` |
| 4 | SCSS formatting | ✅ Properly indented |
| 5 | Hardcoded `#4caf50` | ✅ `var(--fg)` + `var(--primary)` |
| 6 | Shows "0" on load | ✅ `if (!onlineCount) return null` |
| 7 | ZCARD optimization | ✅ Pipeline: prune → add → ZCARD, correct order |

Code-level the fixes look good. Pipeline is the first `multi()` usage in the codebase — `results[2] as number` is fine with node-redis v4 (direct values, not tuples).

PR #515, opened 2026-03-18
colored post author username in comments

LazyKarlson:
The comment author's username is highlighted in var(--primary) when they are the post author; otherwise it stays var(--fgSoftest).

Implementation:

PostPage — passes postAuthorUsername={post.author.username} to the root CommentComponent
CommentComponent — accepts postAuthorUsername, computes isPostAuthor by comparing it to the comment author, passes both down recursively to the full comment tree
SignatureComponent — accepts isPostAuthor, applies classNames(styles.username, isPostAuthor && styles.postAuthor)
SignatureComponent.module.scss — .postAuthor modifier overrides color to var(--primary) on top of the base .username styles
2026-03-18 12 50 17 orbitar local af4ad8960845

### DIFF
--- a/frontend/src/Components/CommentComponent.tsx
+++ b/frontend/src/Components/CommentComponent.tsx
@@ -32,6 +32,7 @@ interface CommentProps {
   unreadOnly?: boolean
   hideRating?: boolean
   currentUsername?: string
+  postAuthorUsername?: string
 }
 
 export default function CommentComponent(props: CommentProps) {
@@ -133,6 +134,7 @@ export default function CommentComponent(props: CommentProps) {
           postLinkIsNew={props.unreadOnly}
           date={created}
           editFlag={editFlag}
+          isPostAuthor={!!props.postAuthorUsername && author.username === props.postAuthorUsername}
         />
         {editingText === false ? (
           showHistory ? (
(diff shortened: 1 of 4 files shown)


Aivean:
@Aivean2, please review

Aivean2 (commented):
Clean, well-scoped PR. The implementation follows existing patterns nicely — `postAuthorUsername` prop drilling mirrors how `currentUsername` is already threaded through.

A few observations:

1. **Other call sites:** `CommentComponent` is also used in `SearchPage`, `InternalLinkExpandComponent`, and `UserProfileComments` — none of which pass `postAuthorUsername`. Since the prop is optional, this degrades gracefully (no highlighting outside PostPage). Worth noting for future — e.g. `UserProfileComments` could also benefit if post author info is available.

2. **Color overlap with hover:** The post author's default color is `var(--primary)`, which is the same as the *hover* color for regular usernames (line 31 of the existing `.username` rule). So a post-author name looks "always hovered" — not necessarily a problem, but something to be aware of visually. The `--primaryHover` on actual hover provides differentiation.

3. Code-wise: `classNames` is already a project dependency (used across 10+ files), the SCSS is properly nested inside `.signature`, and the `&, &:before` selector correctly covers both the username text and the `@` prefix. All clean.

CI passes. Looks good to me overall — leaving to @Aivean for final call.

PR #516, opened 2026-03-19
Display Orbitar Award trophy badges on user profile

LazyKarlson:
Users who won the community "Премия Орбитара" (Orbitar Award) polls now have 🏆 trophy badges displayed next to their username on the profile page. Each badge links to the corresponding award announcement post and shows the award number in a tooltip on hover.

How it works
The Orbitor Award is determined by community polls created by a dedicated account with questions matching the pattern Премия Орбитара #N. The winner is the option (username) with the most votes in the completed poll.

Winner resolution:
`PollRepository.getOrbitorAwardPolls()` fetches the latest completed poll for each unique award question (deduplicates by `GROUP BY question / MAX(created_at)`).
`PostRepository.getOrbitorAwardPosts()` fetches award announcement posts by the same author, matching the same title pattern.
`UserManager.refreshOrbitorAwardWinners()` runs both queries in parallel (`Promise.all`), extracts the award number from each question/title via `/#(\d+)/` regex, determines the winner by `argmax(opt0…opt31)`, and correlates polls with posts by award number to attach a  `postId` to each win.

Caching:
The result is stored as a process-level in-memory `Map<username, {awardNum, postId?}[]>` — one shared cache for all users.
The cache is refreshed lazily: on any profile request, if more than 24 hours have passed since the last update, a refresh is triggered.
A boolean `orbitorAwardRefreshing` flag prevents concurrent refreshes under parallel requests.
The cache is lost on server restart and rebuilt on the first profile request.

API:
`GET /user/profile` response gains an optional field `orbitorAwardWins: {awardNum: number, postId?: number}[]`, omitted entirely when the user has no wins.

Frontend:
 `UserPage` renders one 🏆 badge per win next to the username.
If a matching post is found, the badge is a `<Link to="/p{postId}">` with `title="Лауреат Премии Орбитара #N"`.
If no post is found (fallback), it renders as a `<span>` with the same tooltip.

### DIFF
--- a/backend/src/managers/UserManager.ts
+++ b/backend/src/managers/UserManager.ts
@@ -7,6 +7,7 @@ import { FeedSorting } from '../api/types/entities/common'
 import { TrialProgressDebugInfo } from '../api/types/requests/UserProfile'
 import { config, SiteConfig } from '../config'
 import CommentRepository from '../db/repositories/CommentRepository'
+import PollRepository from '../db/repositories/PollRepository'
 import PostRepository from '../db/repositories/PostRepository'
 import UserCredentials from '../db/repositories/UserCredentials'
 import UserRepository from '../db/repositories/UserRepository'
@@ -53,6 +54,7 @@ export default class UserManager {
   private voteRepository: VoteRepository
   private commentRepository: CommentRepository
   private postRepository: PostRepository
+  private pollRepository: PollRepository
   private notificationManager: NotificationManager
   private webPushRepository: WebPushRepository
   private readonly redis: RedisClientType
@@ -61,6 +63,15 @@ export default class UserManager {
   private readonly logger: Logger
   private readonly mailLogger: Logger
   private readonly parser: TheParser
+
+  private static readonly ORBITOR_AWARD_AUTHOR_ID = 2154
+  private static readonly ORBITOR_AWARD_CACHE_TTL_MS = 24 * 60 * 60 * 1000 // 24h
+
+  private orbitorAwardWinnersCache: {
+    wins: Map<string, { awardNum: number; postId?: number }[]>
+    lastUpdatedTs: number
+  } = { wins: new Map(), lastUpdatedTs: 0 }
+  private orbitorAwardRefreshing = false
 
   private cachedNumActiveUsersThatCanVote = {
     expirationMs: 1000 * 60 * 60 /* 1 hour */,
@@ -91,12 +102,14 @@ export default class UserManager {
     redis: RedisClientType,
     siteConfig: SiteConfig,
     logger: Logger,
+    pollRepository: PollRepository,
   ) {
     this.credentialsRepository = credentialsRepository
     this.userRepository = userRepository
     this.voteRepository = voteRepository
     this.commentRepository = commentRepository
     this.postRepository = postRepository
+    this.pollRepository = pollRepository
     this.notificationManager = notificationManager
     this.webPushRepository = webPushRepository
     this.userCache = userCache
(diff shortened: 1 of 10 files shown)


LazyKarlson:
@Aivean2, please review

backend/src/managers/UserManager.ts
```diff
@@ -53,6 +54,7 @@ export default class UserManager {
   private voteRepository: VoteRepository
   private commentRepository: CommentRepository
   private postRepository: PostRepository
+  private pollRepository: PollRepository
   private notificationManager: NotificationManager
   private webPushRepository: WebPushRepository
   private readonly redis: RedisClientType
```
Aivean2, line 60: What is 2154? A comment like `// OrbitarPremium bot user ID` would help future readers.

backend/src/managers/UserManager.ts
```diff
@@ -107,6 +120,72 @@ export default class UserManager {
...
+    if (this.orbitorAwardRefreshing) {
+      return
+    }
+    this.orbitorAwardRefreshing = true
+    try {
+      const [polls, awardPosts] = await Promise.all([
+        this.pollRepository.getOrbitorAwardPolls(UserManager.ORBITOR_AWARD_AUTHOR_ID),
+        this.postRepository.getOrbitorAwardPosts(UserManager.ORBITOR_AWARD_AUTHOR_ID),
+      ])
+
```
Aivean2, line 134: If regex extraction fails, `awardNum` becomes `0` and you get a badge with tooltip "Лауреат Премии Орбитара #0". Safer to skip:

```suggestion
          const awardNum = parseInt(poll.question.match(/#(\d+)/)?.[1] ?? '', 10)
          if (isNaN(awardNum)) continue
          const entry: { awardNum: number; postId?: number } = { awardNum }
```

frontend/src/Pages/UserPage.module.scss
```diff
@@ -64,6 +64,14 @@
 
 .username {
   font-size: 40px;
+  display: flex;
+  align-items: center;
+  gap: 8px;
+}
+
```
Aivean2, line 71: `cursor: pointer` applies to both `<Link>` and `<span>` fallback, but the span is not clickable — misleading cursor. Maybe `cursor: default` on spans, or just drop cursor from the shared class and let `<Link>` handle it naturally.

backend/src/managers/UserManager.ts
```diff
@@ -107,6 +120,72 @@ export default class UserManager {
     this.parser = parser
```
Aivean2, line 121: Tie-breaking: if two options have equal votes, the first one wins (lower index). Probably fine for this use case, but worth a brief comment.

Aivean2 (commented):
Nice feature — displaying award badges on profile is a natural addition.

**Naming typo**: "Orbitor" → "Orbitar" throughout (method names, variables, type fields, CSS class, constants). The codebase consistently uses "Orbitar" (see `manifest.json`, `Mailer.ts`, `TranslationManager.ts`). Worth fixing before merge since `orbitorAwardWins` in the API response becomes a contract — renaming later is a breaking change.

A few more notes below.

backend/src/managers/UserManager.ts
LazyKarlson: may be we should store it in .env file?

Aivean2: Either works. A named constant with a comment (like `private static readonly ORBITAR_AWARD_AUTHOR_ID = 2154 // OrbitarPremium bot`) is the simpler option since this ID is tied to a specific platform account that won't change. `.env` makes more sense if you want it configurable per deployment, but for a single-instance platform it's probably overengineering. Up to you — main thing is that whoever reads the code next knows what 2154 means.

backend/src/managers/UserManager.ts
LazyKarlson: in this case the rest of code e.g. `wins.set(winner, arr)` will be skipped and badge will not be shown in profile

Aivean2: Right, that's the intent — if we can't parse the award number, better to silently skip than show a "#0" badge. In practice the SQL already filters `question LIKE 'Премия Орбитара #%'` so the regex should always match, but the `continue` is a safety net for unexpected data.

Aivean (changes requested):
**critical directional issue**:  this mechanism is tied to the winners of the specific poll. if we're implementing badges, they should be universal. the mechanism on how to award them should be general, and in this case controlled **externally** (e.g. by webhook that orbitar polls on poll completion)


additionally, even if we entertain this direction, running query every 24 hours is a weird decision. why not update the map on each poll completion?

Aivean2:
Thanks for the update. The rename (Orbitor → Orbitar), magic number comment, and cursor fix all look good.

One minor thing still open from my earlier review: the `awardNum: 0` fallback (line ~166 in UserManager.ts) — still uses `isNaN(awardNum) ? 0 : awardNum` instead of `continue`. In practice the SQL filter makes this unlikely to trigger, but a `#0` badge would be confusing if it did.

That said, the main blocker is @Aivean's [directional feedback](https://github.com/spaceshelter/orbitar/pull/516#pullrequestreview-3976361976) — making the badge system universal rather than hardcoded to Orbitar Award polls. That's an architectural decision that needs to be resolved before the smaller code-level items matter.

PR #520, opened 2026-04-12
Allow http redirect URIs for localhost, 127.0.0.1, and [::1]

pazoozoo42:
OAuth2 client registration and editing now accept `http://` (not just `https://`) as redirect URI scheme for local development hosts: `localhost`, `127.0.0.1` and `[::1]`.
This applies in all environments, not only in development mode, so that TUI/CLI apps can use loopback redirect URIs per RFC 8252 §7.3.

### DIFF
--- a/backend/src/api/ApiMiddleware.ts
+++ b/backend/src/api/ApiMiddleware.ts
@@ -153,12 +153,16 @@ export const urisListValidator = Joi.string().custom((value, helpers) => {
 
       // Strict validation for http/https
       if (urlObj.protocol === 'http:' || urlObj.protocol === 'https:') {
+        const hostname = urlObj.hostname
+        const isLocalhost =
+          hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '[::1]' || hostname === '::1'
+        const allowHttp = isLocalhost || process.env.NODE_ENV === 'development'
         if (
           !isURL(url, {
-            require_tld: process.env.NODE_ENV !== 'development',
+            require_tld: !isLocalhost && process.env.NODE_ENV !== 'development',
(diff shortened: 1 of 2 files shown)


Aivean2 (commented):
Clean, focused PR. RFC 8252 §7.3 compliance for loopback redirect URIs — standard practice for TUI/CLI OAuth flows.

**What I checked:**
- Registration validator in `ApiMiddleware.ts` is the single gatekeeper for redirect URIs (the OAuth library matches against stored URIs at authorization time, so this is the right place to change)
- `require_tld` relaxation for localhost addresses is correct — localhost has no TLD
- Test coverage is thorough: all three loopback variants ± port, negative cases for non-local HTTP, `localhost.localdomain` boundary, mixed lists

**Minor note:** `urlObj.hostname` from Node's WHATWG URL parser returns `[::1]` (with brackets) for IPv6 addresses, so the `hostname === '::1'` branch is technically unreachable through `new URL()`. Harmless — just defensive, but worth knowing.

Looks good to me technically. Deferring stamp to Aivean / 4vanger for the policy call on allowing HTTP in production.

Aivean (approved):

Aivean:
@Aivean2, note, you missed a frontend validator here

Aivean2:
Good catch — you're right. The frontend `validateUrls()` in `UserProfileClientAppsCreateForm.tsx` has the same http/https gate (line 100: `protocols: ['https', ...(process.env.NODE_ENV === 'development' ? ['http'] : [])]`) without the loopback exception. I'll put up a follow-up PR.

Aivean2:
Follow-up: #523

PR #521, opened 2026-04-12
fix: enforce ownership check on OAuth2 client edit

pazoozoo42:
The editClient endpoint did not verify that the requesting user owns the client being edited. Add user_id validation in OAuth2Manager (mirroring deleteClient logic) and an additional WHERE user_id guard in the repository SQL query. Includes unit tests for owner, non-owner, and missing-client scenarios.

### DIFF
--- a/backend/src/managers/OAuth2Manager.ts
+++ b/backend/src/managers/OAuth2Manager.ts
@@ -162,11 +162,27 @@ export default class OAuth2Manager {
 
   async editClient(
     clientId: string,
+    byUserId: number,
     description: string,
     redirectUris: string,
     initialAuthorizationUrl: string,
   ): Promise<boolean> {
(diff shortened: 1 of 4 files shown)


backend/src/managers/OAuth2Manager.ts
```diff
@@ -162,11 +162,27 @@ export default class OAuth2Manager {
 
   async editClient(
     clientId: string,
+    byUserId: number,
     description: string,
     redirectUris: string,
     initialAuthorizationUrl: string,
   ): Promise<boolean> {
-    return await this.oauthRepository.editClient(clientId, description, redirectUris, initialAuthorizationUrl)
+    const client = await this.oauthRepository.getClientByClientId(clientId)
```
Aivean, line 170: nit: this is redundant

Aivean (approved):

PR #525, opened 2026-05-11
Add profile votes feed

iamwavecut:
## Summary

This PR adds a private **Votes** section under **Self-Regulation**, with two tabs:

- **My Votes** shows votes the current user has cast on posts, comments, and user profiles. Votes can be changed or removed directly from the feed.
- **Votes for Me** shows votes received by the current user's posts, comments, and profile. This view is read-only and displays who voted, the vote value, and when it was last changed.

## Behavior

The feed is ordered by the most recent vote change and uses keyset pagination. Zero-value votes are not returned; when the user removes a vote from **My Votes**, the confirmed event is removed from the current feed immediately.

Events are grouped on the client across all loaded pages, so groups remain continuous after **Load more**. In **Votes for Me**, a post, comment, or profile is rendered once as a compact subject with its individual votes listed below it.

Filtering works across content and the other party involved in the vote. The API always uses the authenticated session user and does not accept another profile identity from the request body.

## Performance and consistency

- Feed selection uses bounded keyset scans instead of offset pagination or exact lifetime totals.
- A dedicated read repository returns normalized, direction-specific responses: full entities for **My Votes** and compact subjects for **Votes for Me**.
- Shared entities are serialized once, users are fetched in one batch, and user DTOs are projected through an explicit allowlist.
- Stale frontend requests are cancelled when the tab, filter, session user, or component lifecycle changes; a generation guard remains as a second safety check.
- Karma votes are updated in O(1) using deterministic user locking and atomic deltas, avoiding full-history `SUM(vote)` recalculation and reciprocal-vote deadlocks.
- Content votes use a consistent `entity -> vote -> aggregate` lock order and atomic aggregate upserts, reducing a changed vote to six statements and a no-op to two reads.
- The migration backfills at most 10,000 vote rows per transaction, is resumable, uses online DDL hints, validates exact index definitions, and preserves rollback-compatible additive schema.
- The read endpoint uses a pure restrictions snapshot and cannot trigger mass vote cleanup as a side effect.

## Validation

Verified on an isolated MySQL 5.7.44 instance:

- post and comment backfills completed in `10,000 + 5` row batches with no remaining NULL targets;
- repeated `up -> up -> down -> down` completed successfully;
- all six feed branches (`post/comment/user x mine/received`) avoided filesort;
- a 50-row profile-vote page over a 100,000-row history required 50 handler reads;
- simultaneous same-target votes, reciprocal profile votes, and concurrent content votes preserved their aggregates without deadlocks.

[...]

PR #535, opened 2026-08-25
test(parser): make wrong-key decryption assertions deterministic

iamwavecut:
## Summary

`CryptoUtils.test.ts › wrong input` intermittently fails CI (latest: run 32896847698 on #525 after a dev merge re-triggered the workflow). The test decrypted freshly-encrypted data with a wrong key and demanded `''`, but `aesDecryptFromBase64` is CBC+Pkcs7 without a MAC: with ~1/256 probability the garbage plaintext carries coincidentally valid padding and a non-empty string comes back. The encrypt IV is random, so every run sampled a new lottery ticket.

Measured over 10,000 iterations of the original scenario: 56 non-empty results (~1/178 per assertion; reviewer independently measured 35/8000 and 39/8000) — with two such assertions the test flaked roughly once per ~90 CI runs.

## Change

- The wrong-key and corrupted-input assertions now run against the **fixed ciphertext** already used by the round-trip test, so the strict `''` expectation stays and the RNG leaves the path entirely (per review): deterministic by construction, no weakened contract.
- Documented the no-MAC contract at `aesDecryptFromBase64`; the only production caller (`isValidBarmaliniPassword`) validates the decrypted value by parsing it as a date with a ±1h window (reviewer confirmed 0/8000 garbage outputs pass it).

## Verification

- `npx jest test/parser/CryptoUtils.test.ts` — 3/3, repeated runs stable.
- Both fixed-vector assertions verified to return `''` independently by reviewer and author.

### DIFF
--- a/backend/src/parser/CryptoUtils.ts
+++ b/backend/src/parser/CryptoUtils.ts
@@ -7,6 +7,10 @@ export function aesEncryptToBase64(data, key) {
   return CryptoJS.enc.Base64.stringify(toEncode)
 }
 
+// CBC+Pkcs7 without a MAC: a wrong key or corrupted input usually fails the
+// padding/UTF-8 checks and yields '', but can occasionally decode to garbage.
+// Callers must validate the returned plaintext (isValidBarmaliniPassword
+// parses it as a date and applies a time window).
 export function aesDecryptFromBase64(base64data, key) {
   try {
     const decoded = CryptoJS.enc.Base64.parse(base64data)
--- a/backend/test/parser/CryptoUtils.test.ts
+++ b/backend/test/parser/CryptoUtils.test.ts
@@ -1,30 +1,38 @@
-import {aesDecryptFromBase64, aesEncryptToBase64} from '../../src/parser/CryptoUtils';
+import { aesDecryptFromBase64, aesEncryptToBase64 } from '../../src/parser/CryptoUtils'
 
 test('aes encode decode', () => {
-    const key = 'mVRW8ycqHYZQSgYcBN8PUz8hVRWB5n5q';
-    expect(aesDecryptFromBase64(aesEncryptToBase64('test', key), key)).toBe('test');
-    expect(aesDecryptFromBase64('TGBv/lDGJdJo9SK8kDxiaEnu3i9dFxUhl5pDJVDzCG8=', key)).toBe('test');
-});
+  const key = 'mVRW8ycqHYZQSgYcBN8PUz8hVRWB5n5q'
+  expect(aesDecryptFromBase64(aesEncryptToBase64('test', key), key)).toBe('test')
+  expect(aesDecryptFromBase64('TGBv/lDGJdJo9SK8kDxiaEnu3i9dFxUhl5pDJVDzCG8=', key)).toBe('test')
+})
 
 test('wrong input', () => {
-    const key = 'mVRW8ycqHYZQSgYcBN8PUz8hVRWB5n5q';
-    expect(aesDecryptFromBase64(aesEncryptToBase64('test', key), key + '1')).toBe('');
-    expect(aesDecryptFromBase64('a' + aesEncryptToBase64('test', key), key)).toBe('');
-    expect(aesDecryptFromBase64('TGBv/lDGJdJo9SK8kDxiaEnu4i9dFxUhl5pDJVDzCG8=', key)).toBe('');
-});
+  const key = 'mVRW8ycqHYZQSgYcBN8PUz8hVRWB5n5q'
+  // 'test' encrypted under `key` (same vector as the round-trip test above).
+  // CBC+Pkcs7 without a MAC cannot promise '' for every wrong-key input: with
+  // ~1/256 probability the garbage plaintext carries coincidentally valid
+  // padding, so re-encrypting with a random IV each run made the old strict
+  // assertion flake. Fixed ciphertexts keep the assertion strict AND
+  // deterministic; the random-IV encrypt path stays covered by the
+  // round-trip tests.
+  const cipher = 'TGBv/lDGJdJo9SK8kDxiaEnu3i9dFxUhl5pDJVDzCG8='
+  expect(aesDecryptFromBase64(cipher, key + '1')).toBe('')
+  expect(aesDecryptFromBase64('a' + cipher, key)).toBe('')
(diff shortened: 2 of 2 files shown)


Aivean2 (commented):
Reproduced the diagnosis independently — it holds. Two notes: one on the test, one on the second half of the PR that the description doesn't cover.

## The flaky test — confirmed

8000 iterations of each assertion against `dev`'s `CryptoUtils`:

```
dec(enc('test', key), key + '1')  → non-empty  35/8000  (~1/229)
dec('a' + enc('test', key), key)  → non-empty  39/8000  (~1/205)
```

Same order as your 1/178. Root cause is right: CBC+Pkcs7 with no MAC, and the random IV in `aesEncryptToBase64` re-rolls it every run.

I also checked the thing one worries about whenever a crypto assertion gets weakened — the production path still fails closed. The garbage that comes back is short junk (`"]"`, `"P"`, `"6v%ߣ,"`); `new Date(junk)` → `Invalid Date` → `NaN`, and `Math.abs(NaN - Date.now()) < 3600000` is `false`. 0/8000 non-empty results would have passed `isValidBarmaliniPassword`'s window. So nothing is hiding behind the relaxed assertion.

**Suggestion:** the nondeterminism is in the *input*, not in the contract, so it can be removed without weakening the assertion. Test 1 already carries a fixed ciphertext for `'test'` under `key` — reusing it makes both cases deterministic by construction, no RNG anywhere in the path:

```js
const cipher = 'TGBv/lDGJdJo9SK8kDxiaEnu3i9dFxUhl5pDJVDzCG8=' // 'test' under key
expect(aesDecryptFromBase64(cipher, key + '1')).toBe('')
expect(aesDecryptFromBase64('a' + cipher, key)).toBe('')
```

Ran both — both return `''`. Honest tradeoff: the random-IV form samples a fresh wrong-key case each run, the fixed form pins one. If that sampling is the point, it wants a seeded or statistical test rather than a unit test that flakes once per ~90 CI runs. Your call — `.not.toBe('test')` is defensible; I just prefer keeping the strict assertion when determinism is this cheap. The doc comment on `aesDecryptFromBase64` is a good addition either way.

## The actions bump — this is the part I'd hold on

The second commit isn't in the description, and the split matters here for a concrete reason rather than a tidiness one: **the two halves have opposite verification stories.** The test fix is proven by the CI that ran. The actions bump is proven by nothing that ran.

The five green checks are `build.yml:build-lint` and `build-docker-dev.yml:{build-docker, validate-docker-compose}` — all `ubuntu-latest`, all `pull_request`. Meanwhile every bumped action is now `runs.using: node24`, and the release notes put a floor under self-hosted runners:

- [`checkout@v5`](https://github.com/actions/checkout/releases/tag/v5.0.0) — "Minimum Compatible Runner Version **v2.327.1**"
- [`cache@v5`](https://github.com/actions/cache/releases/tag/v5.0.0) and [`upload-artifact@v6`](https://github.com/actions/upload-artifact/releases/tag/v6.0.0) — same 2.327.1 floor, both explicitly calling out self-hosted
- [`setup-node@v5`](https://github.com/actions/setup-node/releases/tag/v5.0.0) — same floor

Four self-hosted jobs receive new actions, and none of them ran here:

| workflow / job | runner | bump | trigger |
|---|---|---|---|
| `deploy-universal` / `deployment` | `[self-hosted, <env>]` | checkout v3→v7 | dispatch |
| `backup-db` / `backup-db` | `[self-hosted, prod]` | upload-artifact v4→v7 | schedule |
| `backup-media` / `backup-media` | `[self-hosted, stage]` | checkout v3→v7 | schedule |
| `restore-sanitized-db` / `restore-db` | `[self-hosted, stage]` | checkout v3→v7 | schedule |

If the prod/stage runners are below 2.327.1, the first symptom isn't a red check — it's a failed production deploy, or a nightly backup that quietly stops producing artifacts. I can't read the runner versions myself (403 on the runners endpoint), so it's worth confirming under Settings → Actions → Runners before this lands.

**`review-check.yml` v7→v9** — I checked the script against the two documented v9 breaking changes (`require('@actions/github')` no longer works; `getOctokit` is now an injected parameter, so a `const`/`let getOctokit` declaration becomes a `SyntaxError`). Neither pattern appears anywhere in the file, so the script itself should be fine. But the workflow is `pull_request_target: [closed]`, which means GitHub runs the **base branch** copy — this bump cannot be exercised before merge, and it's the gate that audits merges. It has a `workflow_dispatch` with a `pr_number` input, so dispatching it from this branch against an already-merged PR would close that gap cheaply.

Minor, checked so it's on record: `setup-node@v5` added automatic caching when `package.json` carries a `packageManager` field, which would have doubled up with the explicit `cache` step in `build.yml` — neither `backend/` nor `frontend/` has that field, so it's a non-issue.

---

Leaving this as a comment rather than an approve — the deploy and backup paths are past what I'll sign off on alone. The test half reads good to me; splitting the actions bump into its own PR would let it be verified on its own terms.

iamwavecut:
Both points addressed.

**Test** — took the fixed-vector form in `708d7f6`: both wrong-key assertions now run against the round-trip test's ciphertext, the strict `''` expectation is back, and the RNG is out of the path entirely. Verified both vectors return `''` on my side as well; 3/3 across repeated runs. The sampling value of the random-IV form was exactly what flaked, so pinning the vectors is the right trade here.

**Actions bump** — it was never meant to be in this PR: the branch was cut from the fork's `dev`, which carried `24a5496` ahead of upstream, and I missed the stray commit when opening the PR. Rebased onto upstream `dev` (`dd3f553`) — the PR is now the single test commit. The bump goes into its own PR with the verification story you asked for, two data points already gathered from job logs (the runners API 403s for me too):

- **prod runner: 2.336.0** ("Set up job" of last night's `backup-db`, run 32923493952) — above the 2.327.1 floor, so the scheduled prod backup is safe;
- **stage runner: version unobtainable** — the last `restore-sanitized-db` run has its `[self-hosted, stage]` job stuck in `waiting` since 2026-08-24 and the prior one was cancelled, so no stage job log exists to read a version from. That runner looks offline or pending an environment approval — an infra question that predates the bump; flagged to the repo owner in the bump PR.

Your `review-check` dispatch idea and the `packageManager` non-issue note are carried into that PR's checklist as well.

Aivean (approved):
nice find! thanks

PR #537, opened 2026-08-28
fix(profile): don't save the viewed user's display name on a stray Enter

Aivean:
## Problem

Users' display names (`users.name`, the line under the nickname) were getting silently replaced with **other users' display names**. Reported today by lndr (`/u/Lender`), whose name became naked_singularity's «Плод распущенного воображения одержимых личностей».

Prod log for that session (UTC, 2026‑08‑28):

```
15:19:47  POST /user/profile   {"username":"naked_singularity"}
15:19:48  POST /user/comments  {"username":"naked_singularity","filter":""}
15:19:51  POST /user/savename  {"name":"Плод распущенного воображения одержимых личностей"}
15:19:52  POST /user/comments  {"username":"naked_singularity","filter":"призраками"}
```

i.e. open someone's profile → Comments tab → type in the filter box → press **Enter** → your own display name is overwritten with theirs. At least two more users were hit the same way in the last 5 days (`mahinaci9` → "Denis Reich", `pesnya_skazka` → "прекрасное").

## Cause

`UserProfileName` registered its `enter` hotkey unconditionally (since #340):

```ts
const refEditName = useHotkeys<HTMLInputElement>('enter', () => handleEditNameComplete(), {
  enableOnFormTags: ['INPUT'],
  preventDefault: true,
})
```

- The ref is only attached to the `<input>` while editing. When it's detached, react-hotkeys-hook binds the listener to `document` instead, and `enableOnFormTags: ['INPUT']` lets it fire from **any** `<input>` on the page (posts/comments filter, karma inputs…).
- `handleEditNameComplete` didn't check `mine`, and `name` state mirrors the *viewed* profile's name — so it called `saveName(<other user's name>)` with the visitor's session.

## Fix

- `enabled: editing && props.mine` on the hotkey — with `enabled: false` the hook doesn't attach a listener at all, so Enter in the filter box is no longer intercepted (nor `preventDefault`ed).
- Defensive early return in `handleEditNameComplete` unless `props.mine && editing`.

## Test plan

- [x] `eslint`, `prettier --check`, `tsc --noEmit` pass
- [x] Open another user's profile → Posts/Comments tab → type in the filter → Enter: filter applies, **no** `POST /user/savename` in the network tab — verified locally (logged in as `Lender`, `/u/naked_singularity/comments`, filter «призраками» + Enter → only `/user/comments`, DB name unchanged)
- [x] Control: same steps with the pre-fix component → `POST /user/savename {"name":"🕳️"}` fires and `Lender`'s name is overwritten (bug reproduced, then fix restored)
- [x] Own profile → pencil → edit name → Enter saves (`POST /user/savename` fires, name updates) — verified locally

Follow-up (ops, not in this PR): restore the overwritten names on prod — `Lender` (495) → «Иљя Илюша Илюха», `mahinaci9` (383) → «Тайка», `pesnya_skazka` (918) → «де б я був, якби мене не було?» (values from this morning's staging snapshot).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GEthEYzLXbksKU6Yiscjou

### DIFF
--- a/frontend/src/Components/UserProfileName.tsx
+++ b/frontend/src/Components/UserProfileName.tsx
@@ -19,7 +19,12 @@ export default function UserProfileName(props: UserProfileNameProps) {
   const [editing, setEditing] = useState(false)
   const [name, setName] = useState(props.name)
   const api = useAPI()
+  // Only listen for Enter while our own name is actually being edited. When the input is not
+  // rendered the ref is detached and react-hotkeys-hook falls back to a document-wide listener,
+  // so Enter pressed in any other <input> on the page (e.g. the posts/comments filter) would
+  // otherwise save the *viewed* profile's name onto the current user's account.
   const refEditName = useHotkeys<HTMLInputElement>('enter', () => handleEditNameComplete(), {
+    enabled: editing && props.mine,
     enableOnFormTags: ['INPUT'],
     preventDefault: true,
   })
@@ -33,6 +38,9 @@ export default function UserProfileName(props: UserProfileNameProps) {
   }
 
   const handleEditNameComplete = async () => {
+    if (!props.mine || !editing) {
+      return
+    }
     try {
       await api.userAPI.saveName(name)
       setEditing(false)


Aivean2 (approved):
Fix is correct — I checked it against the `react-hotkeys-hook@4.3.8` actually installed, not just the docs. Approving. Four notes, none of them blocking.

### 1. The mechanism checks out, and the two changes are **not** redundant

In v4 the hook body is, in effect:

```js
const memoisedOptions = useDeepEqualMemo(_options)
useSafeLayoutEffect(() => {
  if (memoisedOptions?.enabled === false || !isScopeActive(...)) return   // ← nothing is attached
  ...
  const domNode = ref.current || options?.document || document            // ← the fallback you describe
  domNode.addEventListener('keydown', handleKeyDown)
  ...
}, [_keys, memoisedOptions, enabledScopes])
```

So `enabled: false` really does mean *no listener at all* — which is also why Enter stops being `preventDefault`ed, exactly as the description says.

The part worth writing down before someone "simplifies" it later: **`enabled` is inside the deep-memoised options that form the effect's dependency list.** That makes it do double duty here — it's not only the guard, it's the thing that makes the effect re-run at the moment `editing` flips, and therefore the thing that makes the hook re-read `ref.current` and bind to the freshly mounted `<input>` instead of to `document`. The defensive early return in `handleEditNameComplete` alone would stop the write but leave a document-wide listener with `preventDefault: true` eating Enter in every other input on the page. Both halves earn their keep.

Same for `enableOnFormTags: ['INPUT']` — the tag check runs before the ref-scope check, so it's still load-bearing for the own-name path. Not dead weight now that `enabled` is there.

### 2. `!!!` The window is ~3 years, not 5 days

#340 (merged 2023-08-01) introduced exactly this call — `{enableOnFormTags: ['INPUT'], preventDefault: true}`, no `enabled` — and `react-hotkeys-hook@^4.3.8` has been pinned since #302 (2023-04-10). So the behaviour has been live for about three years, while the restore list is bounded by log retention plus who happened to complain.

`users.name` has no history anywhere in the schema (30 tables, none of them an audit/history table for it; `content_source` covers post/comment bodies only) — the `update users set name = :name` is destructive. So logs are the only *direct* enumeration, and they don't reach back far.

There is a cheap *indirect* one that covers the whole window: a victim's name ends up byte-identical to the profile they were looking at, so surviving cases show up as duplicates.

```sql
select name, count(*) n, group_concat(username separator ' | ') accounts
from users where name <> ''
group by name having n > 1
order by char_length(name) desc;
```

On a recent snapshot most of the output is organic noise from common given names (Андрей ×33, Дмитрий ×32, Александр ×27 …), but the long-name end of the list is short and worth a look:

- `прекрасное` → `daleko | pesnya_skazka` — **this is your `pesnya_skazka` case**, surfaced by the query without being told about it, and it also pins the direction (source = `daleko`). Decent positive control for the method.
- `Омелян Косопизд` → `4vanger | torn` — @4vanger can presumably say in one word whether that's an alt/joke or a hit.
- `Бляцкі Папа Ўсяго Арбітару` → `ZMsKG | ZMsКG` — those two usernames differ only by Latin `K` vs Cyrillic `К`, so almost certainly a deliberate twin rather than this bug. Listing it so it doesn't get "restored" by mistake.

Caveat on what that query can and can't say: it's a lower bound on *surviving* collisions, not a census. It can't see victims who already re-edited their name, or cases where the source has changed theirs since.

### 3. Same footgun one file over — follow-up, not a blocker

`CreateCommentComponent.tsx:191` is the only other `useHotkeys` call site that takes a ref (the other four — `Overlay`, `GalleryComponent` ×3, `SecretMailbox` ×2, `ZoomComponent` — are all document-bound by design and either scoped by mount or already guarded with `enabled`):

```ts
const containerRef = useHotkeys<HTMLDivElement>(allowedKeys.join(','), (e) => handleHotKey(e), {
  enableOnFormTags: ['TEXTAREA'],
  preventDefault: true,
})
```

`containerRef` is attached at line 720, inside `{previewing === null ? (<div className={styles.editor} ref={containerRef}>`. Its options are deep-equal on every render and `allowedKeys.join(',')` is a constant, so the layout effect runs **once per mount** and captures `domNode` then. Toggling Превью unmounts that div (the listener dies with the node); toggling back mounts a *new* div, and nothing re-runs to bind to it.

Predicted symptom — this is a code read, I haven't run it: open the comment editor → Ctrl+B works → Превью → Редактор → Ctrl+Enter / Ctrl+B / Ctrl+I / Ctrl+K / Ctrl+Shift+X are all dead until the component remounts. Different symptom from this PR (hotkeys stop working rather than corrupt data), same root cause: the returned ref is read exactly once, at effect time.

Fix is the same trick this PR uses — putting the mount condition into the deps:

```ts
  const containerRef = useHotkeys<HTMLDivElement>(allowedKeys.join(','), (e) => handleHotKey(e), {
    enabled: previewing === null,
    enableOnFormTags: ['TEXTAREA'],
    preventDefault: true,
  })
```

(`previewing` is declared above at line 154, so no reordering needed.)

### 4. On the green checks

Only `build-lint (backend)` and `build-lint (frontend)` ran, and there's no frontend test suite in the repo — so green here means it compiles and lints, not that the listener stopped binding to `document`. The manual control in your test plan (reproduce pre-fix, verify post-fix) is what actually covers this class; worth noting because "CI is green" carries no signal for it either way.

PR #539, opened 2026-09-04
feat(admin): loopback-only cache admin server

Aivean:
## Summary

- Adds a **loopback-only admin HTTP server** on `127.0.0.1:${ADMIN_PORT}` (default `5002`, `0` disables) inside the backend container. It is not published by Docker and not proxied by Caddy, so reaching it requires `docker exec` on the box. No credentials; a middleware additionally rejects any non-loopback TCP peer (checks `req.socket.remoteAddress`, not `req.ip`).
- Routes are **building blocks over single caches**, never workflows. Anything already easy with SQL or `redis-cli` (data edits, re-parsing via `parser_version`) stays out. `GET /` lists them:
  - `GET /cache/stats`, `GET /cache/users/inspect?userId=|username=` (cache vs DB, `drift` flag)
  - `POST /cache/users/evict {userId|username}`, `/cache/users/evict-all`
  - `POST /cache/usernames/rebuild` (the `@mention` autocomplete trie only grows on registration)
  - `POST /cache/karma/evict {userId|username, asVoter?}`, `/cache/karma/evict-all` (Redis `active_karma_votes_*`, `trial_progress_*`, `remove_votes_when_karma_is_low_*`, `is_user_active_*`; `karma_penalty_*` is state and is never touched)
  - `POST /cache/sites/evict {siteId|site}`, `/cache/sites/evict-all`
  - `POST /cache/sessions/evict {userId|sessionId}` (in-memory only; the `sessions` table is untouched)
- Fixes `UserCache.clearCache` to remove **every username key pointing at the id**, not only the current name. This is the orphan that survived the 2026-09-04 username rename on prod. Adds `clearAll`, `stats`, `inspect`, and an atomic trie rebuild.
- Adds per-cache clear/stats methods on `SiteManager`, `FeedManager`, `PostManager`, `InviteManager`, and in-memory session eviction helpers in `session/Session.ts`.
- **Removes the public `POST /user/clearCache`**: any logged-in user could call it, and it never sent a response (the connection hung until the client timed out). Nothing in the frontend used it.
- Docs: `docs/admin-server.md` (security model, route table, scope rule, rename and forced-logout recipes), linked from `AGENTS.md`.

## Usage

```bash
sudo docker exec orbitar-backend-1 curl -s localhost:5002/
sudo docker exec orbitar-backend-1 curl -s -X POST localhost:5002/cache/users/evict \
  -H 'Content-Type: application/json' -d '{"username":"OldName"}'
```

## Test plan

- [x] `backend`: `tsc --noEmit` clean, `eslint` 0 errors, `jest` 22 suites / 393 tests passing (16 new: admin routes and guard, `UserCache` orphan eviction and trie rebuild, session memory eviction)
- [x] staging (2026-09-04): deploy logs `Admin server is listening on 127.0.0.1:5002`; `GET /` lists the routes
- [x] staging security: host `127.0.0.1:5002` refused (port not published), sibling container `backend:5002` refused (loopback bind), public `stage-api.orbitar.space/cache/*` returns the app's 404
- [x] staging orphan reproduction: rename user 1625 in the DB, look the new name up via the public API, `inspect` shows both username keys with `consistent:false, drift:true`; `evict {username:<stale>}` removes both; the stale name then 404s on the public API; `usernames/rebuild` flips the autocomplete; name restored
- [x] staging karma: `/user/karma` populates `active_karma_votes_*`/`trial_progress_*`/`is_user_active_*`; `karma/evict` deletes the user's keys, `asVoter:true` also deletes the voted target's key, `evict-all` leaves the four families at zero
- [x] staging sessions: a session whose DB row was deleted keeps working until `sessions/evict` by `sessionId` or `userId`, after which the public API answers `auth-required`
- [x] staging errors: 400 for missing/both/typed params, 404 for unknown user and route, 400 for malformed JSON, all as JSON; `ADMIN` log lines carry params and payload

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BrDw6xRn4hdvMeksK6oNDJ

### DIFF
--- a/backend/src/admin/AdminServer.ts
+++ b/backend/src/admin/AdminServer.ts
@@ -190,8 +190,8 @@ export function buildAdminRoutes(deps: AdminDeps): AdminRoute[] {
       path: '/cache/karma/evict',
       description:
         'Delete the Redis karma caches of one user (userId or username): active_karma_votes, trial_progress, ' +
-        'remove_votes_when_karma_is_low, is_user_active. With asVoter=true also the active_karma_votes and ' +
-        'trial_progress of everyone the user has karma-voted (those are keyed by voter username). ' +
+        'remove_votes_when_karma_is_low, is_user_active. With asVoter=true also active_karma_votes_<targetId> of ' +
+        'everyone the user has karma-voted: that value is a map keyed by voter username, so it goes stale on a rename. ' +
         'karma_penalty is state, never touched.',
       schema: userRef.keys({ asVoter: Joi.boolean().default(false) }),
       handler: async (params) => {
--- a/backend/src/admin/AdminServer.ts
+++ b/backend/src/admin/AdminServer.ts
@@ -100,6 +100,14 @@ async function resolveUserId(userManager: UserManager, params: AdminParams): Pro
   return user.id
 }
 
+/** Session ids are bearer credentials; keep them out of the (root-readable) logs. */
+function maskSecrets(params: AdminParams): AdminParams {
+  if (typeof params.sessionId !== 'string') {
+    return params
+  }
+  return { ...params, sessionId: params.sessionId.slice(0, 6) + '…' }
+}
+
 export function buildAdminRoutes(deps: AdminDeps): AdminRoute[] {
   const { userManager, siteManager, feedManager, postManager, inviteManager } = deps
   return [
@@ -260,17 +268,18 @@ export function createAdminApp(deps: AdminDeps, logger: Logger): express.Express
         return
       }
       const params = value as AdminParams
+      const loggedParams = maskSecrets(params)
       try {
         const payload = await route.handler(params)
-        logger.info(`${route.method} ${route.path}`, { params, payload })
+        logger.info(`${route.method} ${route.path}`, { params: loggedParams, payload })
         res.json({ result: 'success', payload })
       } catch (err) {
         if (err instanceof AdminError) {
-          logger.warn(`${route.method} ${route.path}: ${err.message}`, { params })
+          logger.warn(`${route.method} ${route.path}: ${err.message}`, { params: loggedParams })
           res.status(err.status).json({ result: 'error', code: err.code, message: err.message })
           return
         }
-        logger.error(`${route.method} ${route.path} failed`, { params, error: err })
+        logger.error(`${route.method} ${route.path} failed`, { params: loggedParams, error: err })
         res.status(500).json({
           result: 'error',
           code: 'error',
--- a/backend/src/admin/AdminServer.ts
+++ b/backend/src/admin/AdminServer.ts
@@ -1,0 +1,317 @@
+import http from 'http'
+
+import express, { NextFunction, Request, Response } from 'express'
+import Joi from 'joi'
+import { Logger } from 'winston'
+
+import FeedManager from '../managers/FeedManager'
+import InviteManager from '../managers/InviteManager'
+import PostManager from '../managers/PostManager'
+import SiteManager from '../managers/SiteManager'
+import UserManager from '../managers/UserManager'
+import { evictSessionFromMemory, evictSessionsFromMemory, sessionMemoryStats } from '../session/Session'
+
+/**
+ * Loopback-only operations server.
+ *
+ * It listens on 127.0.0.1 inside the backend container, so the only way to reach it is
+ * `docker exec <backend> curl localhost:<ADMIN_PORT>/...`, i.e. root access to the box.
+ * Nothing here is proxied by Caddy and there are no credentials to leak.
+ *
+ * Every route is a building block over a single cache. Anything that is already easy to do
+ * with SQL or redis-cli (data changes, re-parsing content via parser_version, ...) does not
+ * belong here; this exists only for state that lives in the process memory or that cannot
+ * be targeted safely from outside.
+ *
+ * Responses use the same {result, payload} envelope as the public API. `GET /` lists the routes.
+ */
+
+export type AdminDeps = {
+  userManager: UserManager
+  siteManager: SiteManager
+  feedManager: FeedManager
+  postManager: PostManager
+  inviteManager: InviteManager
+}
+
+export class AdminError extends Error {
+  constructor(
+    public readonly status: number,
+    public readonly code: string,
+    message: string,
+  ) {
+    super(message)
+  }
+}
+
+const LOOPBACK_ADDRESSES = new Set(['127.0.0.1', '::1', '::ffff:127.0.0.1'])
+
+export function isLoopbackAddress(address: string | undefined): boolean {
+  return address !== undefined && LOOPBACK_ADDRESSES.has(address)
+}
+
+/**
+ * Defense in depth: the server already binds to 127.0.0.1, this rejects anything else that could
+ * still get through (a changed bind address, a misconfigured proxy). Checks the raw TCP peer, not
+ * `req.ip`, because the latter honours X-Forwarded-For.
+ */
+export function loopbackOnly(logger: Logger) {
+  return (req: Request, res: Response, next: NextFunction) => {
+    const address = req.socket.remoteAddress
+    if (!isLoopbackAddress(address)) {
+      logger.warn('Rejected admin request from a non-loopback address', { address, url: req.url })
+      res.status(403).json({ result: 'error', code: 'forbidden', message: 'Admin API is loopback-only' })
+      return
+    }
+    next()
+  }
+}
+
+type AdminParams = Record<string, unknown>
+
+type AdminRoute = {
+  method: 'GET' | 'POST'
+  path: string
+  description: string
+  schema: Joi.ObjectSchema
+  handler: (params: AdminParams) => Promise<unknown>
+}
+
+const userRef = Joi.object({
+  userId: Joi.number().integer().positive(),
+  username: Joi.string(),
+}).xor('userId', 'username')
+
+const noParams = Joi.object({})
+
+/**
+ * A username is resolved through the user cache first (then the DB), on purpose: the typical
+ * reason to evict by name is that the cache still holds a name the DB no longer has.
+ */
+async function resolveUserId(userManager: UserManager, params: AdminParams): Promise<number> {
+  if (typeof params.userId === 'number') {
+    return params.userId
+  }
+  const username = String(params.username)
+  const user = await userManager.getByUsername(username)
+  if (!user) {
+    throw new AdminError(404, 'not-found', `User "${username}" is neither cached nor in the DB`)
+  }
+  return user.id
+}
+
+export function buildAdminRoutes(deps: AdminDeps): AdminRoute[] {
+  const { userManager, siteManager, feedManager, postManager, inviteManager } = deps
+  return [
+    {
+      method: 'GET',
+      path: '/cache/stats',
+      description: 'Sizes of every in-process cache.',
+      schema: noParams,
+      handler: async () => ({
+        users: userManager.userCacheStats(),
+        sites: siteManager.cacheStats(),
+        feedSubscriptions: feedManager.userSubscriptionsCacheStats(),
+        contentCounters: postManager.contentNumberCacheStats(),
+        invitesAvailability: inviteManager.invitesAvailabilityCacheStats(),
+        sessions: sessionMemoryStats(),
+      }),
+    },
+    {
+      method: 'GET',
+      path: '/cache/users/inspect',
+      description:
+        'Cached user entries next to the DB row (userId or username). ' +
+        '`drift` is true when a cached username differs from the DB, `cache.consistent` is false when a username key is orphaned.',
+      schema: userRef,
+      handler: async (params) => {
+        const userId = await resolveUserId(userManager, params)
+        const cache = userManager.inspectUserCache(userId)
+        const db = await userManager.getUserInfoFromDb(userId)
+        const dbUsername = db ? db.username : undefined
+        const drift =
+          (cache.byId !== undefined && cache.byId.username !== dbUsername) ||
+          cache.usernameKeys.some((username) => username !== dbUsername)
+        return { userId, cache, db, drift }
+      },
+    },
+    {
+      method: 'POST',
+      path: '/cache/users/evict',
+      description:
+        'Evict one user (userId or username) from every in-process cache: identity by id and by any username key, ' +
+        'restrictions, last visit, feed subscriptions, content counters, invite availability. ' +
+        'A username is looked up in the cache first, so a stale name still works.',
+      schema: userRef,
+      handler: async (params) => {
+        const userId = await resolveUserId(userManager, params)
+        return {
+          userId,
+          user: userManager.evictUserCaches(userId),
+          feedSubscriptions: feedManager.clearUserSubscriptionsCache(userId),
+          contentCounters: postManager.clearContentNumberCache(userId),
+          invitesAvailability: inviteManager.clearInvitesAvailabilityCache(userId),
+        }
+      },
+    },
+    {
+      method: 'POST',
+      path: '/cache/users/evict-all',
+      description:
+        'Drop every in-process user cache. The username suggestions trie is kept, rebuild it separately. ' +
+        'Returns the sizes that were dropped.',
+      schema: noParams,
+      handler: async () => ({
+        user: userManager.clearAllCaches(),
+        feedSubscriptions: feedManager.clearUserSubscriptionsCache(),
+        contentCounters: postManager.clearContentNumberCache(),
+        invitesAvailability: inviteManager.clearInvitesAvailabilityCache(),
+      }),
+    },
+    {
+      method: 'POST',
+      path: '/cache/usernames/rebuild',
+      description:
+        'Rebuild the username suggestions trie (@mention autocomplete) from the DB. ' +
+        'The trie only grows on registration, so renames and deletions need this.',
+      schema: noParams,
+      handler: async () => ({ usernames: await userManager.rebuildUsernameSuggestions() }),
+    },
+    {
+      method: 'POST',
+      path: '/cache/karma/evict',
+      description:
+        'Delete the Redis karma caches of one user (userId or username): active_karma_votes, trial_progress, ' +
+        'remove_votes_when_karma_is_low, is_user_active. With asVoter=true also the active_karma_votes and ' +
+        'trial_progress of everyone the user has karma-voted (those are keyed by voter username). ' +
+        'karma_penalty is state, never touched.',
+      schema: userRef.keys({ asVoter: Joi.boolean().default(false) }),
+      handler: async (params) => {
+        const userId = await resolveUserId(userManager, params)
+        return { userId, deleted: await userManager.evictKarmaCaches(userId, params.asVoter === true) }
+      },
+    },
+    {
+      method: 'POST',
+      path: '/cache/karma/evict-all',
+      description:
+        'Delete every Redis karma cache key, same families as /cache/karma/evict. Returns counts per pattern.',
+      schema: noParams,
+      handler: async () => ({ deleted: await userManager.evictAllKarmaCaches() }),
+    },
+    {
+      method: 'POST',
+      path: '/cache/sites/evict',
+      description: 'Evict one site from the in-process site cache (siteId or site subdomain).',
+      schema: Joi.object({
+        siteId: Joi.number().integer().positive(),
+        site: Joi.string(),
+      }).xor('siteId', 'site'),
+      handler: async (params) => ({
+        removed: siteManager.clearCache({ siteId: params.siteId as number, site: params.site as string }),
+      }),
+    },
+    {
+      method: 'POST',
+      path: '/cache/sites/evict-all',
+      description: 'Drop the whole in-process site cache.',
+      schema: noParams,
+      handler: async () => ({ removed: siteManager.clearCache() }),
+    },
+    {
+      method: 'POST',
+      path: '/cache/sessions/evict',
+      description:
+        'Drop in-memory session entries for a user (userId) or one session (sessionId). ' +
+        'The sessions table is not touched: delete rows there first when the goal is a forced logout, ' +
+        'the process keeps honouring cached sessions until they are evicted here.',
+      schema: Joi.object({
+        userId: Joi.number().integer().positive(),
+        sessionId: Joi.string(),
+      }).xor('userId', 'sessionId'),
+      handler: async (params) =>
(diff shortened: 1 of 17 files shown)


backend/src/session/Session.ts
```diff
@@ -28,6 +28,43 @@ function deleteFromUserSessions(userId: number, sessionId: string) {
   }
 }
 
+/**
+ * Drops the in-memory copies of a user's sessions. Rows in the `sessions` table are left alone:
+ * pair this with a DB delete to force a logout. The process keeps honouring cached sessions
+ * until they are evicted here, so a DB-only cleanup is not enough.
+ */
+export function evictSessionsFromMemory(userId: number): number {
```
Aivean2, line 39: `!!!` See the review body. `sessionsByUser` is only ever written by `addToUserSessions` at line 118, i.e. the DB-fallback branch of `restore()` — the login path (`init()` → `data.userId = …` → `store()`) never registers the session. So this returns `0` for every session created since the last backend restart, and the forced-logout recipe in `docs/admin-server.md` silently no-ops on exactly the live ones.

backend/src/managers/UserCache.ts
```diff
@@ -104,14 +141,67 @@ export class UserCache {
...
+   * Evicts everything cached for the user. Username keys are matched by the cached id rather than
+   * by the name of the current id entry, so keys left behind by a rename are removed as well.
+   */
+  public clearCache(userId: number): UserCacheEviction {
+    const byId = userId in this.cacheId
     delete this.cacheId[userId]
-    delete this.cacheUsername[cacheEntry.username]
+
+    const usernameKeys: string[] = []
+    for (const [username, user] of Object.entries(this.cacheUsername)) {
```
Aivean2: This turns `clearCache` from O(1) into O(cached usernames) with a full `Object.entries` materialisation, and `UserManager.removeVotesWhenKarmaIsLow` (`UserManager.ts:506`) calls it once per karma vote in a loop. The orphan fix is correct — keeping a `Record<number, Set<string>>` updated in `cache()` gets the same result in O(1) and also gives `inspect()` its `usernameKeys` directly.

backend/src/admin/AdminServer.ts
```diff
@@ -0,0 +1,317 @@
...
+      const raw = route.method === 'GET' ? req.query : req.body
+      const { error, value } = route.schema.validate(raw || {}, { convert: true })
+      if (error) {
+        res.status(400).json({ result: 'error', code: 'invalid-params', message: error.message })
+        return
+      }
+      const params = value as AdminParams
+      try {
+        const payload = await route.handler(params)
+        logger.info(`${route.method} ${route.path}`, { params, payload })
```
Aivean2: `params` at info level means `POST /cache/sessions/evict {"sessionId": …}` writes a live session token to the backend log. Maybe redact that one key, or log `Object.keys(params)` for the session route.

Aivean2 (commented):
Read the whole diff. One `!!!`, one perf note, the rest is minor — and the `!!!` is a pre-existing index bug that this PR is the first thing to depend on.

### !!! `/cache/sessions/evict {userId}` evicts nothing for sessions created since the last restart

`sessionsByUser` has exactly one writer: `addToUserSessions` at `Session.ts:118`, inside the **DB-fallback branch of `restore()`**. The login path never touches it:

```
AuthController.ts:148   await request.session.init()      // sessionStorage[id] = {…}, data.userId still undefined
AuthController.ts:158   request.session.data.userId = …   // mutates the stored object in place
AuthController.ts:159   await request.session.store()     // `sessions` table only
```

(same shape at `InviteController.ts:157/167`).

So a session joins the index only *after* the process has forgotten it and re-read it from the DB — i.e. only after a backend restart. `evictSessionsFromMemory(userId)` therefore returns `0` and removes nothing for every session signed in since then, which is most of the live ones.

That makes the **Forcing a logout** recipe in `docs/admin-server.md` fail in the dangerous direction: delete the rows, get `{"removed":0}`, read it as "nothing was cached" — and the user stays authenticated indefinitely, because `restore()` hits `sessionStorage` first and never consults the DB again. `/cache/stats` → `sessions.users` under-reports for the same reason.

`evictSessionFromMemory(sessionId)` is fine — it reads `sessionStorage` directly.

Two ways out:

- One line at the commit point. `store()` runs immediately after the `userId` assignment on both paths and already has `this.id` and `this.data.userId`:
  ```ts
  async store() {
    if (!this.id) return
    if (this.data.userId) {
      addToUserSessions(this.data.userId, this.id)
    }
    …
  ```
- Or make the admin route index-independent: `sessionStorage` is a plain object, so scanning it by `data.userId` is O(live sessions) on a hand-called route and stays correct whatever the index does.

The first also fixes the pre-existing sibling: `destroyAllForCurrentUser()` ("log out everywhere") only kills sessions that survived a restart, for the same reason.

Worth noting the new test can't see this — both cases in `SessionMemory.test.ts` seed through `restore()`, which is the one path that populates the index.

### perf: `UserCache.clearCache` went O(1) → O(cached usernames), and it is called in a loop

`Object.entries(this.cacheUsername)` materialises the whole map on every call. `removeVotesWhenKarmaIsLow` (`UserManager.ts:506`) calls it once per karma vote of the user, and `cacheUsername` only grows — seeded with up to 1000 from `getLastActiveUsers()`, then one entry per lookup, never evicted otherwise.

The orphan fix is right; the scan doesn't have to be. A `Record<number, Set<string>>` maintained in `cache()` keeps eviction O(1) and hands `inspect()` its `usernameKeys` for free.

### minor

- `AdminServer.ts:265` logs `params` at info level, so `POST /cache/sessions/evict {"sessionId": …}` writes a live bearer token into the backend log. Everything else that reaches `params` is ids and usernames.
- The PR body and the `/cache/karma/evict` description say the vote caches are "keyed by voter username". The *keys* are `active_karma_votes_<targetId>`; it is the cached *value* that is `Record<voterUsername, vote>` (`UserManager.getActiveKarmaVotes`). Same conclusion for `asVoter`, but `docs/admin-server.md` is the file someone reads mid-rename, so it is worth being exact there.
- `asVoter` also evicts `trial_progress_<targetId>`, which is computed from karma numbers only and cannot go stale on a voter rename. Harmless, just wider than the stated reason.

### checked, came back clean

- `loopbackOnly` is registered before `express.json()` and reads the raw TCP peer, so no `X-Forwarded-For` path. I went looking for the escalation that would matter here — an in-process fetcher of user-supplied URLs would present to this server as `127.0.0.1` — and there isn't one: the only outbound calls in `backend/src` are `Mailer.ts` (fixed host) and `FeedManager` (`FEED_API`). No `network_mode: host` in any compose file either, so a sibling container can't reach the namespace.
- Removing `POST /user/clearCache`: no caller anywhere in `frontend/src`, and it never sent a response, so nothing could have been depending on it.
- `SiteManager.clearCache` handles `{siteId: undefined, site: undefined}` explicitly, so the `.xor` route can't fall through into the clear-all branch.
- `express@4`, so `app.all('*')` is fine, and the 4-arg error middleware still catches `express.json()` parse errors from its position after it.
- `usernamesSuggestionsCount` has one incrementing caller (registration, `UserManager.ts:274`), so `/cache/stats` won't drift from the trie.

Not approving — the new surface and its security model are yours to sign off on, and I don't want to stamp a listener. Nothing else here blocks.

Aivean:
## Staging test results (orbitar2, 2026-09-04, commit bbf60ec4)

Deploy log: `[info] ADMIN Admin server is listening on 127.0.0.1:5002 (loopback only)`. All calls below were made with `docker exec orbitar-backend-1 curl localhost:5002/...`; public API calls used a temporary session for user 11 that was deleted afterwards. Staging DB state was restored at the end (user 1625 is `Stervo` again, no temp sessions left).

### 1. Security boundary

| Vantage point | Result |
| --- | --- |
| inside the backend container, `localhost:5002/` | 200, lists 10 routes |
| staging host, `127.0.0.1:5002/` | connection refused (`docker port` shows nothing published) |
| sibling container (caddy), `http://backend:5002/` | `can't connect to remote host (172.29.0.6): Connection refused`; `backend:5001` answers, so this is the loopback bind, not the network |
| public domain, `https://stage-api.orbitar.space/cache/stats` and `POST .../cache/users/evict-all` | the app's own `{"result":"error","code":"404","message":"Unknown endpoint"}` |

### 2. Orphaned username key: live reproduction of the prod incident

1. `POST /api/v1/user/profile {"username":"Stervo"}` (public) → cached. `GET /cache/users/inspect?userId=1625` → `usernameKeys:["Stervo"], consistent:true, drift:false`.
2. `UPDATE users SET username='Dobrokot' WHERE user_id=1625` → 1 row. Public lookup `{"username":"Dobrokot"}` → 200.
3. `GET /cache/users/inspect?userId=1625` →
   ```json
   "cache":{"byId":{"id":1625,"username":"Dobrokot",...},"usernameKeys":["Stervo","Dobrokot"],"consistent":false,...},
   "db":{"id":1625,"username":"Dobrokot",...},"drift":true
   ```
   `GET /cache/users/inspect?username=Stervo` resolves through the cache to the same user (stale-name lookup works as intended).
4. `POST /cache/users/evict {"username":"Stervo"}` →
   ```json
   {"userId":1625,"user":{"byId":true,"usernameKeys":["Stervo","Dobrokot"],"publicKey":true,"parent":true,"stats":false,"restrictions":true,"lastVisit":false},"feedSubscriptions":0,"contentCounters":{"posts":1,"comments":1},"invitesAvailability":0}
   ```
5. Inspect again → `usernameKeys:[], consistent:true, drift:false`. Public API: `Stervo` → `not-found`, `Dobrokot` → 200. On prod today the old name kept resolving after eviction; here it does not.
6. `POST /cache/usernames/rebuild` → `{"usernames":2352}` (= `SELECT COUNT(*) FROM users`). Public `suggest-username "Dob"` → `["Dobrokot","Dobroom"]`, `"Ste"` no longer lists Stervo.
7. Restored: `UPDATE ... 'Stervo'`, `evict {"userId":1625}` → `usernameKeys:["Dobrokot"]`, rebuild → 2352, `"Ste"` lists `Stervo` again, `Dobrokot` → 404, inspect clean.

### 3. Karma Redis caches

- Populated via public `/user/karma` for Stervo and for user 2350 (a target of voter 11). Redis after: `active_karma_votes_11, active_karma_votes_1625, active_karma_votes_2350 | trial_progress_1625`, `is_user_active_*`=370.
- `POST /cache/karma/evict {"username":"Stervo"}` → `deleted:["active_karma_votes_1625","trial_progress_1625","is_user_active_1625"]`
- `POST /cache/karma/evict {"userId":11,"asVoter":true}` → `deleted:["active_karma_votes_11","is_user_active_11","active_karma_votes_2350"]` (the voted target's key is included)
- `POST /cache/karma/evict-all` → `{"active_karma_votes_*":0,"trial_progress_*":0,"remove_votes_when_karma_is_low_*":0,"is_user_active_*":368}`; SCAN afterwards shows all four families at 0. `karma_penalty_*` was not touched (none present).

### 4. Sessions

- Public profile call with the temp session → 200. `DELETE FROM sessions WHERE id=...` → 1 row. Same public call → **still 200** (served from process memory; this is the problem the route exists for).
- `POST /cache/sessions/evict {"sessionId":"..."}` → `{"removed":1}`. Same public call → `auth-required`.
- Repeated with a fresh session and `{"userId":11}` → `{"removed":1}`, then `auth-required` after the DB row was deleted. `/cache/stats` showed `sessions:{"sessions":2,"users":2}` beforehand.

### 5. Sites

`/cache/stats` → `sites:{"byName":1,"byId":1}`; `evict {"site":"main"}` → `removed:2`; `evict {"siteId":2}` → `removed:0` (not cached); `evict-all` → `removed:0`; stats → `{"byName":0,"byId":0}`.

### 6. Validation and errors (all JSON)

| Request | Status | code / message |
| --- | --- | --- |
| `evict {}` | 400 | `"value" must contain at least one of [userId, username]` |
| `evict {"userId":1,"username":"x"}` | 400 | `contains a conflict between exclusive peers [userId, username]` |
| `evict {"username":"no_such_user_xyz"}` | 404 | `User "no_such_user_xyz" is neither cached nor in the DB` |
| `GET /cache/users/inspect` (no params) | 400 | as above |
| `karma/evict {"userId":"abc"}` | 400 | `"userId" must be a number` |
| `sessions/evict {}` | 400 | `must contain at least one of [userId, sessionId]` |
| `sites/evict {"site":"main","siteId":1}` | 400 | exclusive peers |
| `GET /cache/nope` | 404 | `Unknown admin endpoint, GET / lists them` |
| body `{not json` | 400 | `bad-request` |

### 7. Logging

Every call produced an `ADMIN` line with params and payload, e.g.
`[info] ADMIN POST /cache/karma/evict {"params":{"asVoter":true,"userId":11},"payload":{"deleted":[...]}}`; the 404 case logs at `warn`.

Finding: `sessions/evict` logged the full session id at info level, which no other info-level line does. Fixed in 5522f28d (masked to six characters, unit-tested). That commit is log-only and is not on staging yet.

backend/src/session/Session.ts
Aivean: Confirmed and fixed in bb26a279: `store()` now calls `addToUserSessions` when `data.userId` is set, so the login path (`init()` → `data.userId = …` → `store()`) indexes the session; `destroyAllForCurrentUser()` benefits the same way. `SessionMemory.test.ts` gained a case that seeds through `init()`/`store()` and evicts by user id (ef386737 fixes the describe-block structure of that test). Note for the staging results above: the by-userId case there passed only because the temp session was created by a DB insert and restored via the fallback branch, which is exactly the path that was indexed before. It needs a re-check with a real login once this is deployed.

backend/src/managers/UserCache.ts
Aivean: Fixed in bb26a279. `UserCache` keeps `usernameKeysById: Map<number, Set<string>>`, maintained in `cache()`; `clearCache()` and `inspect()` walk only that set, and a key that another user has taken over since (rename collision) is left alone, with a test for that case. `Object.entries` is gone from both.

backend/src/admin/AdminServer.ts
Aivean: Fixed in 5522f28d: `sessionId` is masked to its first six characters in the log line (the full value is still used for the eviction), unit-tested.

Aivean:
Review addressed in bb26a279 and ef386737 (5522f28d for the log line):

- **`!!!` sessions index**: `store()` indexes the session under its user id, so the login path is covered and `destroyAllForCurrentUser()` is fixed along the way. Test added through `init()`/`store()`.
- **perf**: reverse index `Map<userId, Set<username>>` in `UserCache`; `clearCache()`/`inspect()` are O(keys of that user) again, with a guard for a key taken over by another user.
- **minor**: `sessionId` masked in the log; the `/cache/karma/evict` description and `docs/admin-server.md` now say keys vs values precisely (`active_karma_votes_<targetId>` holds a `Record<voterUsername, vote>`); `asVoter` no longer deletes `trial_progress_<targetId>`, which is a number derived from the votes and cannot go stale on a voter rename.

`tsc`, `npm run lint`, and the full backend suite (22 suites, 396 tests) pass. Staging still runs bbf60ec4; the by-userId session eviction there was exercised only through the DB-fallback path, so it should be re-checked with a real login after the next deploy.

Aivean:
## Staging test results, round 2 (orbitar2, 2026-09-04, deploy of ef386737)

Deployed build verified: `usernameKeysById` present in `build/managers/UserCache.js`, three `addToUserSessions` call sites in `build/session/Session.js` (definition, `restore()`, `store()`), session-id masking in `build/admin/AdminServer.js`.

### Sessions created by a real login are now evictable by user id (the `!!!` item)

Signed in through the public `/auth/signin` as the shared `barmalini` account (password generated in-container with the app's own `aesEncryptToBase64`, never displayed), which is the `init()` → `store()` path.

| step | result |
| --- | --- |
| `/cache/stats` before / after login | `sessions:1, users:2` → `sessions:2, users:3` (indexed at `store()` time) |
| profile call with the login session | 200 |
| `DELETE FROM sessions WHERE id=…` then the same call | 1 row deleted, **still 200** (served from memory) |
| `POST /cache/sessions/evict {"userId":537}` | `{"removed":1}` (would have been `0` before bb26a279) |
| same call again | `auth-required` |
| `/cache/stats` after | `sessions:1, users:2` |

DB session count for user 537 unchanged before/after (2510), only the test row was created and removed.

### Reverse index (perf item), live

Cached Stervo → renamed the DB row to Dobrokot → looked the new name up → inspect: `usernameKeys:["Stervo","Dobrokot"], consistent:false, drift:true`. `evict {"username":"Stervo"}` → both keys removed; inspect clean; public API 404 for Stervo, 200 for Dobrokot. Restored, rebuilt trie (2352), suggestions and profile correct again.

### `asVoter` narrowed

Populated caches for voter 11's targets. `karma/evict {"userId":11,"asVoter":true}` → `deleted:["active_karma_votes_11","is_user_active_11","active_karma_votes_1625","active_karma_votes_2350"]`; `trial_progress_1625` (a voted target) was left in place, as intended. `evict-all` afterwards zeroed all four families.

### Log masking, live

`[info] ADMIN POST /cache/sessions/evict {"params":{"sessionId":"deadbe…"},"payload":{"removed":0}}`

### Found and fixed in c01c60c3

`/cache/stats` reported `users:2` with `sessions:1` after a signout: `destroy()` emptied the user's set but left it in the index. `deleteFromUserSessions` now drops the entry when the set becomes empty; unit-tested. Not on staging yet, stats-only.

Staging state restored: user 1625 is `Stervo`, no temporary sessions left, Redis karma families empty.

Aivean2:
Re-checked at `c01c60c`. All four items addressed; both build-lint jobs green.

On the `!!!` one specifically — `addToUserSessions` now has three call sites (definition, the DB-fallback branch of `restore()`, and `store()`), and `store()` is called from exactly two places in `backend/src`, `AuthController.ts:159` and `InviteController.ts:168`, both of which are the `init()` → `data.userId = …` → `store()` login path. So the coverage is complete rather than incidental: there is no `store()` caller left that could create a session the index does not see.

The part I want to record as retired, since it was the reason I flagged this despite a green suite: `SessionMemory.test.ts` used to seed exclusively through `restore()`, the one path that already populated the index, so it was green *by construction* and could not see the bug. The new `sessions created through the login path…` case builds the session through `init()` → `data.userId = 7` → `store()` and then asserts `evictSessionsFromMemory(7) === 1` — on the old code that returned `0`, so the test now fails for the right reason if the `store()` line is ever dropped.

Two smaller things I noticed while diffing, both correct as written:

- `deleteFromUserSessions` dropping the empty `Set` also fixes `sessionMemoryStats().users` over-counting after the last session of a user goes away — worth knowing since `users` is what an operator reads back from `/cache/sessions/stats`.
- `evictSessionFromMemory` de-indexes too, so an id cannot outlive its `sessionStorage` entry in `sessionsByUser`; combined with the above, the index and the storage can't drift.

The `usernameKeysById` reverse index can retain a key that another user has since taken over, but both `clearCache` and `inspect` re-check ownership against `cacheUsername` before using it, and the collision test pins that — so it is a bounded bookkeeping remnant, not a correctness issue.

Still not approving, for the same reason as the first pass: the residual question is the new loopback listener and its security model, which is your call rather than a correctness finding of mine.

Aivean2:
@Aivean — mechanical status, not a re-review: as far as I can tell this PR is blocked on **me**, and I don't think either of us has said so out loud.

Current state at `c01c60c`: both `build-lint` jobs green, branch `MERGEABLE`, no conflicts — but `mergeStateStatus: BLOCKED` / `reviewDecision: REVIEW_REQUIRED`, and the only outstanding requirement I can see is the review request still sitting on `Aivean2`. My review on this PR is `COMMENTED`, which doesn't satisfy that gate (cf. #535, where a `COMMENTED` from me still needed a separate approving review before it could go in).

I'm not going to convert it to an approval, for the same reason as before: all four of my code findings are fixed and I verified them in the diff rather than taking your word for it, but the loopback listener and the security model around it are your architectural call, not mine. That reservation is about who signs off, not about anything I think is still wrong with the code.

So the approval has to come from somewhere else — dismiss or re-route my review request to another reviewer, or merge past it. Otherwise this sits indefinitely: I'd be waiting on a merge, and the merge would be waiting on me.

Aivean2:
Correction to my own comment above, twenty minutes later — I overstated it.

I said the pending review request on me was the only outstanding requirement I could see. That doesn't survive checking: #532 is `APPROVED`, has no pending review request, no unresolved threads and green checks, and is *also* `mergeStateStatus: BLOCKED`. So whatever blocks merges here is not (only) my review request, and I can't read the branch protection settings to tell you what it is.

What does hold up, and is the part worth your attention: the last eight merges on this repo all carried an approving review and none went in at `REVIEW_REQUIRED`; my review here is `COMMENTED`, which didn't satisfy that on #535 either; and the request is still open on me. I've also now resolved my three review threads — all findings fixed and verified at `c01c60c` — so nothing of mine is left dangling except the approval itself, which I'm still not giving, for the reason above: that call is yours, not mine.

If the merge path here is an admin override anyway, say so and I'll stop treating my pending request as if it means something.

backend/src/managers/UserCache.ts
```diff
@@ -15,31 +43,42 @@ export class UserCache {
...
-        entries.push({
-          k: user.username.toLowerCase(),
-          v: user.username,
-        })
-      })
-      this.usernamesSuggestionsCache.addAll(entries)
+      trie.addAll(usernames.map((user) => ({ k: user.username.toLowerCase(), v: user.username })))
+      count += usernames.length
     }
+    this.usernamesSuggestionsCache = trie
```
iamwavecut, line 79: If someone registers while this rebuild is waiting on the DB, `addUsernameSuggestion()` updates the old trie. This swap then drops their name from autocomplete until the next rebuild or restart. Reproduced locally. Could we preserve additions made during the rebuild and add a test for this overlap?

backend/src/config.ts
```diff
@@ -65,6 +67,7 @@ export type OAuthConfig = {
 
 export const config: Config = {
   port: parseInt(process.env.SERVER_PORT) || 5001,
+  adminPort: process.env.ADMIN_PORT === undefined ? 5002 : parseInt(process.env.ADMIN_PORT) || 0,
```
iamwavecut, line 70: Neither `docker-compose.local.yml` nor `docker-compose.ssl.local.yml` passes `ADMIN_PORT` to the backend, so setting `ADMIN_PORT=0` in the root `.env` still leaves the admin server running on 5002. Could we forward it in both Compose files and add it to `.env.sample`?

PR #542, opened 2026-09-16
feat(votes): group received votes by discussion and add filters

iamwavecut:
Follow-up to #525, prompted by the Discord feedback after the staging rollout: a couple of weeks of votes turn «Оценки мне» into a long wall of `+1` rows. This PR groups the tab by discussion and replaces the text filter with filters that fit votes.

## What changes

- **One card per discussion.** Votes are split into calendar weeks (months for «За год» and «За всё время»). Your post and your comments in the same discussion share a card. Each row shows its sum and the first few voters, and «и ещё N» opens the full timeline. Minus voters always come first, and each week names whoever voted most often.
- **Small stuff folds away.** Discussions with just one or two pluses collapse into one line at the end of the week.
- **Filters.** One dropdown for what (all, comments, posts, karma; pluses and minuses or only minuses), one for the period (week, 2 weeks by default, month, year, all time). Only one is open at a time, and the choice is kept in the URL.
- Clicking a quote expands the comment in place (Ctrl/Cmd-click still opens it). An empty period offers a wider one.
- «Мои оценки» is unchanged.

## API

`/user/votes` with `direction: 'received'` accepts optional `type` (`post` | `comment` | `user`), `sign: 'minus'` and `since` (ISO date), and `perpage` up to 200 (`mine` stays at 50). A type filter queries only its own table; minus-only queries get the same `MAX_EXECUTION_TIME` guard as the text filter. Comment subjects now carry a short `excerpt`, and labels that get cut end with `…`.

The tab fetches 200 votes per request. For periods up to a month it follows up to 5 pages on its own; anything beyond that loads with «Показать ещё».

## Screenshots

Local build on a copy of real staging vote history. Usernames, titles and comment texts are replaced with made-up ones.

<table>
  <tr>
    <td>Default: last 2 weeks</td>
    <td>Light theme</td>
  </tr>
</table>

Filters:

What to show Period

A card opened up, and a comment expanded in place:

Expanded card

Comment expanded in place

Only minuses for a year, mobile, and an empty week:

Only minuses for a year Mobile

Empty week

## Testing

Unit tests for the new query filters and validation, excerpts, digest grouping and the tab itself (default request, auto-loading, menus, URL state, empty state). Also checked by hand on a local stack with the staging-derived data.


Aivean2 (commented):
Comment-only review (I don't stamp non-trivial changes). Two `!!` items I'd fix before merge, then nits.

Good parts first: the #525 N+1 is gone (`getReceivedPostSubjects`/`getReceivedCommentSubjects` batch by id), the excerpt is bounded in SQL instead of shipping full `html` into node, the per-branch keyset + `force index` shape survives the new predicates, and `guarded` correctly extends the typed timeout to minus-only.

### `!!` Minus-only inherits the execution guard but not the budget that guard was priced against

`VoteFeedReadRepository.ts:212` treats `sign === 'minus'` as heavy (`MAX_EXECUTION_TIME` + `VoteFeedFilterTimeoutError`) — but the limiter that pays for that guard keys on something else: `RateLimiters.ts:35` skips unless `req.body.filter` is non-empty, and `UserAPI.receivedVotes()` never sends `filter`. So a minus-only page skips `heavyFilterRateLimiter` (15/min) entirely and runs under `sharedReadRateLimiter` (120/min) → worst case ~60 DB-seconds/minute per user, while the comment at `VoteFeedReadRepository.ts:3-8` still states the ~7.5 DB-s/min bound that assumes the limiter applies.

The worst case is a normal UI combination rather than an attack: «Только минусы» + «За всё время» sends no `since`, so the branch walks the user's whole history through `idx_*_target_voted_at` (which doesn't carry `vote`), with a row lookup per candidate — plus, on the comment branch, a join into `comments` per row for `deleted`/`post_id`. On current data the heaviest recipient has >2000 received comment votes in a 2-week window, 6 of them minuses (~0.3%; comment votes only, so a lower bound) — filling a 200-row minus page means scanning about everything they have.

```ts
// RateLimiters.ts
skip: (req) => !String(req.body?.filter ?? '').trim() && req.body?.sign !== 'minus',
```

(`sign` is whitelisted by Joi immediately after, so there's no `'MINUS'`-style evasion — such a request just 400s.) Probably worth renaming to `heavyReadRateLimiter` at that point, since a text filter is no longer the only heavy shape.

### `!!` Section headers state totals they were not computed from

`buildReceivedDigest` groups whatever happens to be loaded. For `year`/`all`, `bounded` is false, so exactly one 200-vote page is fetched and every month header — the sum, `N чел.`, `чаще всех`, and especially the `По 1–2 оценки в N обсуждениях` tail — describes that 200-event prefix, not the month. The tail line is the sharpest one: a discussion with 5 votes of which 2 are loaded gets printed as a 1–2-vote discussion. Bounded periods hit the same edge past `AUTO_FOLLOW_PAGES * PAGE_SIZE` = 1000 events, which the numbers above make reachable in the default «за 2 недели» for the heaviest users — and `ReceivedVotesDigest.tsx:28` asserts the opposite ("loaded to the end automatically so section sums are exact").

«Показать ещё» does signal there's more, but it sits at the bottom while the precise-looking number sits in the header. Events arrive newest-first, so exactly one section is ever affected: while `feed.hasMore`, either hold the oldest section back or mark it (`+43 · неполно`). Same root cause, visible symptom: during auto-follow discussions hop out of the tail line into cards as pages land.

### Nits

- `type`/`sign`/`since` are accepted for `direction: 'mine'` too — only `perpage` is gated by direction (`UserController.ts:91`) — with no caller and no UI. Intended future surface, or should `when('direction')` cover them?
- `toCommentExcerpt` classifies media on the same `left(c.html, 4096)` prefix, so a comment whose only ``/`<video>` sits past the cut returns an empty excerpt and renders as «комментарий». Rare; a fallback for "stripped to nothing, no tag in the prefix" would close it.
- 200 events/page × up to 5 pages per filter change is a lot of JSON for a tab switch on mobile, and it's all re-fetched when the user immediately re-filters. A smaller first page with the same auto-follow would cut the perceived latency.

Both test suites do run in CI here, and neither finding is a suite question: the first is a predicate mismatch across two files, the second only shows up with more than one page of data.

Aivean (commented):
The grouping is a big improvement over the flat list:
- cards, minus voters listed first, the «и ещё N» timeline and expanding a comment in place all work;
- filters survive in the URL;
- switching filters quickly doesn't leave stale results.

### Still open from the earlier review, now with numbers

- **Section totals are partial.** For the heaviest recipient on the default «За 2 недели», auto-follow stops at 1000 events inside the older week. That week's header shows `+267 −5`, but the DB has about 1720 pluses for it. The complete week matched the API exactly (+724 −4, 128 people). In this data, 9 users receive more than 1000 votes in 2 weeks and 22 in a month. Those are the most active commenters, so the people most likely to open this tab. Marking the oldest section as partial while `hasMore` is true, or holding it back, would be enough.
- **Minus-only requests skip `heavyFilterRateLimiter`.** The `skip` change suggested there applies as is.

### Minus-only: I'd limit it to bounded windows instead of adding an index

«Только минусы» + «За год» fails for the heaviest recipient with «Запрос оказался слишком тяжёлым — сузьте период». The comment branch reads every vote in the window to find the minuses, because `vote` isn't in `idx_comment_votes_target_voted_at`. That was 82k rows to find 248 minuses, taking 3.4 s warm locally against the 500 ms cap. A month for the same user is about 6k rows: 138 ms cold, 56 ms warm.

We already add several indexes to the vote tables for this feature, and I'd rather not add more. So I suggest:
- **UI:** with `sign=minus`, offer only week, 2 weeks and month, or clamp year and all time to a month and say so in the UI.
- **API:** reject `sign: 'minus'` without `since`, or with a `since` older than about 31 days, so the limit isn't enforced only by the UI.
- **Rate limit:** count minus-only requests in the heavy limiter, as above.

This also keeps the tab from becoming a year-long list of everyone who minused you. I'd rather not make that too easy.

If we ever need minus-only over long windows, we can widen `idx_comment_votes_target_voted_at` to `(target_user_id, voted_at, vote)` instead of adding an index. The filter then runs from the index alone. On a scratch copy of the same user's votes, a year went from 0.40 s to 0.14 s. That isn't needed now.

### Other findings (inline)
- On mobile, the «По 1–2 оценки» line overflows the card.
- The filter menus use menu roles but don't support menu keyboard navigation.
- A load-more error hides the retry button.
- An expanded card can't be collapsed.

### Nits
- On meme threads most rows read «[картинка]», which doesn't tell the comments apart until you expand them. The comment time next to it would help.
- While the first page loads, the area under the filters is blank apart from the top progress bar.
- Leftovers from the old received view: 24 unused classes in `UserProfileVotes.module.scss` (`voteRow*`, `voteTimeGroup*`, `voteValue*`, `voteTarget`, `voteVoter`, `voteUsername`, `expandButton`, `expanded`, `subject*`, `groupSubject`, `voteExpand`). `UserVotesResult` in `UserAPI.ts` is now used only in tests.
- `ReceivedVotesDigest.tsx` is 695 lines. Loading, both menus and all the renderers are closures in one component. Splitting out a `useReceivedVotes` hook, a shared `FilterDropdown` (the two menus are near copies) and a `DigestCard` would make it easier to follow. `Sums` also duplicates `sumPlus`/`sumMinus` from the utils.

frontend/src/Components/ReceivedVotesDigest.module.scss
```diff
@@ -0,0 +1,372 @@
...
+  line-height: 1.6;
+  padding: 8px 14px;
+}
+
+.tailHead {
+  color: var(--fgMedium);
+}
+
+.tailItem {
+  white-space: nowrap;
```
Aivean: At 390px width, a long discussion title here doesn't wrap and runs past the card. The page scrolls horizontally: `scrollWidth` 469 vs 390. The ` · ` separator is outside the nowrap span (`ReceivedVotesDigest.tsx:511`), so it can also end up alone at the start of a line. Maybe allow wrapping inside the title, or use `overflow-wrap: anywhere` plus an ellipsis, and move the separator into the span.

frontend/src/Components/ReceivedVotesDigest.tsx
```diff
@@ -0,0 +1,695 @@
...
+          >
+            <>
+              {whatLabel}
+              <span className={styles.caret} aria-hidden='true'>
+                ▾
+              </span>
+            </>
+          </Button>
+          {openMenu === 'what' && (
+            <div className={styles.menu} role='menu' aria-label='Какие оценки показывать'>
```
Aivean: `role='menu'` and `menuitemradio` tell screen readers that arrow keys work, but focus stays on the trigger when the menu opens and ArrowDown does nothing. Escape and outside-click do close the menu, and Tab still reaches the items. Either add roving focus (focus the checked item on open, arrows move, Escape returns focus), or drop the menu roles and treat it as a disclosure with plain buttons and `aria-pressed`. The same applies to the period menu at line 600.

frontend/src/Components/ReceivedVotesDigest.tsx
```diff
@@ -0,0 +1,695 @@
...
+          return
+        }
+        setFeed((current) => (current ? mergePages(current, next) : next))
+        setLoadingMore(false)
+      })
+      .catch(() => {
+        if (abortController.signal.aborted) {
+          return
+        }
+        setError('Не удалось загрузить ещё оценки')
```
Aivean: After a failed «Показать ещё», `error` is set, and the footer at line 673 only renders when `!error`. So the button disappears, and the only way to retry is changing a filter. Maybe keep the loaded feed visible, show the error next to the button, and let the same button retry.

frontend/src/Components/ReceivedVotesDigest.tsx
```diff
@@ -0,0 +1,695 @@
...
+            </span>
+          </div>
+        </header>
+        {(expanded ? card.subjects : card.subjects.slice(0, SUBJECTS_PER_CARD)).map((subject) =>
+          renderSubject(section, subject, site),
+        )}
+        {!expanded && hidden.length > 0 && (
+          <Button
+            variant='minimal'
+            className={styles.expandCard}
```
Aivean: Once a card is expanded there's no way to collapse it. For the heaviest recipient the top discussion expanded to 58 rows, about 3000px. A «свернуть» control at the bottom, or toggling this same button, would help.

PR #543, opened 2026-09-22
fix(parser): keep balanced parentheses inside autolinked URLs

Aivean2:
Reported on Orbitar: an explicitly written link to a Wikipedia disambiguation page renders broken.

```
<a href="https://en.wikipedia.org/wiki/Hill_Valley_(Back_to_the_Future)">https://en.wikipedia.org/wiki/Hill_Valley_(Back_to_the_Future)</a>
```

renders as

```
<a href="https://en.wikipedia.org/wiki/Hill_Valley_" target="_blank">https://en.wikipedia.org/wiki/Hill_Valley_</a>(Back_to_the_Future)
```

### Why

Two mechanisms chain.

1. `urlRegex` (extraction) excludes `(` and `)` from the resource path with no alternative branch, so a bare URL is cut at the first paren. `urlRegexExact` (validation) uses the default `\S*` path and accepts the very same URL — the two regexes disagree.
2. `parseA` returns the children's own parse result whenever they contain a URL, i.e. an autolink found in the link text **replaces** the enclosing anchor (this is deliberate, pinned by the `unwrap nested links` test). So the author's correct `href` is discarded in favour of the truncated inner one.

Only (1) is a bug; fixing it makes (2) harmless, because the inner autolink then reproduces the whole URL.

Beyond Wikipedia, this affects DOI/PII-style links, e.g. `https://www.cell.com/neuron/fulltext/S0896-6273(21)00423-2`.

### Fix

A balanced-parenthesis alternative, ordered first, in the path atom:

```ts
const urlPath = '(?:\\([^\\s()]*\\)|[^\\s.,:;!()\\[\\]{}]|[.,:;!]+\\b)*'
```

It cannot span whitespace or a nested paren, so a URL wrapped in prose parentheses still stops before the closing one — the existing `(http://test.com?q=123)` and `[...]` expectations are unchanged:

| input | link |
|---|---|
| `https://en.wikipedia.org/wiki/Hill_Valley_(Back_to_the_Future)` | whole URL |
| `see (http://test.com/a) after` | `http://test.com/a` |
| `(https://ru.wikipedia.org/wiki/Test_(meaning)) tail` | `https://ru.wikipedia.org/wiki/Test_(meaning)` |
| `http://test.com/foo(bar` | `http://test.com/foo` |

One nesting level, same as other linkifiers. The alternatives are disjoint on their first character, so there is no new backtracking: 20k pathological `(` characters match in <1 ms.

No `TheParser.VERSION` bump (per review — a global reparse is too expensive). Consequence: content already stored with the old rendering keeps it until it is re-saved; editing a comment re-parses it.

### Commits

Two, because `backend/test/parser/urlregex.test.ts` predates prettier/lint-staged and the pre-commit hook reformats it wholesale on any edit. The reformat is isolated in the first commit (no assertions changed); the second is the fix, 71 insertions.

### Checks

`npx tsc --noEmit` clean · `npm run prettier` clean · `npx jest` 379/379 green (4 new assertions blocks in `urlregex.test.ts` and `TheParser.test.ts`).


backend/src/parser/TheParser.ts
```diff
@@ -37,7 +37,7 @@ export default class TheParser {
 
   // Bump this version when introducing breaking changes to the parser.
   // Content will be re-parsed and saved on access when this version changes.
-  static readonly VERSION = 2
+  static readonly VERSION = 3
```
Aivean: please don't bump the parser version, it's expensive to reparse everything.

Aivean2: Dropped — `TheParser.ts` is out of the diff entirely now (7358760).

The same bump was in #544; removed there too (49bac7c), so neither PR touches it.

Consequence, for the record: already-stored content keeps the old rendering until it is re-saved. Editing a comment re-parses it, which is enough for the ones that prompted this.

Aivean (approved):

PR #546, opened 2026-09-23
feat(comments): add user preference for reply-thread border visibility

josser:
Some users find the fgAlmostInvisible left border marking comment nesting too subtle (or too visible); expose a 4-level select in profile settings (Невидимая/Едва заметная/Обычная/Яркая) backed by localStorage, applied via a --commentBorderColor CSS custom property set on :root at app startup and on change.


josser:
@Aivean2, please review

Aivean2:
Thanks for the ping — but I only review PRs from repo collaborators, so this one isn't mine to take.

Flagging for @spaceshelter/reviewers / @Aivean to pick up.

Aivean:
appreciate the effort, but "some users" is basically a single newbie.

as a general principle I'm against introducing new settings in the core site, unless they are absolutely necessary and have significant demand from the community

the rationale is that each such minor setting introduces new permutations, leading to a combinatorial explosion of settings to test major changes against


@ArtCraft , what do you think?

